PR #5, opened 2021-04-26
Lower numpy version requirement

mcara:
`astropy` requires `numpy>=1.17`. This PR brings `numpy` minimum version requirement to `astropy` level. Of course we cannot do that if the code here is using `numpy` features available in let's say `numpy>=1.18`.

CC: @nden @jdavies-st

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,7 @@ setup_requires =
     setuptools_scm
 
 install_requires =
-    numpy>=1.19
+    numpy>=1.17
     astropy
 
 package_dir =


jdavies-st (approved):
Looks good.  Not sure why `numpy` version is specified, but not `astropy`.

jdavies-st:
I'll add that `jwst` supports `numpy==1.17` currently, so it would be good to have this be consistent.  We can also just not  bother with version specifiers in this package and rely on the the packages consuming them to specify versions needed.  I suspect the code currently in this repo runs under a wider range of numpy versions than just >=1.17.

nden (approved):

PR #7, opened 2021-05-14
JP-1950: Data quality bug fix

kmacdonald-stsci:
When computing the median rates, i.e., the NaN median of the first differences of the groups in a given pixel ramp, there is one less number of differences than groups.  A "shifting" must take place to make sure the differences associated with a cosmic ray are properly handled.  In the current code, the "shift" happens twice, which results in an incorrect median computation.

This PR should only be merged after #6 is merged.  From that merge, this PR changes only one line of code in `src/stcal/ramp_fitting/ols_fit.py`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,3 +3,7 @@
 
 - Added code to manipulate bitmasks.
 
+ramp_fitting
+------------
+
+- Added ramp fitting code [#]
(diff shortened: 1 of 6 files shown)


dmggh (approved):
I see the median bug fix you made, which you and I discussed on 4/27. Looks good to me.

nden:
It looks like the base for this this PR was #6. Could you rebase it? Or perhaps it's easier to redo it?

kmacdonald-stsci:
I'll close this and reopen it.  This bug fix will require changes in JWST.  The changes needed to fix this bug effects two unit tests and all regression tests.

PR #12, opened 2021-05-18
JP-1950: Bug fix for DQ flags

kmacdonald-stsci:
The primary purpose of this PR is to fix the computation of the median of the first differences of a ramp.  Because there is one less difference than there are groups a "shift" is needed to make sure the correct data quality flag is applied.  This "shifting" was incorrectly done twice, when it should have only been done once.

### DIFF
--- a/src/stcal/ramp_fitting/ols_fit.py
+++ b/src/stcal/ramp_fitting/ols_fit.py
@@ -861,8 +861,7 @@ def ramp_fit_slopes(input_model, gain_2d, readnoise_2d, save_opt, weighting):
                 rhi = cubeshape[1]
 
             # Skip data section if it is all NaNs
-            # data_sect = np.float32(data[num_int, :, :, :])
-            data_sect = data[num_int, :, :, :]
+            data_sect = np.float32(data[num_int, :, :, :])
             if np.all(np.isnan(data_sect)):
                 log.error('Current data section is all nans, so not processing the section.')
                 continue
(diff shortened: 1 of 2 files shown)


dmggh (approved):

src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1597,10 +1597,8 @@ def calc_slope(data_sect, gdq_sect, frame_time, opt_res, save_opt, rn_sect,
     err_2d_array = data_sect[0, :, :] * frame_time
 
     # Suppress, then re-enable, harmless arithmetic warnings
-    '''
```
jdavies-st, line 1600: You could use Numpy's `errstate` context manager here:

```python
with np.errstate(divide="ignore", invalid="ignore"):
    err_2d_array[err_2d_array < 0] = 0
```

The current error/warning global config for numpy can be grabbed via `np.geterr()`.

nden: The question is why is e.g. `divide by zero`  filtered out. Is there a chance it's going to be raised?
But since this is not related to the current bug being fixed I suggest if necessary it can be changed in another PR.

nden:
👍 merging after a chang elog is added.

src/stcal/ramp_fitting/ols_fit.py
kmacdonald-stsci: I removed the code changes not directly related to the bug in JP-1950.

nden (approved):

PR #21, opened 2021-05-20
Add pypi publish workflow and fix jwst CI

### LINKED ISSUE #3: Enable jwst tests in CI
Currently tests which require CRDS access don't run.

### LINKED ISSUE #18: Set up CI to run on release branches

### LINKED ISSUE #20: Add publish-to-pypi workflow for releases
Just like `jwst`:

https://github.com/spacetelescope/jwst/blob/master/.github/workflows/publish-to-pypi.yml

- [x] Add the last 2 tags, `0.2.0` and `0.2.1` as official releases on Github.  They should be, before we implement the above.

jdavies-st:
Fixes #20 
Fixes #18 
Fixes #3 

The `py38` and `coverage` tests were doing the same thing except one was reporting coverage, so I removed `py38` matrix element in the workflow.  Coverage still needs to be fixed, as we should be checking coverage based on running the `jwst` and `romancal` tests.  Issue filed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - '*.x'
     tags:
       - '*'
   pull_request:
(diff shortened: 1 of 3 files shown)


nden:
Do we need to test with Python 3.8, given that that's the default for jwst now??

nden (approved):
lgtm

PR #23, opened 2021-05-21
Add jump to STCAL

dmggh:
Adding the jump code to be used by both missions to STCAL. This code has DQ flag values hard-coded into the modules, to be replaced after further work (see PR#13) enables DQ flags to be imported within STCAL. The unit tests will be added soon under a different PR.

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -1,0 +1,206 @@
+import time
+import logging
+import numpy as np
+
+from . import twopoint_difference as twopt
+
+import multiprocessing  # 05/18/21: TODO- commented out now; reinstate later
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+# 05/18/21 :TODO instead import these DQ flags
+DO_NOT_USE = 1
+NO_GAIN_VALUE = 524288
+
+
+def detect_jumps(frames_per_group, data, gdq, pdq, err,
+                 in_refout, gain_2d, readnoise_2d, rejection_thresh,
+                 three_grp_thresh, four_grp_thresh, max_jump_to_flag_neighbors,
+                 min_jump_to_flag_neighbors, flag_4_neighbors):
+    """
+    This is the high-level controlling routine for the jump detection process.
+    It loads and sets the various input data and parameters needed by each of
+    the individual detection methods and then calls the detection methods in
+    turn.
+
+    Note that the detection methods are currently set up on the assumption
+    that the input science and error data arrays will be in units of
+    electrons, hence this routine scales those input arrays by the detector
+    gain. The methods assume that the read noise values will be in units
+    of DN.
+
+    The gain is applied to the science data and error arrays using the
+    appropriate instrument- and detector-dependent values for each pixel of an
+    image.  Also, a 2-dimensional read noise array with appropriate values for
+    each pixel is passed to the detection methods.
+
+    Parameters
+    ----------
+    frames_per_group : int
+        number of frames per group
+
+    data : float, 4D array
+        science array
+
+    gdq : int, 4D array
+        group dq array
+
+    pdq : int, 2D array
+        pixelg dq array
+
+    err : float, 4D array
+        error array
+
+    in_refout : float, 4D array
+        reference output
+
+    gain_2d : float, 2D array
+        gain for all pixels
+
+    readnoise_2d : float, 2D array
+        readnoise for all pixels
+
+    rejection_thresh : float
+        cosmic ray sigma rejection threshold
+
+    three_grp_thresh : float
+        cosmic ray sigma rejection threshold for ramps having 3 groups
+
+    four_grp_thresh : float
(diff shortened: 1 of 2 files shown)


ddavis-stsci (changes requested):
Can you update the CHANGES.rst file and it looks like some of your imports are failing during the test run.

nden:
Needs an `__init__.py` file in the `jump` directory.

src/stcal/jump/jump.py
```diff
@@ -0,0 +1,206 @@
...
+    image.  Also, a 2-dimensional read noise array with appropriate values for
+    each pixel is passed to the detection methods.
+
+    Parameters
+    ----------
+    frames_per_group : int
+        number of frames per group
+
+    data : float, 4D array
+        science array
```
cshanahan1, line 44: put expected units here

src/stcal/jump/jump.py
```diff
@@ -0,0 +1,206 @@
...
+        pixelg dq array
+
+    err : float, 4D array
+        error array
+
+    in_refout : float, 4D array
+        reference output
+
+    gain_2d : float, 2D array
+        gain for all pixels
```
cshanahan1, line 59: and units here, and everywhere else where the input has a unit

src/stcal/jump/jump.py
```diff
@@ -0,0 +1,206 @@
...
+        if set to True (default is True), it will cause the four perpendicular
+        neighbors of all detected jumps to also be flagged as a jump.
+
+    Returns
+    -------
+    gdq : int, 4D array
+        updated group dq array
+
+    pdq : int, 2D array
+        updated pixelg dq array
```
cshanahan1, line 93: typo

cshanahan1 (commented):
The ramp fit code already in stcal takes a data model as input, while the `detect_jumps` function here takes the data arrays and attributes that are normally stored in the model as arguments. There should be consistency I think if the inputs are data models or everything separately.

cshanahan1:
Also, what will be the argument to set the # of cores used for multiprocessing once thats turned on? Will it accept the same values that JWST currently does which are 'half', 'all', 'none'/None? I want to at least have something in as a placeholder in JWST for now.

src/stcal/jump/jump.py
```diff
@@ -0,0 +1,206 @@
...
+    gdq : int, 4D array
+        group dq array
+
+    pdq : int, 2D array
+        pixelg dq array
+
+    err : float, 4D array
+        error array
+
+    in_refout : float, 4D array
```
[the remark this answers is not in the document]
dmggh: It's not used;  removing it.

src/stcal/jump/jump.py
dmggh: Fixed.

dmggh:
@nden , @ddavis-stsci , @kmacdonald-stsci  - feel free to make corrections here :

@cshanahan1 :  It was decided that since Roman and JWST ramp data models would not be identical (they have slight differences in their metadata), it wouldn't be feasible to pass either type from the step code into the common code for a calibration step. When we (Nadia, Dave D, Ken, and myself) discussed this several weeks ago, two possible solutions were proposed. Instead of passing the ramp data model : 1) the needed arrays and metadata would be passed as arguments, or 2) the needed arrays and metadata would be repackaged into a new class and passed. At that time there was a preference for option 1.

Since I began modifying the JWST code for the jump step to move it from the jwst repo to the stcal repo (this PR) I have been using the 1st approach. Using the second approach, Ken has most recently been modifying the ramp fitting to move it to stcal ( [spacetelescope/stcal] Al-553: Reduced data model dependency (#26) )

As you have noted, it would be preferable to have consistency. Perhaps all of us should continue this discussion.

Regarding your question about the number of cores: for now just use what JWST uses.

src/stcal/jump/jump.py
dmggh: The units may be mission specific, so I think should be described in the step code instead.

src/stcal/jump/jump.py
dmggh: See above.

src/stcal/jump/jump.py
cshanahan1: The description describes the expected units:

'Note that the detection methods are currently set up on the assumption
    that the input science and error data arrays will be in units of
    electrons, hence this routine scales those input arrays by the detector
    gain. The methods assume that the read noise values will be in units
    of DN.'

so should that also be left vague, or say something like 'they should be in the SAME units' instead?

cshanahan1:
Another non-issue but small suggestion I have would be to change how its imported (this would apply to ramp fit and any future steps too). 

Right now if you want to call detect jumps instead of jump.detect jumps, the import looks like: 

      > from stcal.jump.jump import detect_jumps

But it might be slightly nicer to just have it be
      > from stcal.jump import detect_jumps

Would this be done by adding it to __init__ in the directory above this one?

nden:
Superseded by #27

Issue #24, opened 2021-05-24
For ramp fitting, create a simple internal class that can be pickled for use in multiprocessing.


kmacdonald-stsci:
The `ramp_fit` function in ramp fitting takes a data model as an input parameter.  The full features of a data model are not needed internally for ramp fit processing.  Because of the complexity of the data models they cannot be pickled, which is a requirement for using the `multiprocessing` module in python.

In order to use the `multiprocessing` module internally to ramp fit and simple class can be created from the data model that can easily be pickled.

kmacdonald-stsci:
PR #26 resolves this issue.

PR #26, opened 2021-05-26
Al-553: Reduced data model dependency

kmacdonald-stsci:
To reduce data model dependency an internal ramp fit class has been added.   This has been tested against JWST unit and regression tests.  All tests pass.

Closes issue #24

### DIFF
--- a/src/stcal/ramp_fitting/ols_fit.py
+++ b/src/stcal/ramp_fitting/ols_fit.py
@@ -31,7 +31,7 @@ def ols_ramp_fit_multi(
 
     Parameters
     ----------
-    ramp_data: RampData
+    ramp_data : RampData
         Input data necessary for computing ramp fitting.
 
     buffsize : int
@@ -62,13 +62,13 @@ def ols_ramp_fit_multi(
 
     Returns
     -------
-    image_info: tuple
+    image_info : tuple
         The tuple of computed ramp fitting arrays.
 
-    integ_info: tuple
+    integ_info : tuple
         The tuple of computed integration fitting arrays.
 
-    opt_info: tuple
+    opt_info : tuple
         The tuple of computed optional results arrays for fitting.
 
     gls_opt_model : GLS_RampFitModel object or None
@@ -543,7 +543,7 @@ def ols_ramp_fit_single(
 
     Parameters
     ----------
-    ramp_data: RampData
+    ramp_data : RampData
         Input data necessary for computing ramp fitting.
 
     int_times : None
@@ -566,13 +566,13 @@ def ols_ramp_fit_single(
 
     Return
     ------
-    image_info: tuple
+    image_info : tuple
         The tuple of computed ramp fitting arrays.
 
-    integ_info: tuple
+    integ_info : tuple
         The tuple of computed integration fitting arrays.
 
-    opt_info: tuple
+    opt_info : tuple
         The tuple of computed optional results arrays for fitting.
     """
     tstart = time.time()
@@ -634,12 +634,12 @@ def discard_miri_groups(ramp_data):
 
     Parameters
     ----------
-    ramp_data: RampData
+    ramp_data : RampData
         Input data necessary for computing ramp fitting.
 
     Returns
     -------
-    bool:
+    bool :
         False if no data to process after discarding unusable data.
         True if useable data available for further processing.
     """
@@ -721,7 +721,7 @@ def ramp_fit_slopes(ramp_data, gain_2d, readnoise_2d, save_opt, weighting):
 
     Parameters
     ----------
-    ramp_data: RampData
+    ramp_data : RampData
         Input data necessary for computing ramp fitting.
 
     gain_2d : ndarrays
@@ -1011,7 +1011,7 @@ def ramp_fit_compute_variances(ramp_data, gain_2d, readnoise_2d, fit_slopes_ans)
 
     Parameters
     ----------
-    ramp_data: RampData
+    ramp_data : ramp_fit_class.RampData
         Input data necessary for computing ramp fitting.
 
     gain_2d : ndarray
@@ -1213,13 +1213,13 @@ def ramp_fit_overall(
 
     Parameters
     ----------
-    ramp_data: RampData
+    ramp_data : RampData
         Input data necessary for computing ramp fitting.
 
     orig_cubeshape : tuple
        Original shape cube of input dataset
 
-    orig_ngroups: int
+    orig_ngroups : int
        Original number of groups
 
     buffsize : int
@@ -1242,13 +1242,13 @@ def ramp_fit_overall(
 
     Return
     ------
-    image_info: tuple
+    image_info : tuple
         The tuple of computed ramp fitting arrays.
 
-    integ_info: tuple
+    integ_info : tuple
         The tuple of computed integration fitting arrays.
 
-    opt_info: tuple
+    opt_info : tuple
         The tuple of computed optional results arrays for fitting.
     """
     # Get image data information
@@ -1862,7 +1862,7 @@ def fit_next_segment_all_other(wh_check, start, end_st, end_heads, ngroups):
 
     Parameters
     ----------
-    wh_check: ndarray
+    wh_check : ndarray
         pixels for current segment processing and updating, 1-D
 
     start : ndarray
@@ -1874,7 +1874,7 @@ def fit_next_segment_all_other(wh_check, start, end_st, end_heads, ngroups):
     end_heads : ndarray
         number of endpoints for each pixel, 1-D int
 
-    ngroups: int
+    ngroups : int
         number of groups in exposure
     """
     these_pix = wh_check[0]
@@ -1909,7 +1909,7 @@ def fit_next_segment_good_0th_bad_1st(
 
     Parameters
     ----------
-    wh_check: ndarray
+    wh_check : ndarray
         pixels for current segment processing and updating, 1-D
 
     start : ndarray
@@ -1921,10 +1921,10 @@ def fit_next_segment_good_0th_bad_1st(
     end_heads : ndarray
         number of endpoints for each pixel, 1-D int
 
-    got_case: ndarray
+    got_case : ndarray
         classification of pixel for current semiramp, 1-D
 
-    ngroups: int
+    ngroups : int
         number of groups in exposure
     """
     these_pix = wh_check[0]
@@ -1951,7 +1951,7 @@ def fit_next_segment_only_good_0th_group(
 
     Parameters
     ----------
-    wh_check: ndarray
+    wh_check : ndarray
         pixels for current segment processing and updating, 1-D
 
     start : ndarray
@@ -1966,7 +1966,7 @@ def fit_next_segment_only_good_0th_group(
     pixel_done : ndarray
         whether each pixel's calculations are completed, 1-D bool
 
-    got_case: ndarray
+    got_case : ndarray
         classification of pixel for current semiramp, 1-D
 
     f_max_seg : int
@@ -1980,20 +1980,20 @@ def fit_next_segment_only_good_0th_group(
     num_seg : ndarray
         numbers of segments for good pixels, 1-D int
 
-    slope: ndarray
+    slope : ndarray
        weighted slope for current iteration's pixels for data section, 1-D
        float
 
-    intercept: ndarray
+    intercept : ndarray
        y-intercepts from fit for data section, 1-D float
 
-    variance: ndarray
+    variance : ndarray
        variance of residuals for fit for data section, 1-D float
 
-    sig_intercept: ndarray
+    sig_intercept : ndarray
        sigma of y-intercepts from fit for data section, 1-D float
 
-    sig_slope: ndarray
+    sig_slope : ndarray
        sigma of slopes from fit for data section (for a single segment), 1-D
        float
 
@@ -2053,7 +2053,7 @@ def fit_next_segment_short_seg_not_at_end(
 
     Parameters
     ----------
-    wh_check: ndarray
+    wh_check : ndarray
         pixels for current segment processing and updating, 1-D
 
     start : ndarray
@@ -2068,7 +2068,7 @@ def fit_next_segment_short_seg_not_at_end(
     pixel_done : ndarray
         whether each pixel's calculations are completed, 1-D bool
 
-    got_case: ndarray
+    got_case : ndarray
         classification of pixel for current semiramp, 1-D
 
     f_max_seg : int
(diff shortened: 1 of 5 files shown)


src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -87,7 +121,65 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
...
+
+    return ramp_fit_internal(
+        ramp_class, buffsize, save_opt, readnoise_2d, gain_2d,
+        algorithm, weighting, max_cores, dqflags)
+
+
+def ramp_fit_internal(ramp_class, buffsize, save_opt, readnoise_2d, gain_2d,
+                      algorithm, weighting, max_cores, dqflags):
+    """
+    ramp_class: RampFitInternal
```
dmggh, line 134: A summary here of what this does would be helpful, with so many functions already similarly named.

src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -22,7 +22,7 @@
 
 
 def ols_ramp_fit_multi(
-        input_model, buffsize, save_opt, readnoise_2d, gain_2d, weighting, max_cores):
+        ramp_class, buffsize, save_opt, readnoise_2d, gain_2d, weighting, max_cores):
```
dmggh, line 25: I find it a bit confusing to have the name of an instance of a model be 'ramp_class'. Instead, perhaps something containing the words 'model' and 'ramp' but not mission-specific, and something clearly distinct from anything we already use.

src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1194,23 +1194,23 @@ def ramp_fit_compute_variances(input_model, gain_2d, readnoise_2d, fit_slopes_an
...
+    ramp_class.groupdq = groupdq
+    ramp_class.pixeldq = inpixeldq
 
     return var_p3, var_r3, var_p4, var_r4, var_both4, var_both3, inv_var_both4, \
         s_inv_var_p3, s_inv_var_r3, s_inv_var_both3
 
 
 def ramp_fit_overall(
-        input_model, orig_cubeshape, orig_ngroups, buffsize, fit_slopes_ans,
+        ramp_class, orig_cubeshape, orig_ngroups, buffsize, fit_slopes_ans,
```
dmggh, line 1207: Also here, a summary of what this function does would be helpful, considering the number of existing similarly named functions.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -87,7 +121,65 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
         Object containing optional GLS-specific ramp fitting data for the
         exposure
     """
+    ramp_class = create_ramp_fit_class(model, dqflags)
```
dmggh, line 124: Perhaps a comment stating that this is an instance of an internal ramp class.

src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -0,0 +1,118 @@
+class RampFitInternal:
```
dmggh, line 1: I think this would work, but then we would have JWST ramp models, Roman ramp models, and these internal general ramp models. All of these are quite similar, so perhaps they could instead be inherited from a parent ramp fit class, importable from within jwst, romancal, and stcal.  And within this new class structure we could incorporate the mission-specific DQ flags information, instead of having a new parameter.

dmggh (commented):
Comments on the new model code -

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: This function exists as a buffer to allow for passage of a data model to `ramp_fit`.  That function creates a class to containing only the needed information for ramp fitting then calls this function.  All code after this function call no longer depends on a data model.

src/stcal/ramp_fitting/ols_fit.py
kmacdonald-stsci: The `ramp_class` parameter is an instance of the `RampFitInternal` class, not a data model.  The `RampFitInternal` class inherits no other class and has only `ndarray` and basic data types.  No code beyond `ramp_fit` in `ramp_fit.py` depends on a data model instance.  The class is used in preparation for implementing multiprocessing.  The data models cannot be used with the `multiprocessing` module because they cannot be pickled.  A simpler data structure is required for this.

src/stcal/ramp_fitting/ols_fit.py
kmacdonald-stsci: Thank you for noticing this.  I added a summary of the function to the docstring.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I added a comment to clarify a class is being created using only the needed information from the data model.

src/stcal/ramp_fitting/ramp_fit_class.py
kmacdonald-stsci: The reason I created this simple class is because data models are too complex to be pickled.  The point of this PR is to create a simple class containing only the information needed for ramp fitting computation in order to be pickled in preparation for using the `multiprocessing` module.

src/stcal/ramp_fitting/ramp_fit.py
dmggh: Ok.

src/stcal/ramp_fitting/ols_fit.py
dmggh: Ok, I understand now. This class could also be useful for other calibration steps.

src/stcal/ramp_fitting/ramp_fit_class.py
dmggh: Yeah, I know that data models can't be pickled.

Now that I've looked at this more closely, I understand how this class can be used for multiprocessing.

src/stcal/ramp_fitting/ramp_fit.py
jdavies-st: Yeah, this should have at least 1-line docstring before the summary of Parameters.

```
"""
Description of what this function does

Parameters
----------
ramp_class : RampFitInternal
...
```

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -19,13 +19,47 @@
...
+    ramp_class.set_meta(
+        name=model.meta.instrument.name,
+        frame_time=model.meta.exposure.frame_time,
+        exp_ngroups=model.meta.exposure.ngroups,
+        group_time=model.meta.exposure.group_time,
+        groupgap=model.meta.exposure.groupgap,
+        nframes=model.meta.exposure.nframes,
+        drop_frames1=model.meta.exposure.drop_frames1)
+
+    ramp_class.set_dqflags(dqflags)
```
jdavies-st, line 58: Is there a reason to have this all called here instead of being bundled in `RampFitInternal.__init__()`?  I.e. can't this function be that init method?  Is there a reason to have it here?

Or should this be in the calling code from `RampFitStep`?  Will `jwst` and `romancal` construct this class in exactly the same way?  I suspect not, as `romancal` will have to broadcast their ndim=3 data to ndim=4.

src/stcal/ramp_fitting/ols_fit.py
jdavies-st, line 25: I think having an instance called "ramp_class" is a bit confusing. I know it is an instance of a class, but that's always the case with python objects.  It would be better to not have "class" in the name.  Maybe `ramp_data`? Or something similar?

src/stcal/ramp_fitting/ramp_fit_class.py
kmacdonald-stsci: I changed the class name to `RampData`.

src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        # Arrays from the data model
+        self.data = None
+        self.err = None
+        self.groupdq = None
+        self.pixeldq = None
+        self.int_times = None
+
+        # Meta information
+        self.instrume = None
```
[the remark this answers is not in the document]
kmacdonald-stsci: I changed this class variable to `instrument_name`.

src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -0,0 +1,118 @@
...
+
+        self.frame_time = None
+        self.exp_ngroups = None
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed the `exp_ngroups` variable.

src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -0,0 +1,118 @@
...
+        self.drop_frames1 = None
+
+        # Data quality flags
+        self.flags_do_not_use = None
+        self.flags_jump_det = None
+        self.flags_saturated = None
+        self.flags_no_gain_val = None
+        self.flags_unreliable_slope = None
+
+    def set_arrays(self, data, err, groupdq, pixeldq, int_times):
```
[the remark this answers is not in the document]
kmacdonald-stsci: I have been told the pieces in `jwst` and `roman` will be the same and have the same variable names, so the a data model can be passed to `ramp_fit`.  If that is not the case, then the creation of this class to be used in ramp fitting will need to be pushed to the step code.

The `__init__` function defines all the necessary variables.  The class variables are set using methods for a couple of reasons.  There are 16 variables.  Having 16 input parameters into a single function is cumbersome and hard to read.  The class methods set variables of similar types.  This approach makes it more flexible to use in different pipelines, should this class instantiation need to be pushed back to step code, depending on how and where these pieces of information are defined in a particular pipeline.

The class can be pickled.  It needs to be able to be pickled in order to for mulitprocessing.  Since the class can be pickled disassembly and reassembly of the components of the class are unnecessary.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Thanks for catching the docstring.  I added it.

src/stcal/ramp_fitting/ols_fit.py
kmacdonald-stsci: I changed all instances of "ramp_class" as a name to "ramp_data".

src/stcal/ramp_fitting/ramp_fit_class.py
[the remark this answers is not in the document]
dmggh: As I understand it, not all of the jwst and roman pieces are the same. For example, instead of  JWST 'filter', Roman has 'optical_element'.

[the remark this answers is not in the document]
nden: I think the point is that this class represents the data needed in ramp fitting only, not the entire data model. Two different data models can be passed in but as long as they have `data, err, groupdq, pixeldq, int_times` .
That said, if there are differences in the (meta)data the Roman converters can take care of making sure they are the same, e.g. `optical_element` can be turned into `filter`.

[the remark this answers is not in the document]
kmacdonald-stsci: The code is designed such that a data model can be passed to `ramp_fit`, so long as it contains the proper values and variable names.  The creation and population of the class can occur in the step code with the step code calling `ramp_fit_internal`.  I changed the name from `ramp_fit_internal` to ramp_fit_data`, since that function may not only be used internally within the ramp fitting code.

jdavies-st (dismissed):
This looks good.  Minor question below.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -19,13 +19,46 @@
...
 
 
+def create_ramp_fit_class(model, dqflags=None):
+    """
+    Create an internal ramp fit class from a data model.
+
+    Parameters
+    ----------
+    model : data model
+        input data model, assumed to be of type RampModel
```
nden, line 37: Add `dqflags` to `Parameters`

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -19,13 +19,46 @@
...
+
+    Return
+    ------
+    ramp_data : RampData
```
nden, line 41:
```suggestion
    ramp_data : `ramp_fit_class.RampData`
```

nden: The above will create a link to the class.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -87,7 +120,74 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
...
+def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
+                  algorithm, weighting, max_cores, dqflags):
+    """
+    This function begins the ramp fit computation after the creation of the
+    RampData class.  It determines the proper path for computation to take
+    depending on the choice of ramp fitting algorithms (which is only ordinary
+    least squares right now) and the choice of single or muliprocessing.
+
+
+    ramp_data: RampData
```
nden, line 142:
```suggestion
    ramp_data : `ramp_fit_class.RampData`
```

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -87,7 +120,74 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
...
+        Input data necessary for computing ramp fitting.
+
+    buffsize : int
+        size of data section (buffer) in bytes
+
+    save_opt : bool
+       calculate optional fitting results
+
+    readnoise_2d: ndarray
```
nden, line 151: Here ans else where leave a space bwtween the parameter name and `:`.

nden (dismissed):
LGTM.
I left a few comments about docstrings.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I added this parameter.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I updated the docstrings in this file.

nden (approved):

jdavies-st (approved):
🚀

PR #27, opened 2021-06-04
Move common jump code to stcal (RCAL-152).

dmggh:
Modified jump code to be common code within STCAL. Modified DQ flags to be retrieved from either ROMANCAL or JWST. I still need to commit updated CHANGES.rst from a different PR, as it is in a different (ROMANCAL) repo. After this is merged, romancal unit tests will be filed under RCAL-114 .

### DIFF
--- a/src/stcal/jump/__init__.py
+++ b/src/stcal/jump/__init__.py
@@ -1,0 +1,3 @@
+from .jump_step import JumpStep
+
+__all__ = ['JumpStep']
--- a/src/stcal/jump/__init__.py
+++ b/src/stcal/jump/__init__.py
@@ -1,3 +1,0 @@
-from .jump_step import JumpStep
-
-__all__ = ['JumpStep']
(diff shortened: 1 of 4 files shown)


ddavis-stsci (approved):
LGTM now with the fixes.

nden:
Should #23 be closed?

dmggh:
Yes, once this PR (#27) is approved a 2nd time and merged, I was planning to close #23, which this supercedes.

nden (approved):

nden:
Approved but let's give @cshanahan1 a chance to look at it.

cshanahan1:
My only comments (which I discussed with @dmggh on Slack) are that having a 'constants.py' file for each step individually might not be the best solution to accomplish what it does, but it works for now. 

Otherwise, I think this looks good. Any other comments about the routines I made are in #23, but those were all addressed.

cshanahan1 (approved):

PR #29, opened 2021-06-14
Fix codecov uploads

jdavies-st:
Use Codecov's github action to upload coverage.


jdavies-st:
Merging this to get a baseline coverage for the `main` branch.  Can do further tweaks to get the `jwst` and `romancal` tests to properly report coverage.  Or move the tests in those packages here, which is probably the better and more reliable solution.

PR #34, opened 2021-07-10
DQ Support

PaulHuwe:
Closes #

Partially resolves RCAL-92 & 137

Description
This PR adds support for DQ flagging. 

Checklist

[ x ] Tests

[ ] Documentation

 Change log

[ ] Milestone

[ ] Label(s)


src/stcal/dq_init/dq_initialization.py
```diff
@@ -0,0 +1,99 @@
...
+    -------
+    output_model : science datamodel
+        The dq corrected science datamodel
+    """
+
+    # Inflate empty DQ array, if necessary
+    check_dimensions(input_model)
+
+    # Create output model as copy of input
+    output_model = input_model.copy()
```
nden, line 28: I wonder if this copy is really necessary but this is a discussion for the JWST group.

src/stcal/dq_init/dq_initialization.py
```diff
@@ -0,0 +1,99 @@
...
+    #         # If the shape is different, then the mask model should have
+    #         # a shape of (0,0).
+    #         # If that's the case, create the array
+    #         if input_model.dq.shape == (0, 0):
+    #             input_model.dq = np.zeros((input_shape[-2:])).astype('uint32')
+    #         else:
+    #             log.error("DQ array has the wrong shape: (%d, %d)" %
+    #                       input_model.dq.shape)
+    #
+    # else:   # RampModel
```
nden, line 78: If this is commented out how is it going to be used by JWST. 
Would it be better to leave `check_dimensions` in the respective pipelines?

nden (changes requested):

jdavies-st (commented):
This is difficult to review, since it is not clear if anything in the copied-over code from `jwst.dq_init.dq_initialization` was changed when copying.  I do see that the `correct_model()` function is no longer defined in this module.  Presumably that will be modified in `jwst` at a later date?

Also at what commit hash in `jwst` was this code copied?  That would be good to record in this PR via a commit message.  I.e. a commit with only the copied code verbatim, then subsequent commit(s) showing changes needed to make it work as an independent module in `stcal`, and specifically any changes needed to make it work with `romancal`.

As it is, this PR is a black box, as the diff is not useful, the code is not tested via unit tests in `stcal`, and it's not clear it will work with `jwst` as-is.

Generally, it would be good to do any structural changes that need to be done within the `jwst` package in a PR there, as that's where all the unit tests are.  And we don't want to break those.  Then when moving to `stcal`, the only thing needed is to copy a single module verbatim into `stcal` and change the import in `jwst` to get the code from `stcal`.  That keeps things very simple and makes understanding what is happening much easier, as the diffs make sense, and the order of operations of what changed when and for what reasons make sense.

nden:
I agree with @jdavies-st .

PaulHuwe:
@jdavies-st  Yeah, I did not test replacing JWST's dq code with this yet, as there was so much to do with getting it working for roman, with all the new libraries. Maybe it was premature to add you to the ticket, but I am glad for your feedback. 

@nden @ddavis-stsci  @jdavies-st My development plan is as follows (please let me know if you think I should be doing something different):

1. Sync up dq code from jwst
2. Get DQ code working for Roman (across all 4 libraries: RAD, RDM, STCAL, RomanCAL).
3. Fix any issues brought by reviewers. 
4. Get DQ code working with JWST (and Roman), then shift JWST over to STCAL (via JWST PR, obviously). 

I can undo what I have here and modify step 1 to match something closer to what James suggested above, if desired. Please give me any other changes to my proposed method that you have.

src/stcal/dq_init/dq_initialization.py
hbushouse: That's been debated back and forth over the years. The purpose of the copy is so that when people run the step interactively from within python their input data does *not* get overwritten/updated in place by the output, but instead returns a separate output. It does lead to extra memory use, unfortunately. At one point we contemplated having a step-level param that would indicate whether the input should be updated in place or a new output created.

hbushouse: The update-in-place approach could be the default for use in Ops, where no one cares about the intermediate results of individual steps that are passed along within a pipeline module.

src/stcal/dq_init/dq_initialization.py
hbushouse: Guider data models were implemented a long time ago, so this code could be updated or activated on the JWST side. Obviously no one has looked at this step in such a long time that this to-do item was forgotten.

hbushouse: Wait. I don't see this commented code in the current JWST implementation anywhere. Where did it come from? Was this commented out for the purpose of the Roman implementation?

jdavies-st: Yeah, it's commented out here, but not in `jwst`.  So that's a non-starter.

Howard, is there a better way we can identify this data other than using an `isinstance()` call?  I.e. is there a specific EXP_TYPE or similar that could live here and would not need access to `jwst.datamodels.GuiderRawModel`?  I.e. how do we know when to run `calwebb_guider` on an exposure?

hbushouse: Yes, the jwst version defines a list of EXP_TYPE's that are exclusive to guide star data, so any datamodel that has one of those EXP_TYPE's should be a straight substitute for testing the model instance. For JWST the selection of `calwebb_guider` is done within the SDP workflow manager. But the reason for this test in the `dq_init` step is because that step is used for both normal science processing (`calwebb_detector1`) and guide star data processing (`calwebb_guider`). And unfortunately the content of the science and guider data models is slightly different, in that science modes have PIXELDQ and GROUPDQ arrays, while guider only has a generic DQ array. Hence the need for this test in the `dq_init` step.

PaulHuwe:
Closing this to restart DQ implementation under the proper method.

PR #37, opened 2021-07-21
added tests for two point difference

cshanahan1:
Adding tests for two point difference in jump detection.


nden:
@cshanahan1 Can you add 

- docstrings to tests with a description of what they test
- reason for xfailing tests in the `xfail` decorator

Both will facilitate reviews and further development.

dmggh (commented):
The existing tests look fine. I think it would be useful (in terms of code coverage ) to have a wider variety of values of the arguments in find_crs( ), such as :

1) All flag_4_neighbors are now "False", so add some 'True' tests. 

2)  All values of the rejection threshold arguments ( normal_rej_thresh, two_diff_rej_thresh, and three_diff_rej_thresh)  are equal to each other. Add some tests with physically plausible and non-equal values.

3)  All values of the min_jump_to_flag_neighbors argument = 10, and all values of the max_jump_to_flag_neighbors argument = 200. Add a test or 2 with different values.

nden (approved):

PR #39, opened 2021-08-06
adding common code for saturation

cshanahan1:
This PR adds the saturation flagging code.

The unit tests for the corresponding PR in JWST are passing : https://github.com/spacetelescope/jwst/pull/6317

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+0.3.0 (unreleased)
+==================
+
+saturation
+----------
+
+- Adds common code for saturation [#39]
+
 0.2.5 (2021-08-27)
 ==================
 
--- a/src/stcal/saturation/saturation.py
+++ b/src/stcal/saturation/saturation.py
@@ -1,0 +1,95 @@
+import numpy as np
+import logging
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+
+def flag_saturated_pixels(data, gdq, pdq, sat_thresh, sat_dq, atod_limit,
+                          dqflags):
+    """
+    Short Summary
+    -------------
+    Apply flagging for saturation based on threshold values stored in the
+    saturation reference file data `sat_thresh` and A/D floor based on testing
+    for 0 DN values. For A/D floor flagged groups, the DO_NOT_USE flag is also
+    set.
+
+    Parameters
+    ----------
+    data : float, 4D array
+        science array
+
+    gdq : int, 4D array
+        group dq array
+
+    pdq : int, 2D array
+        pixelg dq array
+
+    sat_thresh : `np.array`
+        Pixel-wise threshold for saturation, same shape `data`
+
+    sat_dq : `np.array`
+        data quality flags associated with `sat_thresh`
+
+    atod_limit : int
+        hard DN limit of 16-bit A-to-D converter
+
+    dqflags: dict
+        A dictionary with at least the following keywords:
+        DO_NOT_USE, SATURATED, AD_FLOOR, NO_SAT_CHECK
+
+
+    Returns
+    -------
+    gdq : int, 4D array
+        updated group dq array
+
+    pdq : int, 2D array
+        updated pixel dq array
+    """
+
+    nints = data.shape[0]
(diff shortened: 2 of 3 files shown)


cshanahan1:
i'm not sure why this is decreasing test coverage by 50%, the only files I added are 100% covered

jdavies-st:
> i'm not sure why this is decreasing test coverage by 50%, the only files I added are 100% covered

You typed too soon.  To get accurate coverage, you have to wait for all the CI jobs to report their coverage, and that includes the `jwst` job.  Those tests cover this code the best, but they take the longest.

Now that all checks are completed, you can see that the coverage is good.

CHANGES.rst
```diff
@@ -18,6 +23,11 @@ Workaround for setuptools_scm issues with recent versions of pip. [#45]
 
 ramp_fitting
 ------------
+=======
+saturation
+----------
+- Move common saturation code to stcal [#39]
+
```
jdavies-st, line 30: This should not be here.

src/stcal/saturation/saturation.py
```diff
@@ -0,0 +1,96 @@
+import logging
+
+import numpy as np
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
```
jdavies-st, line 6: We should not be setting log levels in any library code like this.

jdavies-st (commented):
Looks good.  Tests pass with https://github.com/spacetelescope/jwst/pull/6317 so everything works.  Just some comments below.

Also, the unit tests should not be inline in `src/stcal/saturation`.  Instead they should be in the top-level `tests` dir.  Just name the module `test_saturation.py` or something and put them there.  This keeps tests separated from the code.

jdavies-st:
Oh and I see the `jump` tests are also inline. Feel free to move those out into the top-level `tests` dir as well in a separate PR.

src/stcal/saturation/saturation.py
hbushouse: This is a carryover from the kludge needed in the jwst package, due to the odd way the logging mechanism interacts with stpipe (or something). If you don't have line 6 in there, you won't ever be able to get anything other than INFO level messages out of the function. This is not actually setting the logging level to DEBUG for every usage, it's just *enabling* DEBUG level to be used when desired. 

Of course if someone has fixed all of this nonsense in stcal, such that this is no longer needed, then ignore everything I just said (a lot of people do anyway).

src/stcal/saturation/tests/test_saturation.py
```diff
@@ -0,0 +1,43 @@
...
+Unit tests for saturation flagging
+
+"""
+
+import numpy as np
+
+from stcal.saturation.saturation import flag_saturated_pixels
+
+
+def test_basic_saturation_flagging():
```
kmacdonald-stsci, line 12: Will the other tests from JWST be moved to this unit tests or will they remain in the JWST unit tests?

kmacdonald-stsci (approved):
Looks good.  I have only one question about the unit tests.  Will the other unit tests in JWST be moved to STCAL or will they remain in JWST?

jdavies-st (changes requested):
.

src/stcal/saturation/saturation.py
jdavies-st: There are no debug-level log messages in this code, so I'd suggest we don't need it regardless.  But yeah, we should get that fixed in `stpipe`.

src/stcal/saturation/tests/test_saturation.py
cshanahan1: i think they should eventually all be moved to stcal. i put a simple test in there just so there was coverage, it will take a while to rewrite the tests without datamodels so i will do that in a separate PR

CHANGES.rst
```diff
@@ -6,6 +6,11 @@ ramp_fitting
 
 - Adds support for Roman ramp data. [#49]
 
+saturation
```
nden, line 9: This needs to be moved to a new section - version 0.3.

jdavies-st (approved):

PR #43, opened 2021-08-20
Adds support for roman ramp data

nden:
The Roman `RampModel` is missing `meta.exposure.drop_frames1` and `model.int_times`. 
This adds a check that those attributes exist before attempting to access them.


kmacdonald-stsci (approved):
Looks good.

PR #54, opened 2021-10-06
JP-2290: Global DQ variable removed

kmacdonald-stsci:
Removed use of global DQ flag variable.  The DQ flags are now passed through the RampData class.


src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -16,7 +16,7 @@
 import numpy as np
 import logging
 
-from . import constants
+# from . import constants
```
stscieisenhamer, line 19: If commented out, should these lines just be removed?

stscieisenhamer (approved):
LGTM

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Yes.  I removed the comment.

Issue #33, opened 2021-06-21
Multiprocessing for ramp fitting returns different values than single processing


kmacdonald-stsci:
With the re-implementation of multiprocessing in ramp fitting after moving ramp fitting to STCAL a bug was found that shows multiprocessing computing different values than single processing.  The computation of the Poisson variance are different for a large number of pixels and the differences have large magnitudes.  

Multiprocessing returns the same optional results product as single processing does, so the problem comes in when combining these results across groups and integrations.  The testing done so far also tests single integration inputs, so the problem may just be across groups.  Multiprocessing divides the work up by row.

kmacdonald-stsci:
This is caused by the following lines: 


https://github.com/kmacdonald-stsci/stcal/blob/9935c7a96fba7d0d1127a51912b5842f80f8bf97/src/stcal/ramp_fitting/ols_fit.py#L1165

 
https://github.com/kmacdonald-stsci/stcal/blob/9935c7a96fba7d0d1127a51912b5842f80f8bf97/src/stcal/ramp_fitting/ols_fit.py#L1172

The sum at line 1172 is done over the first dimension of a cube, which has dimensions (max_num_segs, rows, cols).  During multiprocessing the data is sliced up by row.  During the creation of the cube over which this sum is done could have a different maximum number of segments, since that maximum is computed over each sliced data set, rather than the whole data.

This problem occurs for the degenerate case where the slope estimate is non-positive.  The computation of the Poisson variance for a pixel begins with the Poisson variance of a segment in a pixel ramp.  The segment Poisson variance depends on the slope estimate.  The slope estimate is the NAN median of the first differences of all groups in a ramp over all integrations.  If this is non-positive, each segment Poisson variance will be non-positive, so all Poisson segment variances will be set to a large variance defined as done at line 1165.  

The resultant sum is a multiple of the inverse of that large variance.  The multiple is the maximum number of segments in a slice.   The maximum number of segments in a slice could very enough to cause very large differences in the exposure level Poisson variance for a pixel.

kmacdonald-stsci:
This is fixed with PRs #68.

PR #59, opened 2021-10-20
JP-2293 Poisson computation for negative median rates

kmacdonald-stsci:
A degenerate case occurs if a negative value for the median of the first differences for a ramp is computed.   The causes the Poisson variance to be computed as zero, resulting in nonsense values for variance computation.  For all products returned by ramp fitting the Poisson variance is computed as zero for all pixels with negative median rates and the ERR arrays are functions of only the read noise for those pixels.

Fixes issue #33

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 ramp_fitting
 ------------
 
-- For slopes with negative median rates, the Poisson variance is zero.
+- For slopes with negative median rates, the Poisson variance is zero. [#33]
 
 0.4.1 (2021-10-14)
 ==================


tests/test_ramp_fitting.py
```diff
@@ -0,0 +1,408 @@
...
+}
+
+
+# -----------------------------------------------------------------------------
+#                           Test Suite
+
+
+def base_neg_med_rates_single_integration():
+    """
+    Ensures negative Poisson
```
hbushouse, line 28: Not sure that this comment really tells the full story. We don't want negative Poisson noise values, so why would we want to ensure it?  I assume this is meant to indicate something along the lines of "Ensure negative slopes don't result in negative Poisson variance" or something like that?

tests/test_ramp_fitting.py
```diff
@@ -0,0 +1,408 @@
...
+    sdata, sdq, svp, svr, serr = slopes
+    idata, idq, ivp, ivr, int_times, ierr = cube
+    oslope, osigslope, ovp, ovr, \
+        oyint, osigyint, opedestal, oweights, ocrmag = optional
+    """
+
+    print(" ")
+    print_slopes(slopes)
+    print_integ(cube)
+    print_optional(optional)
```
hbushouse, line 408: Do we really want all this info printed during test runs? What happens to the output when these are run in CI or Jenkins environments?

hbushouse (commented):
The functionality in the OLS code looks good. Just a couple comments on the new tests.

Given that this results in a change in behavior of the ramp_fit step output, it seems it should be documented in a change log entry in the jwst repo, because jwst makes use of it (if you only document in the stcal change log, many folks may not see it).

tests/test_ramp_fitting.py
kmacdonald-stsci: These print functions are not run every time tests run.  They are not called in any test.  I use them for debugging and find myself rewriting them every time I write tests, so I kept them in the file this time.  I can remove them if you don't think they belong there.

tests/test_ramp_fitting.py
kmacdonald-stsci: I updated the comments for these functions to make them more clear, especially for the base functions that tests used.

dmggh (commented):
1) To  follow up on Howard's comment ; perhaps a comment such as  'Create single integration data for testing that negative median rates are handled appropriately'.

2)  These are tests for the new code that handles negative median rates - why are these the only ramp fitting tests in this file ?

3)  Minor question - why do you use both np.testing.assert_AAA(  ) and assert(  ) ?

kmacdonald-stsci:
Dave,

1. I have changed the comments on these tests and helper functions.
2. The original ramp fitting tests remained in JWST when the ramp fitting code got moved to STCAL.
3. The `assert()` is used when checking a single value, while `np.testing.assert_allclose()` is used to check numpy arrays.

dmggh (approved):
Looks good to me.

PR #60, opened 2021-10-21
JO-2297: Properly hand DO_NOT_USE DQ flag for multi-integration processing

kmacdonald-stsci:
It is possible during ramp fitting for a pixel to be marked as `DO_NOT_USE` due to saturation in one integration, but not another.  For pixels good data in one or more integrations having unusable date in other integrations, the final DQ flag should not be marked as `DO_NOT_USE`.

Fixes JWST issue #6379 https://github.com/spacetelescope/jwst/issues/6379

### DIFF
--- a/src/stcal/ramp_fitting/utils.py
+++ b/src/stcal/ramp_fitting/utils.py
@@ -1390,7 +1390,7 @@ def compute_slices(max_cores):
     return number_slices
 
 
-def dq_compress_final(dq_int, n_int):
+def dq_compress_final(dq_int, n_int, dnu_flag):
     """
     Combine the integration-specific dq arrays (which have already been
     compressed and combined with the PIXELDQ array) to create the dq array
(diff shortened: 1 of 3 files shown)


src/stcal/ramp_fitting/utils.py
```diff
@@ -1411,10 +1411,24 @@ def dq_compress_final(dq_int, n_int):
...
+    # the number of integrations with DO_NOT_USE set is less than the
+    # total number of integrations, that means at least one integration
+    # has good data, so the final DQ flag for that pixel should NOT
+    # include the DO_NOT_USE flag.
+    dnu = np.zeros(dq_int.shape, dtype=np.uint32)
+    dnu[np.where(np.bitwise_and(dq_int, dnu_flag))] = 1
+    dnu_sum = dnu.sum(axis=0)
+    not_dnu = np.where(dnu_sum < nints)
+
+    not_dnu_flag = ~dnu_flag & 0xffffffff  # Forces it to be 32-bits
```
hbushouse, line 1429: would it also work to just add ".astype(uint32)" to "not_dnu_flag"?

hbushouse (commented):
Just one minor comment/question

src/stcal/ramp_fitting/utils.py
kmacdonald-stsci: It doesn't appear that `.astype(uint32)` works for an integer, only for an `ndarray` or `pandas` data frame.  However, there is an `np.uint32()` function available.  I changed the code to use this function, instead of the cryptic `& 0xffffffff`.

hbushouse (approved):
Needs a change log entry. Other than that, it looks good.

tests/test_ramp_fitting.py
```diff
@@ -0,0 +1,246 @@
...
+    # Make sure DO_NOT_USE is set in the expected final DQ.
+    assert(dq[0, 0] & dqflags["DO_NOT_USE"])
+    assert(not(dq[0, 1] & dqflags["DO_NOT_USE"]))
+    assert(not(dq[0, 2] & dqflags["DO_NOT_USE"]))
+
+
+# -----------------------------------------------------------------------------
+#                           Set up functions
+
+# Need test for multi-ints near zero with positive and negative slopes
```
dmggh, line 73: "Near zero" ? I don't quite follow.

tests/test_ramp_fitting.py
```diff
@@ -0,0 +1,246 @@
...
+}
+
+
+# -----------------------------------------------------------------------------
+#                           Test Suite
+
+
+def test_utils_dq_compress_final():
+    """
+    Set up a multi-integration 3 pixel data array each ramp as the following:
```
dmggh, line 28: I don't follow what is being tested here - can you add some clarification ?

kmacdonald-stsci: I am expanding the comment for this test.  Let me know if I need to expand it further.

PR #51, opened 2021-08-30
AL-576: Ramp fit GLS

kmacdonald-stsci:
Add the GLS algorithm back to ramp fitting, as it was before refactoring and putting the code in STCAL.  The GLS code functions, but still needs work.  Many of the original tests were skipped or marked as failed in the original JWST code.  The tests have been added to STCAL and modified to run without dependence on anything outside of STCAL.

Issue #38

PR #64, opened 2021-11-10
AL-576: Adding GLS to ramp fitting.

kmacdonald-stsci:
Adding GLS back to the ramp fitting module.  It works as before the JWST refactor that began in February 2021.  This is a replacement PR for https://github.com/spacetelescope/stcal/pull/51.  Please refer to the old PR for historical comments.

The ramp fitting unit tests in both JWST and STCAL pass or are skipped/xfailed as expected.

### DIFF
--- a/src/stcal/ramp_fitting/utils.py
+++ b/src/stcal/ramp_fitting/utils.py
@@ -724,102 +724,10 @@ def output_integ(slope_int, dq_int, effintim, var_p3, var_r3, var_both3,
     return integ_info
 
 
-'''
-# BEGIN remove GLS
-def gls_output_integ(model, slope_int, slope_err_int, dq_int):
-    """
-    For the GLS algorithm, construct the output integration-specific results.
-    Parameters
-    ----------
-    model : instance of Data Model
-        DM object for input
-
-    slope_int : ndarray
-        Data cube of weighted slopes for each integration, 3-D float
-
-    slope_err_int : ndarray
-        Data cube of slope errors for each integration, 3-D float
-
-    dq_int : ndarray
-        Data cube of DQ arrays for each integration, 3-D flag
-
-    Returns
-    -------
-    cubemod : Data Model object
-    """
-    # Suppress harmless arithmetic warnings for now
-    warnings.filterwarnings("ignore", ".*invalid value.*", RuntimeWarning)
-    warnings.filterwarnings("ignore", ".*divide by zero.*", RuntimeWarning)
-
-    cubemod = datamodels.CubeModel()
-    cubemod.data = slope_int
-    cubemod.err = slope_err_int
-    cubemod.dq = dq_int
-
-    # Reset the warnings filter to its original state
-    warnings.resetwarnings()
-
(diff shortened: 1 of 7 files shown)


nden:
@kmacdonald-stsci There are changes which don't belong to this PR. Could you rebase on main?

kmacdonald-stsci:
I rebased again and this seems to have corrected the "Files Changed" problem from the previous rebase.

src/stcal/ramp_fitting/utils.py
```diff
@@ -724,102 +724,10 @@ def output_integ(slope_int, dq_int, effintim, var_p3, var_r3, var_both3,
...
-
-    # Reset the warnings filter to its original state
-    warnings.resetwarnings()
-
-    cubemod.update(model)  # keys from input needed for photom step
-
-    return cubemod
-
-
-def gls_output_optional(model, intercept_int, intercept_err_int,
```
hbushouse, line 767: I don't see (in an obvious way) that `gls_output_integ` and `gls_output_optional` are replicated anywhere else in the code, so are they no longer needed?

tests/test_ramp_fitting_gls_fit.py
```diff
@@ -0,0 +1,922 @@
...
+    # These arguments don't make sense to me.
+    # np.testing.assert_allclose(actual, desired, rtol, atol, ...)
+    # out_slope=slopes[0][50, 50] = 5.782827
+    # check = (deltaDN1 + deltaDN2)/2.0 = 100.0
+    # WTF?
+    # np.testing.assert_allclose(out_slope, (deltaDN1 + deltaDN2)/2.0, 75.0, 1e-6)
+    np.testing.assert_allclose(out_slope, check, 75.0, 1e-6)
+
+
+@pytest.mark.skip(reason="GLS returns NaN's, but should return zeros.")
```
hbushouse, line 305: Do we expect that the problem of returning NaN's is or will be fixed in the new version of the code?

hbushouse (approved):
Looks good overall. Just a couple questions. Seems like quite a few of the unit tests are skipped for various reasons. Would be nice if someone could figure out how to solve the issues that are causing the problems.

src/stcal/ramp_fitting/utils.py
kmacdonald-stsci: They are no longer needed.  They were used when they code returned data models, but since the move to STCAL data models are no longer used internally, including as return values.

tests/test_ramp_fitting_gls_fit.py
kmacdonald-stsci: The code as it runs right now needs a lot of work.  It runs as it did before the refactor, but the GLS code before the refactor was not complete.  The GLS code still needs a lot of work to get it to completion.  The skipped tests are a guide for many of the things that need to be fixed.

nden:
@kmacdonald-stsci Add a `0.6.0` section in CHANGES.rst and add an entry.
Since this is the code before the refactor I'd say we should merge it as is and continue fixing issues.

nden (approved):

PR #70, opened 2021-12-09
JP-2346: Changed the 2 Group Ramp Special Case

kmacdonald-stsci:
Changed how `fit_2_group` functions.  It improperly assumed if group 0 was flagged, then the flag was `SATURATED`, which also invalidates group 1.  It is possible for group 0 to be flagged such that group 1 contains valid data to be used during ramp fitting.  Now the function properly uses valid data in group 1, when valid, even if group 0 is flagged with a non-`SATURATED` flag.

A unit test was added to test the various possibilities for a 2 group ramp.  All unit tests in JWST and STCAL pass.

This PR addresses the the following JIRA ticket.
https://jira.stsci.edu/browse/JP-2346

### DIFF
--- a/src/stcal/ramp_fitting/ols_fit.py
+++ b/src/stcal/ramp_fitting/ols_fit.py
@@ -21,13 +21,6 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-# ******************************************************************************
-# XXX - Debugging functions
-import sys; sys.path.insert(0, "/Users/kmacdonald/bin")
-from debug_funcs import *
-# ******************************************************************************
-
-
 def ols_ramp_fit_multi(
         ramp_data, buffsize, save_opt, readnoise_2d, gain_2d, weighting, max_cores):
     """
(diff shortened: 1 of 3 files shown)


src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1581,7 +1581,7 @@ def calc_slope(data_sect, gdq_sect, frame_time, opt_res, save_opt, rn_sect,
         of all ramps; later used when truncating arrays before output.
 
     remp_data : RampClass
-        The ramp data and metadata, specifically the relevant DQ flags.
+        The ramp data and metadata, specifically the relavent DQ flags.
```
hbushouse, line 1584: typo: relavent -> relevant

hbushouse (approved):
Looks OK to me. Just one minor typo in a comment.

hbushouse:
Needs 1 flake8 failure fixed.

dmggh (approved):
Looks good to me

nden:
Merging. Needs a change log in jwst.

PR #71, opened 2021-12-15
WIP fixes to twopoint difference

cshanahan1:
Several major fixes to the twopoint difference routine in jump detection.

1: According to the description of the algorithm, groups that have a 'difference ratio' larger than the 'threshold' are considered a jump. The code was incorrectly using the 'difference ratio' of the group with the largest positive first difference, rather than just the largest ratio. The largest positive first difference often, but not always, has the largest ratio as well. With this fix, the check for 'is the largest ratio above the threshold?' is now being done as intended.

A note: after the first CR is found and clipped, the same process is applied to all groups in a pixel where another CR is found, clipped, and 'ratio' is recomputed. Previously, even though 'ratio' is recomputed during each iteration, the array that sorts 'positive first diffs' was only computed once before the clipping and the 'ratio' values corresponding to the largest / next largest / so on values in this array were being chosen to look for subsequent CRs. What really should be done is: ratio is recomputed, the largest ratio of this new array is selected for the threshold comparison, and if it's above the threshold the original group corresponding to the index of this largest ratio should be flagged. Figuring out how to do this is what is holding me up right now with this - I am clipping pixels during each iteration, but I need to keep track of what index this corresponds to in the original non-clipped array. 

2. There are two functions that compute the median of first differences (which is used in the calculation of the 'difference ratio' therefore impacts what groups are flagged as jumps). The function 'get_clipped_median_array' is used in the initial check for a jump before iterating to find subsequent jumps. In this function, two issues were fixed. First, the median was incorrectly being rounded to the nearest integer value (e.g a median of -5.25542 became -5), which is now fixed. Second, when there were an even number of elements in the array, the two middle values were not being averaged correctly.

3. There was another issue in both clipped median calculating functions -'get_clipped_median_array' which is used in the initial check for jumps and then 'get_clipped_median_vector' which is used on each pixel after the initial jump is found. Both of these functions claim to calculate the clipped median of 'first differences'. In what I assume is a measure to increase efficiency, np.median is not used but rather since the array that sorts the first differences was already computed before these functions were called, the middle element(s) of this array of indexes is selected and then the 'first differences' array is indexed using these to find the median. There was an error in this - while 'first differences' was being passed to these function as the array to get the median of, the sorting array wasn't actually sorting 'first differences', it was sorting the absolute value of first differences (i.e 'positive first differences'). So, the middle element of this sorting array corresponds to the index containing the median of 'positive first differences', but then it was being used to index 'first differences'. I'm still not actually sure reading the description of the algorithm if the median of 'first differences' or 'positive first differences' should be used - it says verbatim 'Compute the clipped (dropping the largest difference) median of the first differences for each pixel.'. Does 'largest' here mean most positive, in which case it would be 'first differences', or largest in magnitude in which case it would be 'positive first differences'. I'm leaning toward the latter, but the former is what is in the current implementation. 

3. I will note that there is an outstanding issue with flagging jumps in 2-group integrations, but I have reverted that back to its original state and will fix that in another PR later since thats not as pressing.

### DIFF
--- a/src/stcal/jump/new_twopoint_difference.py
+++ b/src/stcal/jump/new_twopoint_difference.py
@@ -19,8 +19,7 @@ HUGE_NUM = np.finfo(np.float32).max
 
 import matplotlib.pyplot as plt
 
-
-def find_crs(data, group_dq, read_noise, normal_rej_thresh,
+def new_find_crs(data, group_dq, read_noise, normal_rej_thresh,
              two_diff_rej_thresh, three_diff_rej_thresh, nframes,
              flag_4_neighbors, max_jump_to_flag_neighbors,
              min_jump_to_flag_neighbors, dqflags):
@@ -157,6 +156,7 @@ def find_crs(data, group_dq, read_noise, normal_rej_thresh,
         ratio = np.abs(first_diffs - median_diffs[:, :, np.newaxis]) /\
             sigma[:, :, np.newaxis]
         ratio3d = np.reshape(ratio, (nrows, ncols, ndiffs))
+        sort_index_ratio = np.argsort(ratio3d)
 
         # Get the group index for each pixel of the largest non-saturated
         # group, assuming the indices are sorted. 2 is subtracted from ngroups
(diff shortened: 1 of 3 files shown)


hbushouse:
Superseded by #72

PR #72, opened 2022-01-06
Changes to jump detection 

cshanahan1:
This PR takes the place of #71 with a more final version on a new branch. I will update this description later. 

Several fixes to jump detection including:
1. fixes to the computation of the median that is used to compute jump thresholds
2. fixing an error that was causing the largest difference ratio to be missed when comparing to the threshold to classify jumps, resulting in many jumps being missed
3. restructuring code and adding comments that describe each step
4. fixes an issue with MIRI 3 & 4 group integrations (link PR / issue here)

### DIFF
--- a/src/stcal/jump/twopoint_difference.py
+++ b/src/stcal/jump/twopoint_difference.py
@@ -1,27 +1,11 @@
-"""
-Two-Point Difference method for finding outliers in a 4-D ramp data array.
-The scheme used in this variation of the method uses numpy array methods
-to compute first-differences and find the max outlier in each pixel while
-still working in the full 4-D data array. This makes detection of the first
-outlier very fast. We then iterate pixel-by-pixel over only those pixels
-that are already known to contain an outlier, to look for any additional
-outliers and set the appropriate DQ mask for all outliers in the pixel.
-This is MUCH faster than doing all the work on a pixel-by-pixel basis.
-"""
-
-import logging
 import numpy as np
 
-log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
 
-HUGE_NUM = np.finfo(np.float32).max
-
-
-def find_crs(data, group_dq, read_noise, normal_rej_thresh,
+def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
              two_diff_rej_thresh, three_diff_rej_thresh, nframes,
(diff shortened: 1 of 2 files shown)


hbushouse:
Do we know why the 1 unit test (in "CI / jwst tests with coverage") is still failing? Is that a legitimate different result?

hbushouse:
The one unit test that is still not passing is "test_nirspec_saturated_pix", which is supposed to verify that groups flagged as saturated don't get CR flags added to them (i.e. no outputs with DQ of 2+4=6). That part is working fine. The one difference is for pixel [2,2], which has SCI values of:
    
model.data[0, :, 2, 2] = [8.25666812e+05, -1.10471914e+05, 1.95755371e+02, 1.83118457e+03,
                              1.72250879e+03, 1.81733496e+03, 1.65188281e+03]

There's a large negative jump in group 2, which apparently used to get flagged as a CR by the old routine, but now it isn't anymore. This could be legit, in that it's no longer above the rejection threshold, given the fixes that've been made. My only concern is that it might also be indicating that something in the new routines is no longer detecting any negative jumps.

cshanahan1:
> Do we know why the 1 unit test (in "CI / jwst tests with coverage") is still failing? Is that a legitimate different result?


In this test, a test dataset is constructed and four pixels - ((1, 1), (2, 2), (3, 3), 4,4)) are given test scenarios to see if flagging for saturated pixels is done correctly. 

The tests for pixels (1, 1) and (3, 3) did not have different results - while the internal calculations changed based on changes to the code, the resulting pixels flagged as jumps ended up being the same.

for the test case of pixel (2, 2), 5 of the 7 groups in the test data are set to be saturated, meaning there is only one non saturated diff so no jumps should be flagged. previously, it was incorrectly flagging the 3rd group as a jump when NO groups should have been flagged, so the truth value of this test has changed. I kept this test but changed the truth to assert that NO groups are flagged in this case. 

pixel(4, 4) was the same case as (2, 2), where 5 out of 7 groups were saturated so no jumps should be flagged but they were being flagged. I removed this redundant example.

hbushouse (approved):
After a ridiculous amount of reviewing and testing, I hereby approve this PR.

tapastro (approved):
LGTM?

kmacdonald-stsci (approved):

PR #74, opened 2022-01-12
Jump step: fix median, add tests, add logging

cshanahan1:
Fixes a small bug in the computation of the median for pixels with two usable diffs and no flagged groups. Also adds tests that cover this median function for all possible inputs. Adds in logging statements for two point difference routine. 

Additional fixes for [JP-2206](https://jira.stsci.edu/browse/JP-2206).

### DIFF
--- a/src/stcal/jump/twopoint_difference.py
+++ b/src/stcal/jump/twopoint_difference.py
@@ -125,12 +125,10 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
         row2cr, col2cr = np.where(np.logical_and(ndiffs - num_unusable_groups == 2,
                                   max_ratio > two_diff_rej_thresh))
 
-        log.info(f'From highest outlier Two-point found {len(row4cr)} pixels \
-                 with at least one CR and at least five groups')
-        log.info(f'From highest outlier Two-point found {len(row3cr)} pixels \
-                 with at least one CR and four groups')
-        log.info(f'From highest outlier Two-point found {len(row2cr)} pixels \
-                 with at least one CR and three groups')
+        log_str = 'From highest outlier, two-point found {} pixels with at least one CR from {} groups.'
+        log.info(log_str.format(len(row4cr), 'five'))
+        log.info(log_str.format(len(row3cr), 'four'))
+        log.info(log_str.format(len(row2cr), 'three'))
 
         # get the rows, col pairs for all pixels with at least one CR
         all_crs_row = np.concatenate((row4cr, row3cr, row2cr))
--- a/src/stcal/jump/twopoint_difference.py
+++ b/src/stcal/jump/twopoint_difference.py
@@ -1,4 +1,8 @@
+import logging
 import numpy as np
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
 
 
 def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
@@ -80,6 +84,8 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
 
     for integ in range(nints):
 
+        log.info(f'Working on integration {integ + 1}:')
+
         # get data, gdq for this integration
         dat = dataa[integ]
         gdq_integ = gdq[integ]
@@ -118,6 +124,13 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
                                   max_ratio > three_diff_rej_thresh))
         row2cr, col2cr = np.where(np.logical_and(ndiffs - num_unusable_groups == 2,
                                   max_ratio > two_diff_rej_thresh))
+
+        log.info(f'From highest outlier Two-point found {len(row4cr)} pixels \
+                 with at least one CR and at least five groups')
+        log.info(f'From highest outlier Two-point found {len(row3cr)} pixels \
+                 with at least one CR and four groups')
+        log.info(f'From highest outlier Two-point found {len(row2cr)} pixels \
+                 with at least one CR and three groups')
 
         # get the rows, col pairs for all pixels with at least one CR
         all_crs_row = np.concatenate((row4cr, row3cr, row2cr))
@@ -263,8 +276,8 @@ def calc_med_first_diffs(first_diffs):
             return np.nanmedian(first_diffs[mask])
         elif num_usable_groups == 3:  # if 3, no clipping just return median
             return(np.nanmedian(first_diffs))
-        elif num_usable_groups == 2:  # if 2, return minimum
-            return(np.nanmin(np.abs(first_diffs)))
+        elif num_usable_groups == 2:  # if 2, return diff with minimum abs
+            return(first_diffs[np.nanargmin(np.abs(first_diffs))])
         else:
             return(np.nan)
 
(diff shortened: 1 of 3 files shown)


src/stcal/jump/twopoint_difference.py
```diff
@@ -119,6 +125,13 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
         row2cr, col2cr = np.where(np.logical_and(ndiffs - num_unusable_groups == 2,
                                   max_ratio > two_diff_rej_thresh))
 
+        log.info(f'From highest outlier Two-point found {len(row4cr)} pixels \
+                 with at least one CR and at least four groups')
```
hbushouse, line 129: I realized while doing debugging in the last few days that these log messages are incorrect or at least deceptive, in that they refer to the number of **groups**, when in fact the quote numbers are actually the number of **differences** being worked with. It's quite impossible to do any jump detection on ramps that only have 2 **groups**, but you can do it if you have 2 **differences** available. So rather than changing the wording in the messages to say "differences" instead of "groups", let's just change the numbers that are quoted (bump them all up by 1). I.e. first message is regarding **five** groups, next one is **four** groups, and last one is **three** groups.

src/stcal/jump/twopoint_difference.py
```diff
@@ -291,7 +304,7 @@ def calc_med_first_diffs(first_diffs):
     row2, col2 = np.where(num_usable_groups == 2)  # locations of >= 4 usable group pixels
     if len(row2) > 0:
```
hbushouse, line 305: I'm wondering if we should enclose the previous blocks of code that work with row4,col4 and row3,col3 in IF-blocks, just like we do here for row2,col2?  Based on one quick test, it appears that when row4=row3=0, the lines above don't do any damage, but it just seems safer to me to not even call lines 293-296 and 300-301 if row4=0 or row3=0.

tests/test_twopoint_difference.py
```diff
@@ -833,3 +833,75 @@ def test_first_last_3group(setup_cube):
     assert outgdq[0, 0, 0, 0] == 0
     assert outgdq[0, 2, 0, 0] == DQFLAGS['DO_NOT_USE']
     assert outgdq[0, 1, 0, 0] == 0
+
+
+def test_median_func():
+
+    """
+      Test the function `calc_med_first_diffs` that computes median of pixels
```
hbushouse, line 841: Minor nit: missing a period at the end this sentence.

src/stcal/jump/twopoint_difference.py
```diff
@@ -291,7 +304,7 @@ def calc_med_first_diffs(first_diffs):
...
         two_slice = first_diffs[:, row2, col2]
-        median_diffs[row2, col2] = np.nanmin(two_slice)  # add median to return arr for these pix
+        median_diffs[row2, col2] = np.nanmin(np.abs(two_slice), axis=0)  # add med. to return arr
```
hbushouse, line 307: Why do we want to use `np.abs()` here? This function is being used to find the median of the first differences, in order to guesstimate an approximate real countrate for a given pixel. If we eliminate negatives, we're going to bias the result to always be positive. For example, in a pixel that has no real source signal in it, the first differences will bounce around an average of zero (since zero is the real countrate), with an amplitude equal to the read noise, e.g. 10, -7, -2, 5, -1, 8, etc. We want the median to come out as close as possible to the real countrate, which in this case is zero, but if we take the abs value of all the entries, it'll come out significantly non-zero.

tests/test_twopoint_difference.py
```diff
@@ -833,3 +833,75 @@ def test_first_last_3group(setup_cube):
...
+    assert calc_med_first_diffs(arr)[0, 0] == 2
+    # 1d, with nans
+    arr = np.array([1., 2., 3., np.nan, np.nan])
+    assert calc_med_first_diffs(arr) == 2
+    # 3d, with nans
+    arr = np.zeros(5 * 2 * 2).reshape(5, 2, 2)
+    arr[:, 0, 0] = np.array([1., 2., 3., np.nan, np.nan])
+    assert calc_med_first_diffs(arr)[0, 0] == 2
+
+    # # single pix with exactly 2 good diffs, should return the element with the minimum abs val
```
hbushouse, line 894: These tests may need to change, if we remove the use of abs value above.

src/stcal/jump/twopoint_difference.py
cshanahan1: For the median, when there are only two diffs left, the one with the smallest magnitude (i.e the smallest jump in any direction) should be chosen. So if they were [-10, 2], I would select 2 because it represents a smaller jump. You're right thought that this isn't doing the correct thing - should use np.abs() to find the location of the minimum, but then return the value with the sign attached.

tests/test_twopoint_difference.py
cshanahan1: The only tests that would change by changing the use of np.abs on line 307 would be those for the two-diff examples, and I don't think they would have to change in this case because the diff with the lowest absolute value is already positive. ( I set the test up with arr = [1, -2], so the min(arr) is the same as min(abs(arr)) in this case)

src/stcal/jump/twopoint_difference.py
cshanahan1: that sounds like a good thing to do, i'll go ahead and fix that

src/stcal/jump/twopoint_difference.py
cshanahan1: I noticed this and wanted to change the logging messages, but I wasn't sure if when you were comparing old version to the new version if you wanted the exact same logging to appear for consistency. If you don't need the messages to be the same, I will go ahead and scrap these and put some different more informative messages in (including logging during/after the iteration for additional CRs). What do you think about this?

hbushouse: Works for me.

src/stcal/jump/twopoint_difference.py
hbushouse: Ah, right. I was forgetting about the fact that this is only for the special case of having 2 diffs left to compare. In that case it makes sense, in terms of returning the one that has the smallest deviation.

hbushouse (commented):
Other than the message formatting issue, this looks good to me. I've tested it against a couple regression datasets and see only a few pixels having changes in their CR flags (as expected).

Needs a change log entry.

ddavis-stsci (approved):
LGTM, I also ran it against the romancal jump unit tests and they pass as well.

src/stcal/jump/twopoint_difference.py
```diff
@@ -119,6 +125,13 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
         row2cr, col2cr = np.where(np.logical_and(ndiffs - num_unusable_groups == 2,
                                   max_ratio > two_diff_rej_thresh))
 
+        log.info(f'From highest outlier Two-point found {len(row4cr)} pixels \
+                 with at least one CR and at least five groups')
```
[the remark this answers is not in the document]
cshanahan1: Did that do the trick?

cshanahan1:
I also updated the changelog for the next release (0.6.0) which will be done after this is merged.

src/stcal/jump/twopoint_difference.py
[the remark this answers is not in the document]
hbushouse: Yep, looks good now.

hbushouse (approved):

hbushouse:
Ta-Da!  Clean tests. I'm merging.

PR #75, opened 2022-02-11
JP-2071: Added a boolean switch to turn off calculations for one group ramp

kmacdonald-stsci:
This PR resolves https://github.com/spacetelescope/jwst/issues/6013

A boolean switch has been added to STCAL ramp fitting to suppress ramps that have only one good group.  These ramps will be treated as having zero good groups.  This is done by finding the one good group in the ramp, then setting the DQ flag for that ramp to DO_NOT_USE.

Two tests have been added to test this new feature.

### DIFF
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -144,12 +144,10 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d, algorithm,
         Object containing optional GLS-specific ramp fitting data for the
         exposure
     """
-    if suppress_one_group:
-        log.info("Ramps with one good group suppressed as no good group.")
-        if model.data.shape[1] == 1:
-            # Suppressing one ramp groups and having data with only one group
-            # means there is no work to do.
-            return None, None, None, None
+    if suppress_one_group and model.data.shape[1] == 1:
+        # One group ramp suppression should only be done on data with
+        # ramps having more than one group.
+        suppress_one_group = False
 
     # Create an instance of the internal ramp class, using only values needed
     # for ramp fitting from the to remove further ramp fitting dependence on
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -26,7 +26,7 @@ log.setLevel(logging.DEBUG)
 BUFSIZE = 1024 * 300000  # 300Mb cache size for data section
 
 
-def create_ramp_fit_class(model, dqflags=None):
+def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
     """
     Create an internal ramp fit class from a data model.
 
@@ -37,6 +37,10 @@ def create_ramp_fit_class(model, dqflags=None):
 
     dqflags : dict
         The data quality flags needed for ramp fitting.
+
+    suppress_one_group : bool
+        Find ramps with only one good group and treat it like it has zero good
+        groups.
 
     Return
     ------
@@ -70,13 +74,13 @@ def create_ramp_fit_class(model, dqflags=None):
     ramp_data.start_row = 0
     ramp_data.num_rows = ramp_data.data.shape[2]
 
-    # XXX self.suppress_one_group_ramps = False
+    ramp_data.suppress_one_group_ramps = suppress_one_group
 
     return ramp_data
 
 
-def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
-             algorithm, weighting, max_cores, dqflags):
+def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d, algorithm,
+             weighting, max_cores, dqflags, suppress_one_group=False):
     """
     Calculate the count rate for each pixel in all data cube sections and all
     integrations, equal to the slope for all sections (intervals between
@@ -120,6 +124,10 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
     dqflags : dict
         A dictionary with at least the following keywords:
         DO_NOT_USE, SATURATED, JUMP_DET, NO_GAIN_VALUE, UNRELIABLE_SLOPE
+
+    suppress_one_group : bool
+        Find ramps with only one good group and treat it like it has zero good
+        groups.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -212,9 +229,46 @@ def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
         nframes = ramp_data.nframes
         readnoise_2d *= gain_2d / np.sqrt(2. * nframes)
 
+        # Suppress one group ramps, if desired.
+        if ramp_data.suppress_one_group_ramps:
+            suppress_one_group_ramps(ramp_data)
```
dmggh, line 234: Within this call to suppress_one_group_ramps(ramp_data), intdq marks the groups of pixels having only a single group. It's not obvious to me how that information gets passed back into ramp_fit_data( ) - via ramp_data  ?

dmggh (commented):
I'm still reviewing, but I have a few comments/questions so far.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -134,6 +144,13 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
         Object containing optional GLS-specific ramp fitting data for the
         exposure
     """
+    if suppress_one_group:
+        log.info("Ramps with one good group suppressed as no good group.")
+        if model.data.shape[2] == 1:
```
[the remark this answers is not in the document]
kmacdonald-stsci: I changed this to be `if model.data.shape[1] == 1`.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -212,9 +229,46 @@ def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
...
             ramp_data, buffsize, save_opt, readnoise_2d, gain_2d, weighting, max_cores)
         gls_opt_info = None
 
     return image_info, integ_info, opt_info, gls_opt_info
+
+
+def suppress_one_group_ramps(ramp_data):
+    """
+    Finds one group ramps in each integration and suppresses them, i.e. turns
+    the in to zero group ramps.
```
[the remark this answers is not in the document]
kmacdonald-stsci: Fixed this typo.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: The `intdq` array is essentially a pointer to a subarray of `ramp_data.groupdq`.  Any value change to `intdq` is a value change in `ramp_data.groupdq`.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -212,9 +229,46 @@ def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
...
+    nints, ngroups, nrows, ncols = dq.shape
+    for k in range(nints):
+        intdq = dq[k, :, :, :]
+        good_groups = np.zeros(intdq.shape, dtype=int)
+
+        # Compute the number of good groups in each ramp
+        good_groups[intdq == 0] = 1
+        ngood_groups = good_groups.sum(axis=0)
+
+        # For each pixel that has only one good group, mark that groups as
```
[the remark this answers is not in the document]
kmacdonald-stsci: Fixed this typo.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -134,10 +144,17 @@ def ramp_fit(model, buffsize, save_opt, readnoise_2d, gain_2d,
         Object containing optional GLS-specific ramp fitting data for the
         exposure
     """
+    if suppress_one_group:
+        log.info("Ramps with one good group suppressed as no good group.")
+        if model.data.shape[1] == 1:
+            # Suppressing one ramp groups and having data with only one group
+            # means there is no work to do.
```
hbushouse, line 151: Actually, I just noticed that in the Cal WG request for this change they also said that they did NOT want this to happen when NGROUPS=1. I forgot to put that into the description of the implementation ticket. So there needs to be a little additional logic, such that when NGROUPS=1 (`model.data.shape[1] = 1`) this new switch gets ignored and slopes are still calculated from the 1 available group (at least wherever that 1 group is good). So it's only when NGROUPS>1 that we want this switch to have an effect.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -212,9 +229,47 @@ def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
...
+    """
+    dq = ramp_data.groupdq
+    nints, ngroups, nrows, ncols = dq.shape
+
+    for k in range(nints):
+        intdq = dq[k, :, :, :]
+        good_groups = np.zeros(intdq.shape, dtype=int)
+
+        # Compute the number of good groups in each ramp
+        good_groups[intdq == 0] = 1
```
hbushouse, line 262: I'm not sure that relying on `intdq` being exactly zero to indicate a good group is entirely appropriate. It's possible that for some flag values to be present (non-zero) that are to be treated as information only (e.g. a warning), but does not render the data unusable. It should be only cases where a group has the DO_NOT_USE flag set that they should be considered completely bad. This assumes that any group that has either the SATURATED or JUMP flags set also has DO_NOT_USE set. We would need to check that.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -212,9 +229,47 @@ def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
...
+        ngood_groups = good_groups.sum(axis=0)
+
+        # For each pixel that has only one good group, mark those groups as
+        # DO_NOT_USE, making the ramp effectively zero good groups.
```
hbushouse, line 266: Is this DO_NOT_USE setting only being applied to some internal data array, for use solely within the slope calculations? We don't want those flags to get set in the original input data to the step (i.e. we don't want to make them permanent).

tests/test_ramp_fitting.py
```diff
@@ -503,6 +638,12 @@ def print_slope_data(slopes):
     print(sdata)
 
 
+def print_slope_dq(slopes):
```
hbushouse, line 641: Are all of these various print functions used for interactive debugging or what? I don't see them being used in the unit tests.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: This will require discussion because this is how segments are computed at

tests/test_ramp_fitting.py
kmacdonald-stsci: I use these functions to debug and develop tests.  I have added a comment that all functions below the comment are debugging/development functions.

kmacdonald-stsci:
This PR does not correctly address the feature request.

PR #79, opened 2022-03-25
JP-2525: jump step flagging saturated groups as jumps

cshanahan1:
This PR fixes an issue where saturated pixels were also being flagged as jumps. While the initial pass to find jumps was correctly ignoring saturated and donotuse pixels, when the flag to set the DQ for neighboring pixels was set, if the neighboring pixels contained these flags the jump flag was simply added to them. Now, if a neighboring pixel is saturated or donotuse, it will not also be flagged as a jump.

Fixes [JP-2525](https://jira.stsci.edu/browse/JP-2525)

### DIFF
--- a/src/stcal/jump/twopoint_difference.py
+++ b/src/stcal/jump/twopoint_difference.py
@@ -82,6 +82,11 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
     row_above_gdq = np.zeros((nints, ngroups, ncols), dtype=np.uint8)
     row_below_gdq = np.zeros((nints, ngroups, ncols), dtype=np.uint8)
 
+    # get dq flags for saturated, donotuse, jump
+    sat_flag = dqflags["SATURATED"]
+    dnu_flag = dqflags["DO_NOT_USE"]
+    jump_flag = dqflags["JUMP_DET"]
+
     for integ in range(nints):
 
         log.info(f'Working on integration {integ + 1}:')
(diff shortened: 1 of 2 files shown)


src/stcal/jump/twopoint_difference.py
```diff
@@ -209,32 +209,49 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
                     # find_jumps to use when it reconstructs the full group dq
                     # array from the slices.
                     if cr_row[j] != 0:
-                        gdq[integ, cr_group[j], cr_row[j] - 1, cr_col[j]] =\
-                            np.bitwise_or(gdq[integ, cr_group[j], cr_row[j] -
-                                          1, cr_col[j]], dqflags["JUMP_DET"])
+
+                        if (gdq[integ, cr_group[j], cr_row[j] - 1, cr_col[j]] \
```
hbushouse, line 213: Should probably add a comment in the code here to indicate that we only want to flag neighbors if they aren't saturated and do_not_use.

hbushouse (approved):
Looks good. Just one minor comment.

hbushouse:
How can we kick-start the CI tests?

ddavis-stsci:
I'm seeing "Workflow runs completed with no jobs"
Was something updated?

cshanahan1:
i'm not sure why the CI paused, that's never happened before. i'll try to figure out what's going on.

hbushouse:
I tried to get everything to restart by making a minor (benign) edit to the change log, but they still haven't started. Some problem with Github Actions?

cshanahan1:
@hbushouse ha, i just did the same thing. it doesnt look like that kicked them off.

hbushouse:
Oooh, now they're running!

PR #81, opened 2022-03-31
JP-374: Use ZEROFRAME for Saturated Ramps.

kmacdonald-stsci:
Resolves [JP-374](https://jira.stsci.edu/browse/JP-374)

Adding processing of the ZEROFRAME array to all necessary steps to be properly used in the ramp fitting step.  This PR will need to be merged before the corresponding PR in JWST can be merged.

The ZEROFRAME array contains the zeroeth frame for a ramp.  If a ramp is found to be fully saturated, but the corresponding data in the ZEROFRAME array is good data, then when performing ramp fitting use the ZEROFRAME data in place of the zeroeth group for that saturated ramp.  To ensure the data is usable in the ramp fitting step, all applicable steps must also be applied to the ZEROFRAME array, in addition to the SCI array.  There is no corresponding ZEROFRAME data quality array, so if any data become unusable, i.e., should be flagged, then the data in the ZEROFRAME is set to 0.  In ramp fitting, when searching for saturated ramps for which to use the ZEROFRAME data, the ZEROFRAME data must be non-zero to be considered "good" and used for the zeroeth group for that ramp.


src/stcal/linearity/linearity.py
```diff
@@ -24,35 +25,105 @@ def linearity_correction(data, gdq, pdq, lin_coeffs, lin_dq, dqflags):
...
+        zlin_dq = lin_dq.copy()
+
+    # Do linear correction on SCI data
+    data, new_pdq = linearity_correction_branch(
+        data, gdq, pdq, lin_coeffs, lin_dq, dqflags, False)
+
+    zdata = None  # zframe needs to be returned, so initialize it to None.
+    if zframe is not None:
+        # Do linear correction on ZEROFRAME
+        # Zero ZEROFRAME values indicates bad data, so remember where this
```
hbushouse, line 71: Where did you find out that a 0 value in the zeroframe array indicates bad data? Was that in some of the documentation provided to us from the NIRCam team? I don't remember seeing that anywhere.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -47,6 +47,10 @@ def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
     ramp_data : ramp_fit_class.RampData
         The internal ramp class.
     """
+    if not suppress_one_group and hasattr(model, "zeroframe"):
```
hbushouse, line 50: Does "hasattr()" actually work as needed here? Because zeroframe is defined in the RampModel, the attribute may always exist regardless of whether or not there's any data in the array. You might need to actually check for something like non-zero size of the zeroframe array.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -278,3 +279,67 @@ def suppress_one_group_saturated_ramps(ramp_data):
...
+        whs_cols = wh_sat[1]
+        for n in range(len(whs_rows)):
+            row = whs_rows[n]
+            col = whs_cols[n]
+
+            # For ramps completely saturated look for data in the ZEROFRAME
+            # that is non-zero.  If it is non-zero, replace group zero in the
+            # ramp with the data in ZEROFRAME.
+            if model.zeroframe[integ, row, col] != 0:
+                model.data[integ, 0, row, col] = model.zeroframe[integ, row, col]
```
hbushouse, line 320: Not completely sure that this direct replacement of group zero with zeroframe will eventually give the correct slope value downstream. For cases where only the first group is good, the slope is computed by simply dividing the data value by the group integration time. But because the groups can (and usually will) contain data that've been averaged over several frames (this averaging is done on-board before the data are downlinked), the effective group integration time will be longer than the effective integration time of the zeroframe by itself. Hence the slope may get computed using the wrong integration time.

src/stcal/linearity/linearity.py
hbushouse: Oh wait, now that I've reviewed your updates to the saturation step it looks like you're setting saturated locations to zero in the zeroframe array. Never mind. Might want to mention this briefly in the comments here, that saturated values in zeroframe have been reset to zero in the saturation step.

src/stcal/linearity/linearity.py
```diff
@@ -24,35 +25,105 @@ def linearity_correction(data, gdq, pdq, lin_coeffs, lin_dq, dqflags):
...
+
+    return data, new_pdq, zdata
+
+
+def linearity_correction_branch(
+        data, gdq, pdq, lin_coeffs, lin_dq, dqflags, zframe):
+    """
+    Parameters
+    ----------
+    data : `np.array`
```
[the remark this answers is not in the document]
kmacdonald-stsci: Fixed.

src/stcal/linearity/linearity.py
kmacdonald-stsci: I expanded this comment to explain why zero values indicate bad data in the `ZEROFRAME` array.

src/stcal/linearity/linearity.py
```diff
@@ -24,35 +25,105 @@ def linearity_correction(data, gdq, pdq, lin_coeffs, lin_dq, dqflags):
...
+
+    gdq : ndarray
+        The 4D roup dq array.
+
+    pdq : ndarray
+        The 2D pixel dq array.
+
+    lin_coeffs : ndarray.
+        The 3D array containing pixel-by-pixel linearity coefficient
+        values or each term in the polynomial fit.
```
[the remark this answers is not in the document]
kmacdonald-stsci: Fixed.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Before I new about the flag `model.meta.exposure.zero_frame` I used `hasattr`.  I thought I replaced all uses of `hasattr` with `model.meta.exposure.zero_frame`.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Timing is not handled until the computation of final slopes.  The slope of each integration, not divided by time, is used till the final slope computation.  The final slopes are computed by dividing the integration slope by `effintim`.  This variable is set by `effintim = (nframes + groupgap) * frame_time`.

src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1009,22 +963,109 @@ def ramp_fit_slopes(ramp_data, gain_2d, readnoise_2d, save_opt, weighting):
...
+        Contains the data needed to compute the median rates.
+
+    median_rates : ndarray
+        The computed median rates for each pixel.
+    """
+    nints, ngroups, nrows, ncols = ramp_data.data.shape
+    imshape = (nrows, ncols)
+
+    group_time = ramp_data.group_time
+    frame_time = ramp_data.group_time
```
dmggh, line 993: Should the RHS be ramp_data.frame_time ?

src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1447,6 +1490,8 @@ def ramp_fit_overall(
     log.debug('The execution time in seconds: %f', tstop - tstart)
 
     # Compute the 2D variances due to Poisson and read noise
+    sum_in_var_p2 = s_inv_var_p3.sum(axis=0)
+    wh_zero = np.where(sum_in_var_p2 == 0.)
```
dmggh, line 1494: I don't see that wh_zero is used.

dmggh (changes requested):
Just a couple (probable) typos.

src/stcal/ramp_fitting/ols_fit.py
kmacdonald-stsci: Yes.  I fixed that.

src/stcal/ramp_fitting/ols_fit.py
kmacdonald-stsci: These two lines were debugging lines of code that I was using for analysis.  I will remove them.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Timing is now properly computed using frame time for `ZEROFRAME` data, rather than group time.

hbushouse (approved):

PR #83, opened 2022-04-22
JP-2556: Handle charge spilling out of saturated pixels into neighbors

cshanahan1:
Pixels adjacent to saturated pixels are now also flagged as saturated. By default, every pixel that touches the saturated pixels will also be flagged. The parameter 'n_pix_grow_sat' controls how many pixels out are also flagged, and this can also be set to 0.

### DIFF
--- a/src/stcal/saturation/saturation.py
+++ b/src/stcal/saturation/saturation.py
@@ -1,12 +1,15 @@
 import numpy as np
 import logging
 
+import copy
+from scipy import ndimage
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
 def flag_saturated_pixels(data, gdq, pdq, sat_thresh, sat_dq, atod_limit,
-                          dqflags):
+                          dqflags, n_pix_grow_sat=1):
     """
     Short Summary
     -------------
@@ -38,6 +41,10 @@ def flag_saturated_pixels(data, gdq, pdq, sat_thresh, sat_dq, atod_limit,
     dqflags: dict
         A dictionary with at least the following keywords:
         DO_NOT_USE, SATURATED, AD_FLOOR, NO_SAT_CHECK
+
+    n_pix_grow_sat : int
+        Number of pixels that each flagged saturated pixel should be 'grown',
+        to account for charge spilling. Default is 1.
 
 
     Returns
@@ -84,6 +91,15 @@ def flag_saturated_pixels(data, gdq, pdq, sat_thresh, sat_dq, atod_limit,
             # for A/D floor, the flag is only set of the current plane
             np.bitwise_or(gdq[ints, group, :, :], flaglowarray,
                           gdq[ints, group, :, :])
+
+            # now, flag any pixels that border saturated pixels (not A/D floor pix)
+            if n_pix_grow_sat > 0:
+                gdq_slice = copy.copy(gdq[ints, group, :, :]).astype(int)
+                only_sat = np.bitwise_and(gdq_slice, saturated).astype(np.uint8)
+                box_dim = (n_pix_grow_sat * 2) + 1
+                struct = np.ones((box_dim, box_dim)).astype(bool)
+                dialated = ndimage.binary_dilation(only_sat, structure=struct).astype(only_sat.dtype)
+                gdq[ints, group, :, :] = np.bitwise_or(gdq[ints, group, :, :], (dialated * saturated))
 
     n_sat = np.any(np.any(np.bitwise_and(gdq, saturated), axis=0), axis=0).sum()
     log.info(f'Detected {n_sat} saturated pixels')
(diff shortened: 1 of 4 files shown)


cshanahan1:
not sure why jwst tests are failing, they run when i run them myself. looks like maybe it can't access a url? 

roman tests are failing as expected, i opened a pr in romancal to turn off the additional flagging so they'll pass once that's merged.

nden:
@cshanahan1 The jwst test failures are unrelated. Is it ready for a review otherwise?

cshanahan1:
@nden yes, ready for review

tests/test_saturation.py
```diff
@@ -8,6 +8,11 @@
 
 from stcal.saturation.saturation import flag_saturated_pixels
 
+# dictionary with required DQ flags
+DQFLAGS = {'DO_NOT_USE': 1, 'SATURATED': 2, 'AD_FLOOR': 64,
```
hbushouse, line 12: Might be safer to load dqflags.py and use the mnemonics from there. Hardcoding the values here could lead to problems if the dqflags definitions ever get changed in the future.

cshanahan1: just looked through dqflags.py, and im not sure i understand the purpose of these functions

wouldn't i still have to hard code a dictionary? i'm looking at the docstrings and it seems like i'd still have to define the exact dictionary above. the functions being tested also require a dictionary of dqflags to be passed in

nden: They are coded in 
```
from jwst.datamodels.dqflags import pixel
```

hbushouse: Or in this case you might need to import the "group" definitions, as opposed to "pixel". But they might be copies of one another for the flags of interest here.

cshanahan1: im pretty sure they are the same for pixel and group. also, jwst is not a dependency of stcal right so i shouldn't import dqflags from there?

src/stcal/saturation/saturation.py
```diff
@@ -85,6 +92,16 @@ def flag_saturated_pixels(data, gdq, pdq, sat_thresh, sat_dq, atod_limit,
...
                           gdq[ints, group, :, :])
 
+            # now, flag any pixels that border saturated pixels (not A/D floor pix)
+            if n_pix_grow_sat > 0:
+                gdq_slice = copy.copy(gdq[ints, group, :, :]).astype(int)
+                only_sat = np.bitwise_and(gdq_slice, saturated).astype(np.uint8)
+                box_dim = (n_pix_grow_sat * 2) + 1
+                struct = np.ones((box_dim, box_dim)).astype(bool)
+                dialated = ndimage.binary_dilation(only_sat, structure=struct).astype(only_sat.dtype)
+                gdq[ints, group, :, :] += (dialated * saturated)  # add back to original dq array
```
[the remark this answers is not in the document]
cshanahan1: done

tests/test_saturation.py
cshanahan1: all the tests in STCAL have a hard coded dictionary of dq flags at the top of the file, and all of them happen to use JWST's dqflags. should we do this a different way for all of the tests in another PR and just merge this as is?

nden: Right. Sorry, I confused you. The dqflags are passed from jwst (and romancal) in the top level calling routine.
https://github.com/spacetelescope/stcal/blob/main/src/stcal/jump/jump.py#L17

nden (approved):
LGTM

src/stcal/saturation/saturation.py
```diff
@@ -85,6 +92,15 @@ def flag_saturated_pixels(data, gdq, pdq, sat_thresh, sat_dq, atod_limit,
             np.bitwise_or(gdq[ints, group, :, :], flaglowarray,
                           gdq[ints, group, :, :])
 
+            # now, flag any pixels that border saturated pixels (not A/D floor pix)
+            if n_pix_grow_sat > 0:
+                gdq_slice = copy.copy(gdq[ints, group, :, :]).astype(int)
+                only_sat = np.bitwise_and(gdq_slice, saturated).astype(np.uint8)
+                box_dim = (n_pix_grow_sat * 2) + 1
+                struct = np.ones((box_dim, box_dim)).astype(bool)
+                dialated = ndimage.binary_dilation(only_sat, structure=struct).astype(only_sat.dtype)
```
hbushouse, line 101: I like this `binary_dilation` method (never seen it before). Is it possible to use it in a mode where you would send it the entire 4D groupdq array after the initial round of flagging is complete (i.e. the loops over groups and integrations is complete) and have it only expand the flags in the 3rd and 4th axes (i.e. image rows and cols)? That way it could be called just once, instead of within the loops here.

hbushouse (approved):
Looks good. Just one question about a potential (but unnecessary) change. Can you tell what kind of effect this has on processing speed, i.e. does it significantly slow down the step at all?

hbushouse:
Does this need a change to the top-level `saturation_step.py` module in the jwst package to add `n_pix_grow_sat` as user-settable parameter? I'm guessing users might want control of this for off-line processing.

PR #85, opened 2022-05-01
Changes to restore multiprocessing and unit tests

mwregan2:
This PR restores the multiprocessing capability to the jump and adds back the unit tests for the twopoint_difference.py. Note that there is a parallel PR for the JWST repo that is required for this to work.

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -5,7 +5,7 @@ import numpy as np
 from . import twopoint_difference as twopt
 from . import constants
 
-import multiprocessing  # 05/18/21: TODO- commented out now; reinstate later
+import multiprocessing
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
@@ -13,7 +13,7 @@ log.setLevel(logging.DEBUG)
 
 def detect_jumps(frames_per_group, data, gdq, pdq, err,
                  gain_2d, readnoise_2d, rejection_thresh,
-                 three_grp_thresh, four_grp_thresh, max_jump_to_flag_neighbors,
+                 three_grp_thresh, four_grp_thresh, max_cores, max_jump_to_flag_neighbors,
                  min_jump_to_flag_neighbors, flag_4_neighbors, dqflags):
     """
     This is the high-level controlling routine for the jump detection process.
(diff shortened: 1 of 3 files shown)


kmacdonald-stsci (commented):
Besides style failures it looks good.

kmacdonald-stsci (approved):

nden:
@mwregan2 I see the tests in this package are in stcal/tests, the top directory and there're tests for two point differences. I did not check if they are the same tests though.

Also did you compare results from processing a data set with multiprocessing on and off. It was previously disabled because it gave different results.

src/stcal/jump/jump.py
```diff
@@ -129,33 +129,45 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
     # Multiprocessing has been disabled for now, so the nunber of slices
     # is here set to 1. I'm leaving the related code in to ease the eventual
     # re-enablement of this code.
-    n_slices = 1
-
-    yinc = int(n_rows / n_slices)
+    print('new stcal jump')
```
hbushouse, line 132: can't have print statements in production code. Either remove or replace with log messages.

mwregan2:
The unit tests are different from what is in the JWST repo. These test the twopoint_difference routine.

mwregan2:
Nadia,
  Are there tests that show the differences? That would really help in finding the problem. I’ll add some tests to see if I can find any differences.
Mike

From: Nadia Dencheva ***@***.***>
Date: Monday, May 2, 2022 at 10:13 AM
Subject: Re: [spacetelescope/stcal] Changes to restore multiprocessing and unit tests (PR #85)


@mwregan2<https://urldefense.com/v3/__https:/github.com/mwregan2__;!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!w-8rwDIUBDUhrQj6B020MyNPewMTKxPOgD15X_qBOAWpcFFeNzA2OdL9PvO04ucuNSUgmdtrQWtbKAvsAup9y_4$> I see the tests in this package are in stcal/tests, the top directory and there're tests for two point differences. I did not check if they are the same tests though.

Also did you compare results from processing a data set with multiprocessing on and off. It was previously disabled because it gave different results.

nden:
@mwregan2 Sorry for the late reply. I see you added a test in jwst comparing single vs multiprocessing run and it passes. 

[test_single_CR_neighbor_flag](https://github.com/spacetelescope/jwst/blob/4316c729f5c0f3c1734feac1a4ec9dad327498a5/jwst/jump/tests/test_detect_jumps.py#L737) gives different results if run in multiprocessing model. I don't know the details of the algorithm to judge why one is passing and the other fails.

mwregan2:
I fixed the issue with multiprocessing giving different results. This was caused by the code changing the input model and then the same model being sent back in for the second call. This could not happen in the operational system.
  Even so, I fixed the problem and added a test case that fails without the fix. I just needed to have a new variable for the updated gdq array.

nden:
Superseded by #87

PR #87, opened 2022-05-16
JP-2581: fix to allow multiprocessing for jump

cshanahan1:
Fixes an issue in jump where the input gdq arrays were being modified which was causing tests to fail in JWST when the step was being run in multiprocessing mode. If multiple processes are run, the gdq/data arrays are copied before `find_crs` and not again within the function (which in this mode is running on slices).

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -13,7 +13,7 @@ log.setLevel(logging.DEBUG)
 
 def detect_jumps(frames_per_group, data, gdq, pdq, err,
                  gain_2d, readnoise_2d, rejection_thresh,
-                 three_grp_thresh, four_grp_thresh, max_jump_to_flag_neighbors,
+                 three_grp_thresh, four_grp_thresh, max_cores, max_jump_to_flag_neighbors,
                  min_jump_to_flag_neighbors, flag_4_neighbors, dqflags):
     """
     This is the high-level controlling routine for the jump detection process.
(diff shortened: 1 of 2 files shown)


src/stcal/jump/jump.py
```diff
@@ -164,6 +156,38 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
...
+
+        # must copy arrays here, find_crs will make copies but if slices
+        # are being passed in for multiprocessing then the original gdq will be
+        # modified unless copied beforehand
+        gdq = gdq.copy()
+        data = data.copy()
+        copy_arrs = False   # we dont need to copy arrays again in find_crs
+
+        for i in range(n_slices - 1):
+            print('setting up slices')
```
hbushouse: Either remove or change to a log message

src/stcal/jump/twopoint_difference.py
```diff
@@ -68,8 +72,11 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
     """
 
     # copy data and group DQ array
-    dataa = dataa.copy()
-    gdq = group_dq.copy()
+    if copy_arrs:
+        dataa = dataa.copy()
+        gdq = group_dq.copy()
+    else:
+        group_dq = gdq
```
hbushouse, line 79: Is this expression backwards? The input to the function is `group_dq`, while `gdq` is used for subsequent processing below. So it should be `gdq = group_dq` (as above, but without the copy operation).

cshanahan1: yes, you're right

hbushouse (approved):
Looks good now.

Needs a change log entry and should we worry about the failing romancal CI test? Or is that due to unrelated changes in the function signatures from saturation step updates?

cshanahan1:
@hbushouse I just opened a PR in jwst to fix some of the tests there and point to this branch. once that's approved i'll change the stcal version back so it can be merged

nden:
@cshanahan1 Do we need to add the same parameter to romancal? If so please submit a PR.

hbushouse:
Are we OK to merge this now on the stcal side, so that we can get a new release out, before we merge things on the jwst side?

PR #89, opened 2022-05-18
Frame zero in saturation correction - Roman does not have zero frame

nden:
This is a follow-up on #81 - adds a guard for "frame zero" , since Roman does not have one.


ddavis-stsci (approved):
I am able to run ramp fit on Roman data with this change!

PR #90, opened 2022-05-19
Update twopoint_difference.py

mwregan2:
This is a fix for JP-2617 - Incorrect Flagging of saturated groups adjacent to pixels with detected jumps.
The problem was a typo in the section of the code that checks to see if the neighbor pixels already have a DO_NOT_USE or SATURATION flag. This made one of the checks not work correctly, resulting in pixels with both SATURATION and JUMP flags. The ramp_fit step does not behave well when groups have both flags set.

### DIFF
--- a/src/stcal/jump/twopoint_difference.py
+++ b/src/stcal/jump/twopoint_difference.py
@@ -251,8 +251,8 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
                                     np.bitwise_or(gdq[integ, group, row, col - 1], jump_flag)
 
                     if cr_col[j] != ncols - 1:
-                        if (gdq[integ, group, row, col - 1] & sat_flag) == 0:
-                            if (gdq[integ, group, row, col - 1] & dnu_flag) == 0:
+                        if (gdq[integ, group, row, col + 1] & sat_flag) == 0:
+                            if (gdq[integ, group, row, col + 1] & dnu_flag) == 0:
                                 gdq[integ, group, row, col + 1] =\
                                     np.bitwise_or(gdq[integ, group, row, col + 1], jump_flag)
     return gdq, row_below_gdq, row_above_gdq


src/stcal/jump/twopoint_difference.py
```diff
@@ -251,8 +251,8 @@ def find_crs(dataa, group_dq, read_noise, normal_rej_thresh,
                                     np.bitwise_or(gdq[integ, group, row, col - 1], jump_flag)
 
                     if cr_col[j] != ncols - 1:
-                        if (gdq[integ, group, row, col - 1] & sat_flag) == 0:
-                            if (gdq[integ, group, row, col - 1] & dnu_flag) == 0:
+                        if (gdq[integ, group, row, col + 1] & sat_flag) == 0:
```
cshanahan1, line 254: I think what actually needs to be done here is this:

```
                    if cr_col[j] != ncols - 1:
                        if (gdq[integ, group, row, col - 1] & sat_flag) == 0:
                            if (gdq[integ, group, row, col - 1] & dnu_flag) == 0:
                                gdq[integ, group, row, col - 1] =\
                                    np.bitwise_or(gdq[integ, group, row, col - 1], jump_flag)
                        if (gdq[integ, group, row, col + 1] & sat_flag) == 0:
                            if (gdq[integ, group, row, col + 1] & dnu_flag) == 0:
                                gdq[integ, group, row, col + 1] =\
                                    np.bitwise_or(gdq[integ, group, row, col + 1], jump_flag)
```

cshanahan1: this block runs if its not a column at the edge, so it should be checking col-1 and col+1 individually.

we also need to apply this logic to the rows above and below.

cshanahan1 (changes requested):
the neighbor flagging in jump looks wrong to me, i left a comment inline but it looks like as it stands only pixels in the column to the left of a 'jump' pixel are checked to see if they're saturated or do not use before the'yre also flagged. this change now just checks if the pixel to the left is saturated/do not use, but it should be checking both. this also applies to rows.

also, i couldn't find this in the description of the step but why is it that we're only flagging 4 adjacent pixels and not every pixel that touches the jump pixel (like in saturation)?

src/stcal/jump/twopoint_difference.py
mwregan2: It looks like to me your proposed version only checks to make sure this pixel is not in the last column. That applies to the first setting of the jump flag. The second line that sets the pixel to the left of the current pixel needs to have a check that this is not the first column. You seem to have removed that check.
All that I changed was correcting the typo in lines 254 and 255 changing the "-" to a "+". I worried that we're not looking at the same version of the code. Is that what you see?

mwregan2:
I missed your question about why do we not flag the diagonal pixels. The flagging for jumps is due to the electronic coupling between pixels. The measured coupling to the diagonal pixels is extremely small. It might be the case that it would affect slopes for very large jump magnitudes, but that is not what we want to do at this time.
The is the first I've heard about flagging diagonal saturated pixels. That doesn't seem correct to me. I'll do some research and see if that is the best thing to do.

cshanahan1 (approved):

src/stcal/jump/twopoint_difference.py
cshanahan1: (summarizing our call)

the block above `if cr_col[j] != 0:` takes care of the left adjacent pixel, so this sign change takes care of the right. everything looks good here. apologies for my misunderstanding

PR #91, opened 2022-05-19
Update to handle single good group

mwregan2:
The baseline code assumes that if there is only one group in an integration it must be the first group. This is not true when MIRI is flagging more than one frame like happens in the RSCD step.
Also, there was the assumption that only saturation could create the single good group but with CR neighbor flagging turned on you can have the first group good and the second group have a jump. So, jumps need to be added to the total number of groups that are bad.
This is a fix for JP-2572.

### DIFF
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -267,20 +267,12 @@ def suppress_one_group_saturated_ramps(ramp_data):
     for integ in range(nints):
         ramp_data.one_groups[integ] = []
         intdq = dq[integ, :, :, :]
-
-        # Find ramps with a good zeroeth group, but saturated in
-        # the remainder of the ramp.
-        wh_one = groups_saturated_in_integration(intdq, sat_flag, ngroups - 1)
-
-        sat_groups = np.zeros(intdq.shape, dtype=int)
-        sat_groups[np.where(np.bitwise_and(intdq, sat_flag))] = 1
-        jump_groups = np.zeros(intdq.shape, dtype=int)
-        jump_groups[np.where(np.bitwise_and(intdq, jump_flag))] = 1
-        sat_groups[np.where(intdq == 1)] = 1
-        jump_groups[np.where(intdq == 4)] = 1
-        nsat_groups = sat_groups.sum(axis=0)
-        njump_groups = jump_groups.sum(axis=0)
-        nbad_groups = nsat_groups + njump_groups
+        # Find ramps with only one group that is not saturated and
+        # not jump (ie only one good group)
+        bad_flags = np.bitwise_or(sat_flag, jump_flag)
+        bad_groups = np.zeros(intdq.shape, dtype=int)
+        bad_groups[np.where(np.bitwise_and(intdq, bad_flags))] = 1
+        nbad_groups = bad_groups.sum(axis=0)
         wh_one = np.where(nbad_groups == (ngroups - 1))
 
         wh1_rows = wh_one[0]
@@ -288,8 +280,7 @@ def suppress_one_group_saturated_ramps(ramp_data):
         for n in range(len(wh1_rows)):
             row = wh1_rows[n]
             col = wh1_cols[n]
-            # For ramps that have good 0th group, but the rest of the
-            # ramp saturated, mark the 0th groups as saturated, too.
+            # For ramps that have only one good group, mark that group as saturated
             good_index = np.where(ramp_data.groupdq[integ, :, row, col] == 0)
             if ramp_data.groupdq[integ, good_index, row, col] == 0:
                 ramp_data.groupdq[integ, good_index, row, col] = sat_flag
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -236,6 +236,7 @@ def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
 
         # Suppress one group ramps, if desired.
         if ramp_data.suppress_one_group_ramps:
+            log.info("Suppressing the first group when saturated in second group")
             suppress_one_group_saturated_ramps(ramp_data)
 
         # Compute ramp fitting using ordinary least squares.
@@ -259,6 +260,7 @@ def suppress_one_group_saturated_ramps(ramp_data):
     dq = ramp_data.groupdq
     nints, ngroups, nrows, ncols = dq.shape
     sat_flag = ramp_data.flags_saturated
+    jump_flag = ramp_data.flags_jump_det
 
     ramp_data.one_groups = [None] * nints
 
@@ -270,6 +272,17 @@ def suppress_one_group_saturated_ramps(ramp_data):
         # the remainder of the ramp.
         wh_one = groups_saturated_in_integration(intdq, sat_flag, ngroups - 1)
 
+        sat_groups = np.zeros(intdq.shape, dtype=int)
+        sat_groups[np.where(np.bitwise_and(intdq, sat_flag))] = 1
+        jump_groups = np.zeros(intdq.shape, dtype=int)
+        jump_groups[np.where(np.bitwise_and(intdq, jump_flag))] = 1
+        sat_groups[np.where(intdq == 1)] = 1
+        jump_groups[np.where(intdq == 4)] = 1
+        nsat_groups = sat_groups.sum(axis=0)
+        njump_groups = jump_groups.sum(axis=0)
+        nbad_groups = nsat_groups + njump_groups
+        wh_one = np.where(nbad_groups == (ngroups - 1))
+
         wh1_rows = wh_one[0]
         wh1_cols = wh_one[1]
         for n in range(len(wh1_rows)):
@@ -277,9 +290,9 @@ def suppress_one_group_saturated_ramps(ramp_data):
             col = wh1_cols[n]
             # For ramps that have good 0th group, but the rest of the
             # ramp saturated, mark the 0th groups as saturated, too.
-
-            if ramp_data.groupdq[integ, 0, row, col] == 0:
-                ramp_data.groupdq[integ, 0, row, col] = sat_flag
+            good_index = np.where(ramp_data.groupdq[integ, :, row, col] == 0)
+            if ramp_data.groupdq[integ, good_index, row, col] == 0:
+                ramp_data.groupdq[integ, good_index, row, col] = sat_flag
                 sat_pix = (row, col)
                 ramp_data.one_groups[integ].append(sat_pix)
 
--- a/tests/test_ramp_fitting.py
+++ b/tests/test_ramp_fitting.py
@@ -261,12 +261,12 @@ def test_utils_dq_compress_final():
     var = (rnoise_val, gain_val)
     tm = (nframes, gtime, dtime)
     ramp_data, rnoise, gain = setup_inputs(dims, var, tm)
+
     ramp_data.groupdq[0, :, 0, 0] = np.array([dqflags["SATURATED"]] * ngroups)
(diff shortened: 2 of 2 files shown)


nden:
@mwregan2 There are conflicts in one file. Can you rebase?

kmacdonald-stsci (commented):
This is branched from an old version missing JP-374 updates that use the ZEROFRAME data.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -263,8 +265,15 @@ def suppress_one_group_saturated_ramps(ramp_data):
...
+        jump_groups = np.zeros(intdq.shape, dtype=int)
+        jump_groups[np.where(np.bitwise_and(intdq, jump_flag))] = 1
+        sat_groups[np.where(intdq == 1)] = 1
+        jump_groups[np.where(intdq == 4)] = 1
         nsat_groups = sat_groups.sum(axis=0)
-        wh_one = np.where(nsat_groups == (ngroups - 1))
+        njump_groups = jump_groups.sum(axis=0)
+        nbad_groups = nsat_groups + njump_groups
+        wh_one = np.where(nbad_groups == (ngroups - 1))
+
```
dmggh, line 276: Does line 269 :
         jump_groups[np.where( np.bitwise_and( intdq, jump_flag ) )] = 1  

    . . .  flag any pixels that won't be flagged anyway by line 271 :

    jump_groups[np.where(intdq == 4)] = 1  ?

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -263,8 +265,15 @@ def suppress_one_group_saturated_ramps(ramp_data):
...
         sat_groups = np.zeros(intdq.shape, dtype=int)
         sat_groups[np.where(np.bitwise_and(intdq, sat_flag))] = 1
+        jump_groups = np.zeros(intdq.shape, dtype=int)
+        jump_groups[np.where(np.bitwise_and(intdq, jump_flag))] = 1
+        sat_groups[np.where(intdq == 1)] = 1
+        jump_groups[np.where(intdq == 4)] = 1
         nsat_groups = sat_groups.sum(axis=0)
-        wh_one = np.where(nsat_groups == (ngroups - 1))
+        njump_groups = jump_groups.sum(axis=0)
+        nbad_groups = nsat_groups + njump_groups
```
dmggh, line 274: It would be a little easier to read if the jump_groups statements are grouped together, and similarly for the sat_groups statements.

dmggh (approved):
Looks good to me.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -263,8 +265,15 @@ def suppress_one_group_saturated_ramps(ramp_data):
         # the remainder of the ramp.
         sat_groups = np.zeros(intdq.shape, dtype=int)
         sat_groups[np.where(np.bitwise_and(intdq, sat_flag))] = 1
+        jump_groups = np.zeros(intdq.shape, dtype=int)
+        jump_groups[np.where(np.bitwise_and(intdq, jump_flag))] = 1
+        sat_groups[np.where(intdq == 1)] = 1
```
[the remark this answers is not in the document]
nden: Does `1` indicate `dqflags['GOOD']` and `4` - `dqflags['JUMP_DET]`?

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -259,27 +260,30 @@ def suppress_one_group_saturated_ramps(ramp_data):
...
         wh1_cols = wh_one[1]
         for n in range(len(wh1_rows)):
             row = wh1_rows[n]
             col = wh1_cols[n]
-            # For ramps that have good 0th group, but the rest of the
-            # ramp saturated, mark the 0th groups as saturated, too.
-
-            if ramp_data.groupdq[integ, 0, row, col] == 0:
-                ramp_data.groupdq[integ, 0, row, col] = sat_flag
+            # For ramps that have only one good group, mark that group as saturated
```
cshanahan1, line 283: i am not super familiar with ramp fitting and am trying to understand this. what is happening here - if there is one good group of data why is the 'saturation' flag being applied? saturation runs before this and if this group wasn't flagged then why is it being called saturated here?

i assume theyre being flagged so a ramp wont be fit? if thats the case why not the donotuse flag or something?

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -259,27 +260,30 @@ def suppress_one_group_saturated_ramps(ramp_data):
...
 
     for integ in range(nints):
         ramp_data.one_groups[integ] = []
         intdq = dq[integ, :, :, :]
-
-        # Find ramps with a good zeroeth group, but saturated in
-        # the remainder of the ramp.
-        wh_one = groups_saturated_in_integration(intdq, sat_flag, ngroups - 1)
+        # Find ramps with only one group that is not saturated and
+        # not jump (ie only one good group)
```
cshanahan1, line 271: along with saturation and jump, could 'do not use' or 'ad floor' flags also cause this scenario?

mwregan2:
The quick answer is that ramp fitting was originally only looking at Saturation and Jump. When we added the DO_NOT_USE flag to groups some of the code did not work  correctly. I thought it was safest to just flag the group as saturated. The key is to not fit a slope for this group.


From: Clare Shanahan ***@***.***>
Date: Friday, May 20, 2022 at 11:41 AM
Subject: Re: [spacetelescope/stcal] Update to handle single good group (PR #91)


@cshanahan1 commented on this pull request.

________________________________

In src/stcal/ramp_fitting/ramp_fit.py<https://urldefense.com/v3/__https:/github.com/spacetelescope/stcal/pull/91*discussion_r878280653__;Iw!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!3oT2OZbOws097wghnmKH4s9J6932m1FEsQ_jLm8-LdqDRUClBNpxmPJmRVDk2YeU7Q5hHL8h07FIeIndAdKEvYk$>:

>

         wh1_rows = wh_one[0]

         wh1_cols = wh_one[1]

         for n in range(len(wh1_rows)):

             row = wh1_rows[n]

             col = wh1_cols[n]

-            # For ramps that have good 0th group, but the rest of the

-            # ramp saturated, mark the 0th groups as saturated, too.

-

-            if ramp_data.groupdq[integ, 0, row, col] == 0:

-                ramp_data.groupdq[integ, 0, row, col] = sat_flag

+            # For ramps that have only one good group, mark that group as saturated

i am not super familiar with ramp fitting and am trying to understand this. what is happening here - if there is one good group of data why is the 'saturation' flag being applied? saturation runs before this and if this group wasn't flagged then why is it being called saturated here?

i assume theyre being flagged so a ramp wont be fit? if thats the case why not the donotuse flag or something?

________________________________

In src/stcal/ramp_fitting/ramp_fit.py<https://urldefense.com/v3/__https:/github.com/spacetelescope/stcal/pull/91*discussion_r878286988__;Iw!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!3oT2OZbOws097wghnmKH4s9J6932m1FEsQ_jLm8-LdqDRUClBNpxmPJmRVDk2YeU7Q5hHL8h07FIeIndAy-dcfg$>:

>

     ramp_data.one_groups = [None] * nints



     for integ in range(nints):

         ramp_data.one_groups[integ] = []

         intdq = dq[integ, :, :, :]

-

-        # Find ramps with a good zeroeth group, but saturated in

-        # the remainder of the ramp.

-        wh_one = groups_saturated_in_integration(intdq, sat_flag, ngroups - 1)

+        # Find ramps with only one group that is not saturated and

+        # not jump (ie only one good group)

along with saturation and jump, could 'do not use' or 'ad floor' flags also cause this scenario?

cshanahan1:
@mwregan2 ok so the 'saturation' flag is not really reflective of it being saturated, but rather just used to control ramp fitting. thats a little confusing but makes sense if that's what is being done elsewhere, maybe it should be changed one day because i can imagine looking at the output and seeing saturated flags (unless theyre reset later?) would be confusing.

mwregan2:
But, when there is no slope calculated the pixel will be flagged as DO_NOT_USE. It will already have the saturated bit set because that’s why there is only one good group.

From: Clare Shanahan ***@***.***>
Date: Friday, May 20, 2022 at 11:47 AM
Subject: Re: [spacetelescope/stcal] Update to handle single good group (PR #91)


@mwregan2<https://urldefense.com/v3/__https:/github.com/mwregan2__;!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!3m-wQSLh1SKPX15S8-e77hKlM1q73l_hWKtfBIGa_HMjPNbSq_QDq1jeyl7x2Jfsn8yAJg28y7jngK6Rec9vZL8$> ok so the 'saturation' flag is not really reflective of it being saturated, but rather just used to control ramp fitting. thats a little confusing but makes sense if that's what is being done elsewhere, maybe it should be changed one day because i can imagine looking at the output and seeing saturated flags (unless theyre reset later?) would be confusing.

cshanahan1:
> But, when there is no slope calculated the pixel will be flagged as DO_NOT_USE. It will already have the saturated bit set because that’s why there is only one good group.>

What if the reason there was only one good group was because the others were flagged as jump and not saturation? Then wont you have the unwanted case of jump + saturation set?

cshanahan1:
one of the ramp fitting tests is failing, does it need to be modified?

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: This is happening during suppression of one group ramps.  There is an option to suppress ramps that saturate in group 1, but still has a good group 0.  To suppress this group 0, its DQ flag is set to saturation.

dmggh:
I agree that labelling an unsaturated group as SATURATED merely to prevent it being used in a fit is misleading. Using DONOTUSE instead is preferable. The code may need to be modified to further differentiate between non-zero DQ values.

mwregan2:
Ken,
  Are you saying that we should set the suppressed group to Saturated? David and Clare are saying it should be set to DO_NOT_USE.

  The current code sets the suppressed group to SATURATED. My changes are just to make this work when the single group is not the first group.
Mike

From: Ken MacDonald ***@***.***>
Date: Friday, May 20, 2022 at 12:17 PM
Subject: Re: [spacetelescope/stcal] Update to handle single good group (PR #91)


@kmacdonald-stsci commented on this pull request.

________________________________

In src/stcal/ramp_fitting/ramp_fit.py<https://urldefense.com/v3/__https:/github.com/spacetelescope/stcal/pull/91*discussion_r878323185__;Iw!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!wv7672g2J2D8TI9nRmGyZ1-NJffCBk2kBghM2Lyz0t_sMaN_FOnybO1mIkHrzcKAfvqZkPT7kfRMfNSZHMcmvxo$>:

>

         wh1_rows = wh_one[0]

         wh1_cols = wh_one[1]

         for n in range(len(wh1_rows)):

             row = wh1_rows[n]

             col = wh1_cols[n]

-            # For ramps that have good 0th group, but the rest of the

-            # ramp saturated, mark the 0th groups as saturated, too.

-

-            if ramp_data.groupdq[integ, 0, row, col] == 0:

-                ramp_data.groupdq[integ, 0, row, col] = sat_flag

+            # For ramps that have only one good group, mark that group as saturated

This is happening during suppression of one group ramps. There is an option to suppress ramps that saturate in group 1, but still has a good group 0. To suppress this group 0, its DQ flag is set to saturation.

kmacdonald-stsci:
> Ken, Are you saying that we should set the suppressed group to Saturated? David and Clare are saying it should be set to DO_NOT_USE. 

I am not saying that is "should" be set to SATURATED.  I am simply saying that is what I did to suppress that group.  I agree it's probably better to flag it as DO_NOT_USE, rather than as SATURATED.  I was just answering Clare's question as to what the code was doing.

nden:
Superseded by #92

PR #92, opened 2022-05-20
JP-2572: Updating the handler for the suppression of one good group ramps.

kmacdonald-stsci:
Due to the possibility the first groups of a ramp could be flagged as DO_NOT_USE, as well as the way the jump detector works, the one good group suppression handler needed to be modified to handle these cases as well.

### DIFF
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -236,7 +236,7 @@ def ramp_fit_data(ramp_data, buffsize, save_opt, readnoise_2d, gain_2d,
 
         # Suppress one group ramps, if desired.
         if ramp_data.suppress_one_group_ramps:
-            suppress_one_group_saturated_ramps(ramp_data)
+            suppress_one_group_saturated_or_jump_ramps(ramp_data)
 
         # Compute ramp fitting using ordinary least squares.
         image_info, integ_info, opt_info = ols_fit.ols_ramp_fit_multi(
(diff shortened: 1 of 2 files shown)


src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -259,16 +259,21 @@ def suppress_one_group_saturated_ramps(ramp_data):
     dq = ramp_data.groupdq
     nints, ngroups, nrows, ncols = dq.shape
     sat_flag = ramp_data.flags_saturated
+    jump_flag = ramp_data.flags_jump_det
```
dmggh, line 262: May want to change the name (and perhaps the description) of this function since this is not strictly for saturation.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -259,16 +259,21 @@ def suppress_one_group_saturated_ramps(ramp_data):
...
-        # Find ramps with a good zeroeth group, but saturated in
-        # the remainder of the ramp.
-        wh_one = groups_saturated_in_integration(intdq, sat_flag, ngroups - 1)
+        # Find ramps with only one group that is not saturated and
+        # not jump (i.e., only one good group).
+        bad_flags = np.bitwise_or(sat_flag, jump_flag)
+        bad_groups = np.zeros(intdq.shape, dtype=int)
+        bad_groups[np.where(np.bitwise_and(intdq, bad_flags))] = 1
+        nbad_groups = bad_groups.sum(axis=0)
+        wh_one = np.where(nbad_groups == (ngroups - 1))
```
dmggh, line 277: Pending these updates, will the groups_saturated_in_integration() function still be needed ?

mwregan2 (approved):
Looks good to me.

dmggh (approved):
Aside from 2 minor comments, looks good.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Yes.  It's used in `use_zeroframe_for_saturated_ramps`.

cshanahan1 (approved):

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I changed the function name.

PR #95, opened 2022-05-23
move build configuration into `pyproject.toml`

zacharyburnett:
`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.


zacharyburnett:
blocked by #98

zacharyburnett:
Now that Python 3.11 includes `tomli` in built-ins, and given that `stcal` doesn't build C extensions, I am comfortable with readying this for review. `flake8` requires its own file, but if we ever move to a different linter in the future we will be able to use `pyproject.toml`

zacharyburnett:
If `setuptools` changes their interface with `pyproject.toml` it would involve updating the `[tool.setuptools]` table, i.e. 
```yaml
 [tool.setuptools]
 zip-safe = true

 [tool.setuptools.packages.find]
 where = ['src']
 ```
 
 As the `[project]` table is defined by PEP621 it would remain unchanged.

nden (approved):
LGTM

PR #96, opened 2022-05-24
make CI dependent on linting

zacharyburnett:


zacharyburnett:
The following status check requirements should also probably be updated to reflect the new names:
image

zacharyburnett:
additionally, I observed that any Python 3.7 tests will fail due to `astropy` no longer supporting that version; should `stcal` still support `>=3.7` as it says in `setup.cfg`?

stscieisenhamer:
You are correct: `stcal` should be only be good for python>=3.8

hbushouse:
Is there a way to retrigger the stalled and/or failing CI tests?

zacharyburnett:
> Is there a way to retrigger the stalled and/or failing CI tests?

yes indeed, I just reran the failed JWST section; is the issue with the CRDS ref files corrected now?

hbushouse:
The issue with the ref files is not yet corrected, but that only affects 2 unit tests. The jwst job log shows hundreds of errors, apparently related to getting crds ref files for many tests?

And why are several of the CI tests just hung (like security audit, macOS, romancal tests with coverage, etc.)? Are they waiting for something?

zacharyburnett:
> The issue with the ref files is not yet corrected, but that only affects 2 unit tests. The jwst job log shows hundreds of errors, apparently related to getting crds ref files for many tests?

I'm unsure why it is so incredibly long; I'm working on a commit to run the tests without `tox` to see if it might be an issue with the installed environment

> And why are several of the CI tests just hung (like security audit, macOS, romancal tests with coverage, etc.)? Are they waiting for something?

those are the "required status checks"; the repo settings are looking for those specific names to pass before approving the merge, but those jobs have new names in the matrix now so those setting would have to be updated

zacharyburnett:
I changed the environment construction method from `tox` to GitHub Actions and it seems to get rid of all of the extraneous errors that were going on in the JWST pipeline tests (except for the currently failing tests)

nden:
It looks good! Could you clean up the `required` checks?
Also the jwst tests should be fixed now, needs a rerun.

zacharyburnett:
> It looks good! Could you clean up the `required` checks?
> Also the jwst tests should be fixed now, needs a rerun.

I reran the JWST tests, it looks like everything passes now. I'll take care of the required checks

nden (approved):

PR #98, opened 2022-05-31
update required Python version to `3.8` to align with `astropy`

zacharyburnett:
> additionally, I observed that any Python 3.7 tests will fail due to `astropy` no longer supporting that version; should `stcal` still support `>=3.7` as it says in `setup.cfg`?
https://github.com/spacetelescope/stcal/pull/96#issuecomment-1138866222

> You are correct: `stcal` should be only be good for python>=3.8
https://github.com/spacetelescope/stcal/pull/96#issuecomment-1138896163


hbushouse (approved):

zacharyburnett:
thank you!

nden (approved):

PR #102, opened 2022-06-21
JP-2651: Fully Saturated Exposures Should Return NoneType

kmacdonald-stsci:
Fully saturated exposures have no good data to process, so should return `NoneType` for all return variables.  The location for the full saturation check moved to before multiprocessing could take place.  This prevents the possibility of multiprocessing from computing ramp fitting for some data, but returning `NoneType` for data that just happens to be saturated after slicing up the data for multiprocessing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ ramp_fitting
 - Updating how one group suppression and ZEROFRAME processing works with
   multiprocessing, as well as fixing the multiprocessing failure. [#99]
 
+- Changing how ramp fitting handles fully saturated ramps. [#102]
+
 Changes to API
 --------------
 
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
Looks good to me. Just needs a change log entry.

hbushouse:
@ddavis-stsci @PaulHuwe should we get a Romancal regtest run against this PR?

ddavis-stsci (approved):
The current roman regression tests pass, https://plwishmaster.stsci.edu:8081/me/my-views/view/all/job/RT/job/Roman-Developers-Pull-Requests/49/
We will need to add a test for fully saturated data.

PaulHuwe (approved):
RomanCAL Unit & regression tests pass, though we have no tests for fully saturated data - this is being worked on at present, and does not hinder this PR.

PR #105, opened 2022-06-23
changing param name in jump twopoint_difference function

cshanahan1:
Changing the normal rejection threshold parameter name in twopoint_difference to match what it is called elsewhere in jump step in jwst/roman


hbushouse (approved):
LGTM. Needs a change log entry.

kmacdonald-stsci (approved):

PR #106, opened 2022-06-23
JP-2611: Update handling of NO_SAT_CHECK in saturation step

hbushouse:
Resolves [JP-2611](https://jira.stsci.edu/browse/JP-2611)

This PR makes a minor modification to the way pixels flagged with NO_SAT_CHECK or having thresholds of NaN are treated in terms of flagging as saturated. Previously, these pixels had their internal thresholds set to 65535 (the 16-bit A-to-D converter limit), but because the comparison is done as >= threshold, they would still get flagged as saturated in their raw values were 65535. JP-2611 requested that pixels flagged with NO_SAT_CHECK should never get flagged, regardless of their values. So this change simply sets the internal threshold for NO_SAT_CHECK and NaN pixels to 65535+1, which means they will never have values >= threshold and hence will never get flagged as saturated.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stcal/saturation/saturation.py
+++ b/src/stcal/saturation/saturation.py
@@ -67,13 +67,14 @@ def flag_saturated_pixels(
     no_sat_check = dqflags['NO_SAT_CHECK']
 
     # For pixels flagged in reference file as NO_SAT_CHECK,
-    # set the saturation check threshold to the A-to-D converter limit.
-    sat_thresh[np.bitwise_and(sat_dq, no_sat_check) == no_sat_check] = atod_limit
+    # set the saturation check threshold above the A-to-D converter limit,
+    # so that they don't get flagged as saturated.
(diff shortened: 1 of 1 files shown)


cshanahan1:
looks good to me. these changes need to be made in jwst for the irs2 mode as well

kmacdonald-stsci (approved):

nden:
@cshanahan1 Please run a romancal regression test with this PR and merge if OK.

hbushouse:
@cshanahan1 are you going to run a romancal regtest against this PR branch?

cshanahan1:
@hbushouse sure i can do that

cshanahan1:
roman regression tests are passing

cshanahan1 (approved):

PR #110, opened 2022-07-26
JP-2693: Flag groups do_not_use after detected ramp jumps

karllark:
Resolves [JP-2693](https://jira.stsci.edu/browse/JP-2693)

This PR implements the enhancement to the jump step where groups after detected ramps are flagged as jumps.  This "corrects" transients seen that make the slope after a jump different than before.  This PR allows for flagging to be different for two different jump thresholds.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -144,12 +144,13 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
     elif max_cores == 'all':
         n_slices = max_available
 
+    flag_n_after_jump = 5
     if n_slices == 1:
         gdq, row_below_dq, row_above_dq = \
             twopt.find_crs(data, gdq, readnoise_2d, rejection_thresh,
                            three_grp_thresh, four_grp_thresh, frames_per_group,
                            flag_4_neighbors, max_jump_to_flag_neighbors,
-                           min_jump_to_flag_neighbors, dqflags)
+                           min_jump_to_flag_neighbors, flag_n_after_jump, dqflags)
 
         elapsed = time.time() - start
     else:
@@ -174,7 +175,8 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
                               rejection_thresh, three_grp_thresh, four_grp_thresh,
                               frames_per_group, flag_4_neighbors,
                               max_jump_to_flag_neighbors,
-                              min_jump_to_flag_neighbors, dqflags, copy_arrs))
+                              min_jump_to_flag_neighbors, flag_n_after_jump, dqflags,
+                              copy_arrs))
 
         # last slice get the rest
         slices.insert(n_slices - 1, (data[:, :, (n_slices - 1) * yinc:n_rows, :],
@@ -183,7 +185,8 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
                                      rejection_thresh, three_grp_thresh,
                                      four_grp_thresh, frames_per_group,
                                      flag_4_neighbors, max_jump_to_flag_neighbors,
-                                     min_jump_to_flag_neighbors, dqflags, copy_arrs))
+                                     min_jump_to_flag_neighbors, flag_n_after_jump,
+                                     dqflags, copy_arrs))
         log.info("Creating %d processes for jump detection " % n_slices)
         pool = multiprocessing.Pool(processes=n_slices)
         # Starts each slice in its own process. Starmap allows more than one
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -134,6 +134,9 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
     data *= gain_2d
     err *= gain_2d
     readnoise_2d *= gain_2d
+    # also apply to the after_jump thresholds
+    after_jump_flag_e1 = after_jump_flag_dn1 * gain_2d
+    after_jump_flag_e2 = after_jump_flag_dn2 * gain_2d
 
     # Apply the 2-point difference method as a first pass
     log.info('Executing two-point difference method')
@@ -168,9 +171,9 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
                            three_grp_thresh, four_grp_thresh, frames_per_group,
                            flag_4_neighbors, max_jump_to_flag_neighbors,
                            min_jump_to_flag_neighbors, dqflags,
-                           after_jump_flag_dn1=after_jump_flag_dn1,
+                           after_jump_flag_e1=after_jump_flag_e1,
                            after_jump_flag_n1=after_jump_flag_n1,
-                           after_jump_flag_dn2=after_jump_flag_dn2,
+                           after_jump_flag_e2=after_jump_flag_e2,
                            after_jump_flag_n2=after_jump_flag_n2)
 
         elapsed = time.time() - start
@@ -196,10 +199,10 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
                               rejection_thresh, three_grp_thresh, four_grp_thresh,
                               frames_per_group, flag_4_neighbors,
                               max_jump_to_flag_neighbors,
-                              min_jump_to_flag_neighbors,
-                              after_jump_flag_dn1, after_jump_flag_n1,
-                              after_jump_flag_dn2, after_jump_flag_n2,
-                              dqflags, copy_arrs))
+                              min_jump_to_flag_neighbors, dqflags,
+                              after_jump_flag_e1, after_jump_flag_n1,
+                              after_jump_flag_e2, after_jump_flag_n2,
+                              copy_arrs))
 
         # last slice get the rest
         slices.insert(n_slices - 1, (data[:, :, (n_slices - 1) * yinc:n_rows, :],
@@ -208,10 +211,10 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
                                      rejection_thresh, three_grp_thresh,
                                      four_grp_thresh, frames_per_group,
                                      flag_4_neighbors, max_jump_to_flag_neighbors,
-                                     min_jump_to_flag_neighbors,
-                                     after_jump_flag_dn1, after_jump_flag_n1,
-                                     after_jump_flag_dn2, after_jump_flag_n2,
-                                     dqflags, copy_arrs))
+                                     min_jump_to_flag_neighbors, dqflags,
+                                     after_jump_flag_e1, after_jump_flag_n1,
+                                     after_jump_flag_e2, after_jump_flag_n2,
+                                     copy_arrs))
         log.info("Creating %d processes for jump detection " % n_slices)
         pool = multiprocessing.Pool(processes=n_slices)
         # Starts each slice in its own process. Starmap allows more than one
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -14,9 +14,10 @@ log.setLevel(logging.DEBUG)
 def detect_jumps(frames_per_group, data, gdq, pdq, err,
                  gain_2d, readnoise_2d, rejection_thresh,
                  three_grp_thresh, four_grp_thresh, max_cores, max_jump_to_flag_neighbors,
+                 min_jump_to_flag_neighbors, flag_4_neighbors,
                  after_jump_flag_dn1, after_jump_flag_n1,
                  after_jump_flag_dn2, after_jump_flag_n2,
-                 min_jump_to_flag_neighbors, flag_4_neighbors, dqflags):
+                 dqflags):
     """
     This is the high-level controlling routine for the jump detection process.
     It loads and sets the various input data and parameters needed by each of
@@ -165,6 +166,7 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
             twopt.find_crs(data, gdq, readnoise_2d, rejection_thresh,
                            three_grp_thresh, four_grp_thresh, frames_per_group,
                            flag_4_neighbors, max_jump_to_flag_neighbors,
+                           min_jump_to_flag_neighbors,
                            after_jump_flag_dn1, after_jump_flag_n1,
                            after_jump_flag_dn2, after_jump_flag_n2,
                            dqflags)
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -93,20 +93,20 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
         DO_NOT_USE, SATURATED, JUMP_DET, NO_GAIN_VALUE, GOOD
 
     after_jump_flag_dn1 : float
-        1st flag after jumps with the specified DN jump
-        to remove the transient seen from the slope calculation
+        Jumps with amplitudes above the specified DN value will have subsequent
+        groups flagged with the number determined by the after_jump_flag_n1
 
     after_jump_flag_n1 : int
-        1st flag n groups after companion threshold
-        to remove the transient seen from the slope calculation
+        Gives the number of groups to flag after jumps with DN values above that
+        given by after_jump_flag_dn1
 
     after_jump_flag_dn2 : float
-        2nd flag after jumps with the specified DN jump
-        to remove the transient seen from the slope calculation
+        Jumps with amplitudes above the specified DN value will have subsequent
+        groups flagged with the number determined by the after_jump_flag_n2
 
     after_jump_flag_n2 : int
-        2nd flag n groups after companion threshold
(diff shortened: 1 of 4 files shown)


karllark:
Not sure how to update the docs for this PR.   There doesn't seem to much in the way of docs at all.  Please advise.

hbushouse:
> Not sure how to update the docs for this PR. There doesn't seem to much in the way of docs at all. Please advise.

If you've at least checked to make sure that no updates are needed, that's good enough. Just leave the box unchecked.

karllark:
Ready for review. Likely, the corresponding PR in jwst should be reviewed at the same time.

src/stcal/jump/jump.py
```diff
@@ -116,6 +134,9 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
     data *= gain_2d
     err *= gain_2d
     readnoise_2d *= gain_2d
+    # also apply to the after_jump thresholds
+    after_jump_flag_e1 = after_jump_flag_dn1 * gain_2d
```
hbushouse, line 138: Given that `gain_2d` is a 2D array of gain values, does that mean `after_jump_flag_e[12]` are also now full 2D arrays?

src/stcal/jump/twopoint_difference.py
```diff
@@ -122,7 +147,8 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
         # compute 'ratio' for each group. this is the value that will be
         # compared to 'threshold' to classify jumps. subtract the median of
         # first_diffs from first_diffs, take the abs. value and divide by sigma.
-        ratio = np.abs(first_diffs - median_diffs[np.newaxis, :, :]) / sigma[np.newaxis, :, :]
+        dn_jump = first_diffs - median_diffs[np.newaxis, :, :]
```
hbushouse, line 150: At this point in the flow aren't the pixel values in units of electrons, hence the name `dn_jump` may be misleading?

src/stcal/jump/twopoint_difference.py
```diff
@@ -138,7 +164,7 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
                                   max_ratio > two_diff_rej_thresh))
 
         log_str = 'From highest outlier, two-point found {} pixels with at least one CR from {} groups.'
-        log.info(log_str.format(len(row4cr), 'five'))
+        log.info(log_str.format(len(row4cr), 'five or more'))
```
hbushouse, line 167: Thanks for adding that. That's been bugging me for a while now!

src/stcal/jump/twopoint_difference.py
```diff
@@ -255,6 +281,52 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
                             if (gdq[integ, group, row, col + 1] & dnu_flag) == 0:
                                 gdq[integ, group, row, col + 1] =\
                                     np.bitwise_or(gdq[integ, group, row, col + 1], jump_flag)
+
+        # flag n groups after all jumps to account for the transient seen
+        flag_dn_threshold = [after_jump_flag_e1, after_jump_flag_e2]
```
hbushouse, line 286: Another potentially misleading variable name, if the units are actually electrons (but not a big deal)

src/stcal/jump/twopoint_difference.py
```diff
@@ -255,6 +281,52 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
...
+        flag_groups = [after_jump_flag_n1, after_jump_flag_n2]
+        # ensure the smallest threshold is 1st
```
hbushouse, line 288: Do we really want to leave all of this unused code in here?

src/stcal/jump/jump.py
karllark: Yes.

karllark: Update the docstring for find_crs.

src/stcal/jump/twopoint_difference.py
karllark: renamed to `e_jump`.

src/stcal/jump/twopoint_difference.py
karllark: Same!

src/stcal/jump/twopoint_difference.py
karllark: Changed to `flag_e_theshold`.

src/stcal/jump/twopoint_difference.py
karllark: No.  Deleted.

src/stcal/jump/jump.py
```diff
@@ -88,7 +92,21 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
         A dictionary with at least the following keywords:
         DO_NOT_USE, SATURATED, JUMP_DET, NO_GAIN_VALUE, GOOD
 
+    after_jump_flag_dn1 : float
```
cshanahan1, line 95: could you add a slightly more descriptive docstring for these, it's not clear exactly what they control just based on this description. also, what do you think about grouping them into 2 tuples so there are not 4 new parameters (each n grouped with the corresponding dn?)

src/stcal/jump/twopoint_difference.py
```diff
@@ -255,6 +281,52 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
...
+                    row = cr_row[j]
+                    col = cr_col[j]
+                    if dn_jump[group - 1, row, col] >= cthres[row, col]:
+                        # print("stcal ", group, group + cgroup + 1,
+                        #       dn_jump[group - 1, row, col], cthres[row, col])
+                        for kk in range(group, min(group + cgroup + 1, ngroups)):
+                            if (gdq[integ, kk, row, col] & sat_flag) == 0:
+                                if (gdq[integ, kk, row, col] & dnu_flag) == 0:
+                                    gdq[integ, kk, row, col] =\
+                                        np.bitwise_or(gdq[integ, kk, row, col], jump_flag)
```
cshanahan1, line 304: the pr description says that pixels after a jump are flagged with 'do not use', but here they are being flagged with jump. 

one thought i have about this is that even though subsequent groups arent really 'jumps', neither are adjacent pixels but those obtain a jump flag too, so maybe it does make sense to use the jump flag here?

src/stcal/jump/twopoint_difference.py
cshanahan1: why is this commented out? maybe we should give a warning if someone tries to pass in thresholds that are out of order?

src/stcal/jump/twopoint_difference.py
cshanahan1: this is in electrons right, if so rename var accordingly to avoid confusion with units

src/stcal/jump/twopoint_difference.py
```diff
@@ -255,6 +281,52 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
...
+        flag_dn_threshold = [after_jump_flag_e1, after_jump_flag_e2]
+        flag_groups = [after_jump_flag_n1, after_jump_flag_n2]
+        # ensure the smallest threshold is 1st
+        # if flag_dn_threshold[0] > flag_dn_threshold[1]:
+        #     flag_dn_threshold = np.flip(flag_dn_threshold)
+        #     flag_groups = np.flip(flag_groups)
+
+        cr_group, cr_row, cr_col = np.where(np.bitwise_and(gdq[integ], jump_flag))
+
+        # **should** work and be faster.  But does not.  Sigh.
```
cshanahan1: remove for now if not working

src/stcal/jump/twopoint_difference.py
```diff
@@ -255,6 +281,52 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
...
+        for cthres, cgroup in zip(flag_dn_threshold, flag_groups):
+            if cgroup > 0:
+                log.info(f"Flagging {cgroup} groups after detected jumps with e >= {np.mean(cthres)}.")
+
+                for j in range(len(cr_group)):
+                    group = cr_group[j]
+                    row = cr_row[j]
+                    col = cr_col[j]
+                    if dn_jump[group - 1, row, col] >= cthres[row, col]:
+                        # print("stcal ", group, group + cgroup + 1,
```
cshanahan1: remove

cshanahan1 (commented):
still looking at this and testing it out but i left a few comments

src/stcal/jump/jump.py
```diff
@@ -14,7 +14,11 @@
 def detect_jumps(frames_per_group, data, gdq, pdq, err,
                  gain_2d, readnoise_2d, rejection_thresh,
                  three_grp_thresh, four_grp_thresh, max_cores, max_jump_to_flag_neighbors,
-                 min_jump_to_flag_neighbors, flag_4_neighbors, dqflags):
+                 min_jump_to_flag_neighbors, flag_4_neighbors, dqflags,
+                 after_jump_flag_dn1=0.0,
```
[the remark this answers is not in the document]
karllark: As these are keywords, I thought that defaults might be nice.  Fine with removing them, but just wanted to confirm that this is policy.

src/stcal/jump/jump.py
karllark: I thought about tuples.  But thought that might be more complex to explain than explicit single values like all the other ones.

karllark: Updated the docstring to (hopefully) be more understandable.

src/stcal/jump/twopoint_difference.py
karllark: Fixed the PR description.  Was wrong.

Yes, following what has been done previously, hence flagging as jumps.

src/stcal/jump/twopoint_difference.py
karllark: Was commented out as it was part of an attempt to make more efficient code that failed.  For the current code, it does not matter what order the thresholds are in.

src/stcal/jump/twopoint_difference.py
karllark: Yep. Fixed.

src/stcal/jump/twopoint_difference.py
karllark: Done.

src/stcal/jump/twopoint_difference.py
karllark: Done.

src/stcal/jump/twopoint_difference.py
```diff
@@ -54,6 +59,26 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
         neighbors (marginal detections). Any primary jump below this value will
         not have its neighbors flagged.
 
+    dqflags: dict
+        A dictionary with at least the following keywords:
+        DO_NOT_USE, SATURATED, JUMP_DET, NO_GAIN_VALUE, GOOD
+
+    after_jump_flag_e1 : float
```
cshanahan1, line 66: the 'min/max_jump_to_flag_neighbors' flags are in units of sigma, does it make more sense to express these new parameters in units of electrons as they currently are, or would sigma be suitable? just wanted to check because i could see this getting potentially confusing with different units, but if it makes more sense based on the nature of the issue then thats fine

src/stcal/jump/twopoint_difference.py
```diff
@@ -255,6 +281,29 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
                             if (gdq[integ, group, row, col + 1] & dnu_flag) == 0:
                                 gdq[integ, group, row, col + 1] =\
                                     np.bitwise_or(gdq[integ, group, row, col + 1], jump_flag)
+
+        # flag n groups after jumps above the specified thresholds to account for
+        # the transient seen after ramp jumps
+        flag_e_threshold = [after_jump_flag_e1, after_jump_flag_e2]
+        flag_groups = [after_jump_flag_n1, after_jump_flag_n2]
+
+        cr_group, cr_row, cr_col = np.where(np.bitwise_and(gdq[integ], jump_flag))
```
cshanahan1, line 290: by recalculating cr_row/cr_col here, you are picking up jumps that were flagged only because they were adjacent to a true jump. is this what you want to do? or should only groups after a 'true' jump be flagged?

karllark: Yes.  All ramp jumps regardless of how they were identified need after jump flagging from the analysis I have done.  For MIRI.

src/stcal/jump/twopoint_difference.py
karllark: The after jump transient effect can change based on the amplitude of the ramp jump, not the ratio to the noise.  Hence we need the actual ramp jump amplitude.

src/stcal/jump/twopoint_difference.py
cshanahan1: ok great, we can add another parameter later if needed to control this then

src/stcal/jump/jump.py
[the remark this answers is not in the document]
karllark: I now realize that if we do not have defaults here, then all the tests need updating to explicitly have all 4 keywords with these defaults.  Having keywords with defaults means that existing tests did not need to be modified.  This was the motivation for using keywords and not regular parameters. 

So, removing defaults means adding more to every test.  If removing the defaults is still desired, say so and I'll start cutting and pasting.  But I would leave the defaults here if possible to avoid this.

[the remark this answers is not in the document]
hbushouse: In that case I withdraw my comment. It certainly doesn't hurt to have them there. Just a matter of convention and/or consistency.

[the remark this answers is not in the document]
karllark: Ok.  Will leave them as they are.

hbushouse (approved):
Looks like all review comments have been addressed. @cshanahan1 are you good with this now?

ddavis-stsci (approved):
The roman regression tests pass with that PR (and pinning jsonschema==4.9.0) so I think we're good to go. 
The regression test is at: https://plwishmaster.stsci.edu:8081/me/my-views/view/all/job/RT/job/romancal/607/

cshanahan1 (approved):

hbushouse:
OK, it sounds like everyone is on-board with this PR and hence I'm merging it now.

jdavies-st:
I'm trying to understand this PR and the behavior of `JumpStep` here.  This allows flagging a pixel in a subsequent group after the initial jump is detected in that pixel.

The main need for this is because in many readout patterns, groups are actually averages of many frames, and the jump occurs within the group, not between groups.

If the jump happens within the group (i.e. between frames in a group that are then averaged), then it corrupts two differences between groups.  This also makes them harder to detect because they can have half the amplitude.

Given the 2-point difference method, is there a reason to believe one should flag the group before or after?  Is it more sensitive to the first or the second difference corrupted by the jump within the group?  If we can't tell, wouldn't it be best to throw out both, i.e. the group before and after the detected jump?  Or is the current model more likely to find the first difference affected, and therefore it's sufficient to flag only the subsequent group?

So given the above, generally, `if model.meta.exposure.nframes > 1`, i.e. there are averaged frames in a group, then one should _always_ turn on after jump flagging with a very low DN threshold?

The other driver for this I suspect is persistence or leakage of charge from saturated pixels in a jump into unsaturated areas in subsequent reads?  One sees this easily around the saturated cores of large events (snowballs) in the NIR detectors.

Just trying to understand these parameters so I can tune them to the problems in the data I'm seeing.

karllark:
Also note that the transient is seen for multiple groups after a ramp jump, so not just one.  For the really large jumps, the transient may go on for quite a few groups.

hbushouse:
@karllark Remind me: is this "transient" after a CR hit only seen in the mid-IR detectors, or near-IR as well?

karllark:
That analysis I did was for the MIR detectors.  Discussions with others (e.g., Karl Misselt) indicated that such transients are also seen on the NIR detectors, especially for the large CRs.

jdavies-st:
@karllark, is the transient charge seen in subsequent groups after a CR hit different from general persistence due to high signal in MIRI?  I.e. is the delta function charge at N counts different from N counts due to a bright star the build linearly-ish?  I suspect so, but also wonder if this after_jump flagging is really flagging persistence.

On the NIR detectors, it is clearly persistence from saturated (or near saturated) pixels in previous groups that then persist for subsequent groups and often subsequent exposures up 2000 to 3000 seconds later.  It doesn't seem to matter whether it's a core of a bright store or a core of a snowball CR. It persists.

For the NIR case, it seems more sensible to take care of this in the persistence step by either masking or modeling it, hopefully exposure-to-exposure at some point via level3 processing.

PR #112, opened 2022-07-28
JP-2669: Fixing the NaN and DO_NOT_USE bug.

kmacdonald-stsci:
Updating handling NaNs in the rateints product, as well as the DQ flags associated with those NaNs.

Resolves [JP-2669](https://jira.stsci.edu/browse/JP-2669)

Closes #

This PR addresses NaNs in the rateints product file and the DQ flags associated with them.  In the rateints product, NaNs were present with `Good` DQ flags.  They shouldn't be.  

Further, INS requested no NaNs be present in the rateints file.  This PR detects NaNs in the rateints file and sets them to 0.0, then sets the corresponding DQ flag to `DO_NOT_USE`.  

Lastly, the rateints DQ flags were set considering only `JUMP_DET` and `SATURATED`.  This has now been changed to consider `DO_NOT_USE` as well.  Specifically, if all groups in an integration for a pixel are flagged as DO_NOT_USE, the corresponding DQ value will be flagged as DO_NOT_USE.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/src/stcal/ramp_fitting/ols_fit.py
+++ b/src/stcal/ramp_fitting/ols_fit.py
@@ -535,7 +535,6 @@ def slice_ramp_data(ramp_data, start_row, nrows):
     ramp_data_slice = ramp_fit_class.RampData()
 
     # Slice data by row
-    # XXX change the type of copy
     data = ramp_data.data[:, :, start_row:start_row + nrows, :].copy()
     err = ramp_data.err[:, :, start_row:start_row + nrows, :].copy()
     groupdq = ramp_data.groupdq[:, :, start_row:start_row + nrows, :].copy()
@@ -1179,7 +1178,9 @@ def ramp_fit_overall(
         variances_ans, save_opt, tstart):
     """
     Computes the final/overall slope and variance values using the
-    intermediate computations previously computed.
+    intermediate computations previously computed.  When computing
+    integration slopes, if NaNs are computed, the corresponding DQ
+    flag is set to DO_NOT_USE and the value is set to 0. per INS.
 
 
     Parameters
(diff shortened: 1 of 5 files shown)


dmggh (approved):
LGTM, including your documentation updates.

hbushouse:
Be aware that there's related discussion going on in JP-2692 as to what values to assign to pixels that have no usable data (i.e. NaN vs. zero vs. ???). I believe this is scheduled for Cal WG discussion next week, so this might need updating based on the outcome of that discussion.

hbushouse (approved):

src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1273,13 +1274,20 @@ def ramp_fit_overall(
     # Suppress, then re-enable harmless arithmetic warnings
     warnings.filterwarnings("ignore", ".*invalid value.*", RuntimeWarning)
     warnings.filterwarnings("ignore", ".*divide by zero.*", RuntimeWarning)
+
     slope_int = the_num / the_den
+
+    # Adjust DQ flags for NaNs.
+    wh_nans = np.where(np.isnan(slope_int))
```
karllark, line 1281: Minor query: is the `np.where` even needed?  My understanding is that the following lines would work without it.

karllark (approved):
Looks good to me.  Very minor comment.

hbushouse:
Merging this in order to fix the bug that was leading to incorrect results. If the Cal WG decides later that pixels with no good data should be set to something other than zero, we'll make that change when necessary.

nden:
This was merged without an approval by the Roman team. 
@ddavis-stsci fyi: I started the roman devdeps regtest job.

hbushouse:
> This was merged without an approval by the Roman team. @ddavis-stsci fyi: I started the roman devdeps regtest job.

Whoops, my bad!

kmacdonald-stsci:
I expect this to cause failures for regtests, since it changes NaNs to 0.0 and sets corresponding DQ flag to DO_NOT_USE.

PR #115, opened 2022-08-11
JP-2817: GLS doesn't properly handle bad gain values

kmacdonald-stsci:
Resolves [JP-2817](https://jira.stsci.edu/browse/JP-2817)

This PR addresses bad gain values.  Negative and NaN values should result in setting DQ flags for DO_NOT_USE and NO_GAIN_VALUE.  Also, NaN gain values results in NaNs in the final image product, which should not happen.  NaNs should be set to zero in the final image.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,9 +22,10 @@ ramp_fitting
 - Updating how NaNs and DO_NOT_USE flags are handled in the rateints
   product. [#112]
 
-- Updating how GLS handles bad gain values.  NaNs found in the image
-  data are set to 0.0 and the corresponding DQ flag is set to 
-  DO_NOT_USE. [#115]
+- Updating how GLS handles bad gain values.  NaNs and negative gain
+  values have the DO_NOT_USE and NO_GAIN_VALUE flag set.  Any NaNs
+  found in the image data are set to 0.0 and the corresponding DQ flag
+  is set to DO_NOT_USE. [#115]
 
 Changes to API
 --------------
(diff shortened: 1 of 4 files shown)


dmggh (approved):
Looks good to me.

src/stcal/ramp_fitting/gls_fit.py
```diff
@@ -656,30 +658,17 @@ def gls_fit_single(ramp_data, gain_2d, readnoise_2d, max_num_cr, save_opt):
     else:
         gls_opt_info = None
 
-    '''
+    # Create new model...
```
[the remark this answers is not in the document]
kmacdonald-stsci: No.  This is an old comment from before I refactored removing data model usage.  I changed the comment.

CHANGES.rst
```diff
@@ -22,6 +22,8 @@ ramp_fitting
 - Updating how NaNs and DO_NOT_USE flags are handled in the rateints
   product. [#112]
 
+- Updating how GLS handles bad gain values. [#115]
```
hbushouse, line 25: Could include a few more details here on what the new handling is, such as the fact that pixels with bad gain values are now flagged as bad and set to zero.

hbushouse (approved):

PaulHuwe (approved):
Passes RomanCAL unit & regression tests.

CHANGES.rst
kmacdonald-stsci: I expanded this to more fully flesh it out.

PR #118, opened 2022-09-11
JP-2645 - Snowball and shower flagging in Jump

mwregan2:
Resolves [JP-2645](https://jira.stsci.edu/browse/JP-2645)
Resolves [RCAL-nnnn](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes #

This PR addresses ...

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


CHANGES.rst
```diff
@@ -1,7 +1,11 @@
 1.1.1 (unreleased)
 ==================
 
--
+jump
+----
+
+- Changes to flag both NIR snowballs and MIRI showers
+for  JP-#264. [#118]
```
hbushouse, line 8: typo here: JP ticket number should be 2645

src/stcal/jump/twopoint_difference.py
```diff
@@ -165,8 +165,6 @@ def find_crs(dataa, group_dq, read_noise, rejection_thresh,
 
         log_str = 'From highest outlier, two-point found {} pixels with at least one CR from {} groups.'
         log.info(log_str.format(len(row4cr), 'five or more'))
-        log.info(log_str.format(len(row3cr), 'four'))
```
hbushouse, line 168: Any particular reason why you're removing this logging info? Is it no longer relevant?

src/stcal/jump/jump.py
```diff
@@ -18,7 +19,13 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
                  after_jump_flag_dn1=0.0,
                  after_jump_flag_n1=0,
                  after_jump_flag_dn2=0.0,
-                 after_jump_flag_n2=0):
+                 after_jump_flag_n2=0,
```
hbushouse, line 22: I assume there will be a corresponding update to jwst/jump_step.py to make these parameters available to the user?

src/stcal/jump/jump.py
```diff
@@ -108,6 +115,30 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
...
+    min_jump_area : float
+        The minimum contour area to trigger the creation of enclosing ellipses or circles.
+
+    expand_factor : float
+        The factor that is used to increase the size of the enclosing circle/ellipse jump
+        flagged pixels.
+
+    use_ellipses : bool
+        If true the minimum enclosing ellipse will be created for jump regions that meet the area
+        requirement. This is best for MIRI showers which are only rarly circular. For the NIR detectors
```
hbushouse, line 131: minor typo: "rarly"

src/stcal/jump/jump.py
```diff
@@ -250,6 +282,11 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
             previous_row_above_gdq = row_above_gdq.copy()
             k += 1
         elapsed = time.time() - start
+        if expand_large_events:
```
hbushouse, line 285: Might want to just add a comment in the code here that this is the call that flags snowballs

src/stcal/jump/jump.py
```diff
@@ -250,6 +282,11 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
             previous_row_above_gdq = row_above_gdq.copy()
             k += 1
         elapsed = time.time() - start
```
hbushouse, line 284: Is this computation of `elapsed` ever actually used or reported anywhere? I see it's repeated again down in line 291 where it is printed to the log.

src/stcal/jump/jump.py
```diff
@@ -262,3 +299,131 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
...
+                        creation of a snowball.
+    :param min_jump_area:  The minimum area of a contour to cause the creation of the minimum
+                        enclosing ellipse or circle
+    :param expand_factor: The factor that increases the size of the snowball or enclosing ellipse.
+    :param use_ellipses:  Use ellipses rather than circles (better for MIRI)
+    :param sat_required_snowball: Require that there is a saturated pixel within the radius of the jump
+                        circle to trigger the formation of a snowball.
+    :return: none
+    """
+    for integration in range(gdq.shape[0]):
```
hbushouse, line 326: Interesting: so this whole routine relies only on information contained in the previously set DQ flags and not the actual pixel data. (nothing wrong with that - just find it interesting!)

src/stcal/jump/jump.py
```diff
@@ -262,3 +299,131 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
...
+        for group in range(1, gdq.shape[1]):
+            if use_ellipses:
+                jump_ellipses = find_ellipses(gdq[integration, group, :, :], jump_flag, min_jump_area)
+                log.info(f"Integration {integration} group {group} number of showers = {len(jump_ellipses)}.")
```
hbushouse, line 330: This log message will contain 0-indexed values of integration and group. Just checking to make sure that's what you want, as opposed to 1-indexed.

tests/test_jump.py
```diff
@@ -0,0 +1,65 @@
...
+    plane[2, 3] = DQFLAGS['JUMP_DET']
+    plane[2, 1] = DQFLAGS['JUMP_DET']
+    plane[1, 3] = DQFLAGS['JUMP_DET']
+    plane[2, 4] = DQFLAGS['JUMP_DET']
+    plane[3, 3] = DQFLAGS['JUMP_DET']
+    ellipse = find_ellipses(plane, DQFLAGS['JUMP_DET'], 1)
+    assert ellipse[0][2] == pytest.approx(45.0, 1e-3) # 90 degree rotation
+    assert ellipse[0][0] == pytest.approx((2.5, 2.0))  # center
+
+@pytest.mark.skip(reason="only for local testing")
```
hbushouse, line 56: Ah, was just about to comment on the use of FITS files in the test, but I see that it's set to be skipped during normal runs. Never mind!

hbushouse (commented):
Looks good overall. Just a few minor comments/corrections.

hbushouse:
I assume that docs on the jwst side will be updated to include info about these new capabilities.

hbushouse:
@ddavis-stsci Can you (or a Roman designee) review this for compatibility with Roman pipeline use?

ddavis-stsci (approved):
This passes the roman regression tests (https://plwishmaster.stsci.edu:8081/me/my-views/view/all/job/RT/job/Roman-devdeps/135/) so I don't see a problem with it.

src/stcal/jump/jump.py
```diff
@@ -262,3 +299,131 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
...
+    image = np.zeros(shape=(plane.shape[0], plane.shape[1], 3), dtype=np.uint8)
+    num_circles = len(snowballs)
+    sat_pix = np.bitwise_and(plane, 2)
+    for snowball in snowballs:
+        jump_radius = snowball[1]
+        jump_center = snowball[0]
+        cenx = jump_center[1]
+        ceny = jump_center[0]
+        extend_radius = round(jump_radius * expansion)
+        image = cv.circle(image, (round(ceny), round(cenx)), extend_radius, (0, 0, 4), -1)
```
cshanahan1, line 366: why are the coordinates for the center of the circle being rounded?

cshanahan1 (commented):
lgtm so far, i am going to test it out on some data before i hit approve just to get a better idea of how this works

CHANGES.rst
mwregan2: fixed

src/stcal/jump/twopoint_difference.py
mwregan2: With multiprocessing this generates a lot of output that > 99% of the time doesn't show any jumps in the 2 and 3 group semi-ramps. It annoyed me. So, I thought it would annoy most other people too.

src/stcal/jump/jump.py
mwregan2: added

src/stcal/jump/jump.py
mwregan2: The combination of the SAT flag and the JUMP flag provides enough information to correctly flag over 90% of the large events. In the future, I think we could increase the fidelity of the flagging using the magnitude of the jumps but I wanted to start simple since this is a major change.

src/stcal/jump/jump.py
mwregan2: The first group doesn't have any jumps. The key reason it starts at index 1 is that I am using the change in the saturation flags to search for the core of the snowballs. That means I have to use the delta from the previous group. So, the loops needs to start at an index of 1.

src/stcal/jump/jump.py
mwregan2: Because cv.circle requires them to be integers. I think the work is done by C routines which need specific types.

src/stcal/jump/jump.py
mwregan2: I think it is a relic from the split off of STCAL. The actual time calculation is done in the JWST code now. I'll remove it.

mwregan2: No, I was wrong the calculation is here but there was redundant code. I removed it.

hbushouse (approved):
OK, I think it all looks good to me now.

hbushouse:
@cshanahan1 have you had a chance to do any testing?

cshanahan1:
@hbushouse yes, will finish that up and add approval + review on corresponding jwst pr

src/stcal/jump/jump.py
```diff
@@ -249,7 +280,12 @@ def detect_jumps(frames_per_group, data, gdq, pdq, err,
             # save the neighbors to be flagged that will be in the next slice
             previous_row_above_gdq = row_above_gdq.copy()
             k += 1
-        elapsed = time.time() - start
+        #  This is the flag that controls the flagging of either snowballs or showers.
+        if expand_large_events:
```
cshanahan1, line 284: i think this block needs to go above so that when multiprocessing is used, the snowball dq flags make it back in to the reconstructed arrays. (i am still trying to verify this is true so i may end up deleting this comment)

mwregan2: I'm not clear about what you are proposing. This section has to be after the full dq arrays have been reconstructed. I've done all of my work using multiprocessing. 
The snowball flags are added to the full dq arrays.

cshanahan1: oh sorry you're right. the snowball flagging doesn't actually use multiprocessing since it's called after

cshanahan1 (approved):
line 348 should be changed, a simple subtraction might cause issues in some cases, but mike has tested this as-is and its fine for now.

PR #121, opened 2022-10-04
JP-2938: Fix total frames computation in dark current

### LINKED ISSUE #120: Bug in comparison of science and dark ramp lengths in stcal/dark_current/dark_sub.py
Line 99 of https://github.com/spacetelescope/stcal/blob/main/src/stcal/dark_current/dark_sub.py is:

`sci_total_frames = sci_ngroups * (sci_nframes + sci_groupgap)`

but it should be

`sci_total_frames = sci_ngroups * sci_nframes + (sci_ngroups - 1) * sci_groupgap`

because there is one less gap than the number of groups.

This was discovered using NIRCam DEEP8 readout with sci_ngroups = 7. The dark reference file has 130 groups which is 2 frames more than the science data. However, the pipeline refused to attempt dark subtraction because the error above led to:
`Not enough data in dark reference file to match to science data.`

hbushouse:
Resolves [JP-2938](https://jira.stsci.edu/browse/JP-2938)

Closes #120 

This PR fixes a bug in the computation of the total number of data frames in science and dark ramps when nframes > 1 and/or groupgap > 0.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@
 Bug Fixes
 ---------
 
+dark_current
+~~~~~~~~~~~~
+
+- Bug fix for computation of the total number of frames when science data
+  use on-board frame averaging and/or group gaps. [#121]
+
 ramp_fitting
 ~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


hbushouse:
@ddavis-stsci I'll leave it to Roman folks to decide whether this change warrants a full Roman regression test run.

ddavis-stsci (approved):
Looks good and the roman tests pass,
https://plwishmaster.stsci.edu:8081/me/my-views/view/all/job/RT/job/Roman-devdeps/144/

cshanahan1 (approved):

PR #124, opened 2022-11-14
RCAL-406: Array quantities For RomanCAL

PaulHuwe:
Resolves [RCAL-406](https://jira.stsci.edu/browse/RCAL-nnnn)

Closes https://github.com/spacetelescope/romancal/issues/582

This PR adds support for Quantities in models required for the RomanCAL pipeline. These changes have been tested against both JWST and Roman unit tests.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)


PaulHuwe:
This passes all unit tests throughout the RomanCAL pipeline and the unit tests in jwst.

ddavis-stsci (approved):
LGTM

src/stcal/dark_current/dark_class.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+from astropy import units as u
```
kmacdonald-stsci, line 2: Is this a standard import idiom for astropy units?  If not, using just 'u' may be too terse.

kmacdonald-stsci (approved):

src/stcal/dark_current/dark_class.py
PaulHuwe: It is: https://docs.astropy.org/en/stable/units/index.html

schlafly (approved):
Looks good to me.  One thing to think about: do we really want these comparisons to be type(x) == u.Quantity?  I usually use isinstance(x, u.Quantity) in these cases to catch subclasses (though I'm not sure if any such subclasses currently exist).  I could also doing something more duck-type-y along the lines of:
```
try:
    x.data = model.data.value
except AttributeError:
   x.data = model.data
```
which would be more flexible for things implementing quantity-like interfaces but could have unintended consequences.

Anyway, some thoughts, but looks good to me to merge.

PaulHuwe:
> Looks good to me. One thing to think about: do we really want these comparisons to be type(x) == u.Quantity? I usually use isinstance(x, u.Quantity) in these cases to catch subclasses (though I'm not sure if any such subclasses currently exist). I could also doing something more duck-type-y along the lines of:
> 
> ```
> try:
>     x.data = model.data.value
> except AttributeError:
>    x.data = model.data
> ```
> 
> which would be more flexible for things implementing quantity-like interfaces but could have unintended consequences.
> 
> Anyway, some thoughts, but looks good to me to merge.

I like this and I will implement the change.

PaulHuwe:
I suspect I lack merge permissions for STCAL - could someone merge this?

PR #125, opened 2022-11-16
JP-2988: Set saturation only if fully saturated.

kmacdonald-stsci:
Resolves [JP-2988](https://jira.stsci.edu/browse/JP-2988)

Closes #

This PR addresses setting the saturation flag.  In the rateints product, the saturation flag will only be set if the ramp for that pixel is fully saturated, i.e., saturation starts in group 0.  In the rate product, the saturation flag will only be set if the saturation flag is set for each integration.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/docs/stcal/ramp_fitting/description.rst
+++ b/docs/stcal/ramp_fitting/description.rst
@@ -284,3 +284,18 @@ For the optional output product, the variance of the slope due to the Poisson
 noise of the segment-specific slope is written to the VAR_POISSON extension.
 Similarly, the variance of the slope due to the read noise of the
 segment-specific slope  is written to the VAR_RNOISE extension.
+
+Data Quality Propagation
+=================
+
+If all groups in an integration are flagged as DO_NOT_USE or SATURATED, then
+the rateints product will be flagged as DO_NOT_USE.  For the rates product, the
+DO_NOT_USE flag will be set if all integrations are flagged as DO_NOT_USE.
+
+If all groups in an integration are flaggd as SATURATED, then the rateints
+product will be flagged as SATURATED and DO_NOT_USE.  If all integrations are
+flagged as SATURATED, the rates product will be flagged as SATURATED and DO_NOT_USE.
+
+If a group in an integration is flagged as JUMP_DET, then rateints will be
+flagged as JUMP_DET.  If any integration is flagged as JUMP_DET, the rates
+product will be flagged as JUMP_DET.
--- a/docs/stcal/ramp_fitting/description.rst
+++ b/docs/stcal/ramp_fitting/description.rst
@@ -286,16 +286,26 @@ Similarly, the variance of the slope due to the read noise of the
 segment-specific slope  is written to the VAR_RNOISE extension.
 
 Data Quality Propagation
-=================
+========================
+For a given pixel, if all groups in an integration are flagged as DO_NOT_USE or
+SATURATED, then that pixel will be flagged as DO_NOT_USE in the corresponding
+integration in the rateints product.  Note this does NOT mean that all groups
+are flagged as SATURATED, nor that all groups are flagged as DO_NOT_USE.  For
+example, suppressed one ramp groups will be flagged as DO_NOT_USE in the
+zeroeth group, but not necessarily any other group, while only groups one and
+on are flagged as SATURATED.  Further, only if all integrations in the rateints
+product are marked as DO_NOT_USE, then the pixel will be flagged as DO_NOT_USE
+in the rate product.
 
-If all groups in an integration are flagged as DO_NOT_USE or SATURATED, then
-the rateints product will be flagged as DO_NOT_USE.  For the rates product, the
-DO_NOT_USE flag will be set if all integrations are flagged as DO_NOT_USE.
+For a given pixel, if all groups in an integration are flagged as SATURATED,
+then that pixel will be flagged as SATURATED and DO_NOT_USE in the corresponding
+integration in the rateints product.  This is different from the above case in
+that this is only for all groups flagged as SATURATED, not for some combination
+of DO_NOT_USE and SATURATED.  Further, only if all integrations in the rateints
+product are marked as SATURATED, then the pixel will be flagged as SATURATED
+and DO_NOT_USE in the rate product.
 
-If all groups in an integration are flaggd as SATURATED, then the rateints
-product will be flagged as SATURATED and DO_NOT_USE.  If all integrations are
-flagged as SATURATED, the rates product will be flagged as SATURATED and DO_NOT_USE.
-
-If a group in an integration is flagged as JUMP_DET, then rateints will be
-flagged as JUMP_DET.  If any integration is flagged as JUMP_DET, the rates
-product will be flagged as JUMP_DET.
+For a given pixel, if any group in an integration is flagged as JUMP_DET, then
+that pixel will be flagged as JUMP_DET in the corresponding integration in the
+rateints product.  Also, that pixel will be flagged as JUMP_DET in the rate
+product.
--- a/src/stcal/ramp_fitting/utils.py
+++ b/src/stcal/ramp_fitting/utils.py
@@ -1414,7 +1414,10 @@ def dq_compress_sect(ramp_data, num_int, gdq_sect, pixeldq_sect):
 
     # XXX JP-2988
     # If all groups are set to SATURATED, mark as SATURATED.
-    is_flag_set_for_all_groups(pixeldq_sect, gdq_sect, sat_flag)
+    # A group 0 marked as saturated implies all groups are saturated.
+    gdq0_sat = np.bitwise_and(gdq_sect[0], sat_flag)
+    pixeldq_sect[gdq0_sat != 0] = np.bitwise_or(
+        pixeldq_sect[gdq0_sat != 0], sat_flag)
 
     # If jump occures mark the appropriate flag.
     cr_loc_r = np.bitwise_and(gdq_sect, jump_flag)
--- a/src/stcal/ramp_fitting/utils.py
+++ b/src/stcal/ramp_fitting/utils.py
@@ -1290,7 +1290,9 @@ def dq_compress_final(dq_int, ramp_data):
     """
     Combine the integration-specific dq arrays (which have already been
     compressed and combined with the PIXELDQ array) to create the dq array
-    of the primary output product.
+    of the primary output product.  The saturation flag is set only if the
+    saturation flag is set for all integrations; in this case, the do not
+    use flag is also set.
 
     Parameters
     ----------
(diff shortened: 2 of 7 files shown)


src/stcal/ramp_fitting/utils.py
```diff
@@ -1368,24 +1405,51 @@ def dq_compress_sect(ramp_data, num_int, gdq_sect, pixeldq_sect):
...
-    sat_loc_r = np.bitwise_and(gdq_sect, sat_flag)
-    sat_loc_im = np.where(sat_loc_r.sum(axis=0) > 0)
-    pixeldq_sect[sat_loc_im] = np.bitwise_or(pixeldq_sect[sat_loc_im], sat_flag)
+    # If all groups are set to SATURATED, mark as SATURATED.
+    # A group 0 marked as saturated implies all groups are saturated.
+    gdq0_sat = np.bitwise_and(gdq_sect[0], sat_flag)
+    pixeldq_sect[gdq0_sat != 0] = np.bitwise_or(
+        pixeldq_sect[gdq0_sat != 0], sat_flag)
 
+    # If jump occures mark the appropriate flag.
```
dmggh, line 1416: should be "occurs"

src/stcal/ramp_fitting/utils.py
```diff
@@ -1317,28 +1319,63 @@ def dq_compress_final(dq_int, dnu_flag):
...
-    locations, and set the SATURATED flag in those locations in the PIXELDQ
-    array. Similarly, get the ramp locations where the data has been flagged as
-    a jump detection in the 4D GROUPDQ array, find the corresponding image
-    locations, and set the JUMP_DET flag in those locations in the PIXELDQ
-    array. These modifications to the section of the PIXELDQ array are not used
-    to flag groups for any computations; they are used only in the integration-
-    specific output.
+    This sets the integration level flags for DO_NOT_USE, JUMP_DET and
+    SATURATED.  If any ramp has a jump, this flag will be set for the
+    integraion.  If all groups in a ramp are flagged as DO_NOT_USE, then the
```
dmggh, line 1376: should be "integration"

src/stcal/ramp_fitting/utils.py
```diff
@@ -1317,28 +1319,63 @@ def dq_compress_final(dq_int, dnu_flag):
...
+    # saturated, so should be marked as DO_NOT_USE.
+    sat_loc = np.zeros(f_dq.shape, dtype=f_dq.dtype)
+    sat_loc[np.where(np.bitwise_and(f_dq, sat_flag))] = 1
+    f_dq[sat_loc == 1] = np.bitwise_or(f_dq[sat_loc == 1], dnu_flag)
 
     return f_dq
 
 
+def is_flag_set_for_all_ints(dq_int, f_dq, flag):
+    """
```
dmggh, line 1339: I had to stare at this function and its calls for a while to discern what it does. Maybe a bit more in the function description would help clarify it.

src/stcal/ramp_fitting/utils.py
```diff
@@ -1368,24 +1405,51 @@ def dq_compress_sect(ramp_data, num_int, gdq_sect, pixeldq_sect):
...
+    # If jump occures mark the appropriate flag.
     cr_loc_r = np.bitwise_and(gdq_sect, jump_flag)
     cr_loc_im = np.where(cr_loc_r.sum(axis=0) > 0)
     pixeldq_sect[cr_loc_im] = np.bitwise_or(pixeldq_sect[cr_loc_im], jump_flag)
 
     return pixeldq_sect
 
 
+def is_flag_set_for_all_groups(pixeldq_sect, gdq_sect, flag):
+    """
```
dmggh, line 1425: Like the is_flag_set for_all_ints( ) function above, a bit more explanation in the docs would be helpful for this function.

tests/test_ramp_fitting.py
```diff
@@ -614,12 +617,11 @@ def test_one_group_ramp_suppressed_two_integrations():
...
     # Check slopes information
     sdata, sdq, svp, svr, serr = slopes
-    sat, dnu = dqflags["SATURATED"], dqflags["DO_NOT_USE"]
 
     check = np.array([[1.0000001, 1.0000001, 1.0000002]])
     np.testing.assert_allclose(sdata, check, tol)
 
-    check = np.array([[sat, sat, 0]])
+    check = np.array([[GOOD, GOOD, GOOD]])
     np.testing.assert_allclose(sdq, check, tol)
```
dmggh, line 625: Comparing with the previous version, should this be np.array([[SAT, SAT, GOOD]]) or is this a different test ? This also applies to another later test.

dmggh (commented):
Looks good aside from just a few mostly minor comments.

PaulHuwe (approved):
This works with the Roman Pipeline. We don't have a test for this specific scenario, so a ticket has been created to add that.

src/stcal/ramp_fitting/utils.py
```diff
@@ -1286,20 +1286,22 @@ def compute_slices(max_cores):
...
 
-def dq_compress_final(dq_int, dnu_flag):
+def dq_compress_final(dq_int, ramp_data):
     """
     Combine the integration-specific dq arrays (which have already been
     compressed and combined with the PIXELDQ array) to create the dq array
-    of the primary output product.
+    of the primary output product.  The saturation flag is set only if the
+    saturation flag is set for all integrations; in this case, the do not
+    use flag is also set.
```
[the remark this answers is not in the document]
kmacdonald-stsci: I made this change.

src/stcal/ramp_fitting/utils.py
```diff
@@ -1317,28 +1319,63 @@ def dq_compress_final(dq_int, dnu_flag):
...
+        Cube of the compressed pixel DQ per integration.  It is
+        a 3-D cube with dimensions (nints, nrows, ncols).
+
+    f_dq : ndarray
+        The flattened DQ array that is only 2-D, (nrows, ncols),
+        containing the Or'ing of each integration.
+
+    flag : int
+        The DQ flag of interest
+    """
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed the return, since that return value is not being used.

src/stcal/ramp_fitting/utils.py
kmacdonald-stsci: Changed the misspelling.

src/stcal/ramp_fitting/utils.py
kmacdonald-stsci: Changed this misspelling.

src/stcal/ramp_fitting/utils.py
kmacdonald-stsci: I changed the docstring, as well as added more comments to clarify the purpose of this function.

tests/test_ramp_fitting.py
kmacdonald-stsci: Since the way the SATURATED flag gets set the expected value for the SATURATED flag.  As the test is written, with the new behavior of setting the SATURATED flag in the rate file, the new expected output is `[GOOD, GOOD, GOOD]`.

dmggh (approved):
Looks good to me.

hbushouse (approved):
I *think* this all looks good. The real test is in the test results, i.e. if we see the desired behavior for the various scenarios.

hbushouse:
Because this is new flagging behavior, docs will also need to be updated, but that may need a jwst PR.

nden:
Because this also affects Roman, it's best if the documentation change is here and jwst links to it.
Unfortunately, there're no docs on saturation in stcal. 🤷‍♀️

hbushouse:
> Because this also affects Roman, it's best if the documentation change is here and jwst links to it. Unfortunately, there're no docs on saturation in stcal. 🤷‍♀️

But this isn't a change to the saturation step. It's a change to ramp_fit in terms of how it handles flags set by the saturation step.

docs/stcal/ramp_fitting/description.rst
```diff
@@ -284,3 +284,18 @@ For the optional output product, the variance of the slope due to the Poisson
 noise of the segment-specific slope is written to the VAR_POISSON extension.
 Similarly, the variance of the slope due to the read noise of the
 segment-specific slope  is written to the VAR_RNOISE extension.
+
+Data Quality Propagation
+=================
```
hbushouse, line 289: This line needs to have the same number of characters as the line above.

docs/stcal/ramp_fitting/description.rst
```diff
@@ -284,3 +284,18 @@ For the optional output product, the variance of the slope due to the Poisson
...
+
+If all groups in an integration are flagged as DO_NOT_USE or SATURATED, then
+the rateints product will be flagged as DO_NOT_USE.  For the rates product, the
```
hbushouse, line 292: Not the entire rateints product. The plane of the DQ array corresponding to that integration will be flagged.  Also, "the rates product" should be "the *rate* product".

docs/stcal/ramp_fitting/description.rst
```diff
@@ -284,3 +284,18 @@ For the optional output product, the variance of the slope due to the Poisson
...
+DO_NOT_USE flag will be set if all integrations are flagged as DO_NOT_USE.
+
+If all groups in an integration are flaggd as SATURATED, then the rateints
```
hbushouse, line 295: Not obvious on first reading how this second case differs from first case. The wording above "If all groups in an integration are flagged as DO_NOT_USE or SATURATED" sounds like it already covers this subset case where all groups are flagged as SATURATED. In the first one, do you mean a mixture of either DO_NOT_USE or SATURATED, such that all groups are flagged with one or the other (e.g. groups 0 and 1 are DO_NOT_USE and all the rest are SATURATED)?

kmacdonald-stsci: I have updated this section to try to make it more clear.  Please let me know if what is there now is more understandable.

schlafly:
For a partially saturated ramp, are any flags set after this PR?  i.e., was SATURATED the only way to detect a partially saturated ramp?  It does seem to me that partially saturated ramps are qualitatively different from other ramps and that they should be identifiable from the data quality flags.

schlafly:
You have a better sense than I do as to how accessible the ramp product is.  Let me mention a specific scenario where a DQ flag indicating partial saturation is needed:
In the long discussion here
https://jira.stsci.edu/browse/JP-2256?focusedId=612618&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-612618
about partially saturated sources & inverse variance weighting of L3 products, I suggest treating partially saturated pixels differently from other pixels.   The natural way to do this would be to use a DQ flag.

I think I understand that with this PR it's now possible to have a partially saturated pixel with no DQ flags in the primary L2 product.  That seems bad to me.  I see in the JWST calibration working group notes that the solution of adding a new PARTIAL_SATURATION flag was suggested and likewise the recommendation was to go back to the full ramp fitting product for this information.  In my opinion this case is definitely important enough to deserve a DQ flag in the rateint product---judging by the action on that issue partial saturation is an important case that people consider!  I'm happy to be overruled, but I thought I would at least weigh in.

kmacdonald-stsci:
In the JIRA ticket for this PR there is a link to the meeting minutes from the Cal WG where you can read the details of what was discussed and agreed to.  All agreed the ramp product will still have groups flagged as SATURATED, but if a ramp is only partially saturated the SATURATED flag is not to be propagated to the rateints, nor rate product.

hbushouse:
Is there anything preventing this from getting merged?  It would be good to get this into an stcal release, so that the jwst side of things can then get tested and merged as well.

nden:
👍 to merging if @kmacdonald-stsci thinks this is ready. Roman will need to deal with this in some way later.

kmacdonald-stsci:
This is ready to merge.  Once this is merged, the JWST PR will need to be checked, especially regression tests as this will break regression tests since the SATURATED flag will be set differently than before.

hbushouse:
Needs an stcal release.

PR #126, opened 2022-11-21
remove dependency on `opencv-python`

zacharyburnett:
Resolves [AL-691](https://jira.stsci.edu/browse/AL-691)

This PR removes unconditional dependency on `opencv-python` added in a previous PR; this was causing installation issues on some users' platforms.

I am also exploring replacing usage of `opencv-python` with `scikit-image`

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ General
 -------
 
 - Moved build configuration from ``setup.cfg`` to ``pyproject.toml`` to support PEP621 [#95]
-
+- made dependency on ``opencv-python`` conditional [#126]
 
 ramp_fitting
 ~~~~~~~~~~~~
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ dependencies = [
     'astropy >=5.0.4',
     'scipy >=1.6.0',
     'numpy >=1.17',
-    'opencv-python >=4.6.0.66',
 ]
 dynamic = ['version']
 
@@ -35,6 +34,9 @@ test = [
     'pytest-cov',
     'pytest-doctestplus',
     'pytest-openfiles >=0.5.0',
+]
+opencv = [
+    'opencv-python >=4.6.0.66',
 ]
 
 [project.urls]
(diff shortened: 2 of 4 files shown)


zacharyburnett:
`opencv-python` is currently used in four functions in `jump.py`:
https://github.com/spacetelescope/stcal/blob/33aa4f4c808a22ca4d822b91b5d6aa8fc6969b83/src/stcal/jump/jump.py#L380-L398
https://github.com/spacetelescope/stcal/blob/33aa4f4c808a22ca4d822b91b5d6aa8fc6969b83/src/stcal/jump/jump.py#L401-L427
https://github.com/spacetelescope/stcal/blob/33aa4f4c808a22ca4d822b91b5d6aa8fc6969b83/src/stcal/jump/jump.py#L430-L437
https://github.com/spacetelescope/stcal/blob/33aa4f4c808a22ca4d822b91b5d6aa8fc6969b83/src/stcal/jump/jump.py#L440-L449

zacharyburnett:
looking into replacing shape construction with `scikit-image`, it appears there is a crucial difference in the drawing method:
```python
import cv2
import numpy
from matplotlib import pyplot
from skimage import draw

image_0 = numpy.full((10, 10), fill_value=0, dtype=numpy.uint8)

center = (4, 4)
radius = 4

disk = draw.disk(center, radius)
image_1 = image_0.copy()
image_1[disk] = 1

image_2 = cv2.circle(image_0.copy(), center, radius, 1, thickness=-1)

figure, axes = pyplot.subplots(nrows=1, ncols=2)

axes[0].imshow(image_1)
axes[0].set_title('scikit-image')

axes[1].imshow(image_2)
axes[1].set_title('opencv')

pyplot.show()
```
Figure_1

zacharyburnett:
@hbushouse for the goal of removing dependence on `opencv-python`, do you think I should explore replacing with `scikit-image`, or is a conditional import for the affected functions (the current state of this PR) sufficient for this use case?

hbushouse:
> @hbushouse for the goal of removing dependence on `opencv-python`, do you think I should explore replacing with `scikit-image`, or is a conditional import for the affected functions (the current state of this PR) sufficient for this use case?

For now let's just stick with the conditional import contained in this PR. It appears the replacement with `scikit-image` will take more work to ensure the same results.

hbushouse (approved):
Looks OK to me, but there's a CI failure (code style) that needs fixing.

PR #131, opened 2022-12-08
JP-3014: This Sets Invalid Slope Data to NaN in the Rate and Rateints products.

kmacdonald-stsci:
Resolves [JP-3014](https://jira.stsci.edu/browse/JP-3014)

Closes #

This PR addresses invalid slope data found through flagging.  Specifically, pixels flagged as SATURATED or DO_NOT_USE have no valid data.  Currently, that data is set to zero, but zero can be valid data, so this data is not set NaN.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/docs/stcal/ramp_fitting/description.rst
+++ b/docs/stcal/ramp_fitting/description.rst
@@ -135,6 +135,13 @@ is the following: the type of noise (when appropriate) will appear as the supers
 and the form of the data will appear as the subscript: ‘s’, ‘i’, ‘o’ for segment,
 integration, or overall (for the entire dataset), respectively.
 
+It is possible for an integration or pixel to have invalid data, so useable
+slope data will not be available.  If a pixel has an invalid integration, the value
+for that integration for that pixel will be set to NaN in the rateints product.
+Further, if all integrations for a given pixel are invalid the pixel value for
+the rate product will be set to NaN.  An example of invalid data would be a
+fully saturated integration for a pixel.
+
 Optimal Weighting Algorithm
 ---------------------------
 The slope of each segment is calculated using the least-squares method with optimal
--- a/src/stcal/ramp_fitting/ols_fit.py
+++ b/src/stcal/ramp_fitting/ols_fit.py
@@ -1280,7 +1280,6 @@ def ramp_fit_overall(
     # Adjust DQ flags for NaNs.
     wh_nans = np.isnan(slope_int)
     dq_int[wh_nans] = np.bitwise_or(dq_int[wh_nans], ramp_data.flags_do_not_use)
-    slope_int[wh_nans] = 0.
     warnings.resetwarnings()
 
     del the_num, the_den, wh_nans
@@ -1347,7 +1346,7 @@ def ramp_fit_overall(
 
     # Output integration-specific results to separate file
     integ_info = utils.output_integ(
-        slope_int, dq_int, effintim, var_p3, var_r3, var_both3)
+        ramp_data, slope_int, dq_int, effintim, var_p3, var_r3, var_both3)
 
     if opt_res is not None:
         del opt_res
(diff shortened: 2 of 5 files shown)


src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1366,6 +1365,12 @@ def ramp_fit_overall(
     #   primary output
     final_pixeldq = utils.dq_compress_final(dq_int, ramp_data)
 
+    # For invalid slope calculations set to NaN.  Pixels flagged as SATURATED or
+    # DO_NOT_USE have invalid data.
+    invalid_data = ramp_data.flags_saturated | ramp_data.flags_do_not_use
```
hbushouse, line 1370: I assume this is applied *after* the point at which saturation flags are weeded out for pixels that are only partially saturated?  In other words, at this point a pixel will only have the saturation flag set if all groups are saturated, right?

hbushouse (commented):
Looks OK. Just a couple questions for clarification.

src/stcal/ramp_fitting/ols_fit.py
kmacdonald-stsci: Yes.  You can see in the lines above that the final DQ computations are done in `final_pixeldq`.  After the final DQ computations are done, those flags are then used to identify invalid slops computed, which at this line in the code is represented by `c_rates`.  This computation on `c_rates` is the last computation done before returning it as the computed slopes.

src/stcal/ramp_fitting/ols_fit.py
```diff
@@ -1366,6 +1365,12 @@ def ramp_fit_overall(
...
+    wh_invalid = np.where(np.bitwise_and(final_pixeldq, invalid_data))
+    c_rates[wh_invalid] = np.nan
```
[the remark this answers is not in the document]
kmacdonald-stsci: Integration level NaN's are set in `utils.py` at line 1136.  This function deals with fully saturated ramps.  I will check other areas of the code dealing with integration level computations to see if NaN's should also be set elsewhere.

hbushouse (approved):

src/stcal/ramp_fitting/ols_fit.py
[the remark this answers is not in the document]
kmacdonald-stsci: I added a further check in `utils.output_integ` for integration level setting of  NaN's.  This did not affect any CI tests.

hbushouse:
Is there anything in the ramp_fit docs that might need updating, i.e. do we say anywhere that pixels without slopes will be set to some particular value (previously zero, now NaN)?

kmacdonald-stsci:
> Is there anything in the ramp_fit docs that might need updating, i.e. do we say anywhere that pixels without slopes will be set to some particular value (previously zero, now NaN)?

I updated the "description" documentation for ramp fitting detailing the setting of NaN for invalid data.

PaulHuwe (approved):
All roman pipeline unit and regression tests work with this.

docs/stcal/ramp_fitting/description.rst
```diff
@@ -135,6 +135,13 @@ is the following: the type of noise (when appropriate) will appear as the supers
 and the form of the data will appear as the subscript: ‘s’, ‘i’, ‘o’ for segment,
 integration, or overall (for the entire dataset), respectively.
 
+It is possible for an integration or pixel to have invalid data, so useable
+slope data will be available.  If a pixel has an invalid integration, the value
```
hbushouse, line 139: don't you mean "will NOT be available"?

docs/stcal/ramp_fitting/description.rst
```diff
@@ -135,6 +135,13 @@ is the following: the type of noise (when appropriate) will appear as the supers
...
+for that integration for that pixel will be set to NaN in the rateints product.
+Further, if all integrations for a given pixel are invalid the pixel value for
+the rates product will be set to NaN.  An example of invalid data would be a
```
hbushouse, line 142: "rates" should be "rate" (singular)

docs/stcal/ramp_fitting/description.rst
kmacdonald-stsci: I fixed this.

docs/stcal/ramp_fitting/description.rst
kmacdonald-stsci: I fixed this.

hbushouse:
Looks good to me. CI failures appear to be dozens of CRDS access failures. Shall we merge?

kmacdonald-stsci:
> Looks good to me. CI failures appear to be dozens of CRDS access failures. Shall we merge?

I think so.

PR #134, opened 2022-12-15
refactor CI

zacharyburnett:
This PR applies some fixes to `tox.ini` and renames the `toxenv`s to be more in line with other repositories. Additionally, it rearranges the GitHub Actions workflow to use the `toxenvs` and reduce duplication of the configuration in the YAML file by adding `include`s to the matrix for specific `toxenv`s. This orders the CI as follows:
image

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


hbushouse:
What is the purpose and goal of this PR? It's not stated anywhere in the description.

zacharyburnett:
> What is the purpose and goal of this PR? It's not stated anywhere in the description.

I apologize, I thought I had filled this out. Added now

kmacdonald-stsci (approved):

hbushouse:
Looks OK, but do we know why the test-jwst-xdist and test-romancal-xdist CI runs have failed and/or not run at all?

hbushouse:
I'm also seeing warnings in the CI logs about `opencv-python`. I recall changes were made recently to the way that's imported. Are those changes causing problems here (in the CI tests)?  I seem to recall other people mentioning that there are times when they can no longer get to the imported opencv functions either, even when just running interactively in python.

zacharyburnett:
> Looks OK, but do we know why the test-jwst-xdist and test-romancal-xdist CI runs have failed and/or not run at all?

It looks like the downstream tests are [also failing on the main branch](https://github.com/spacetelescope/stcal/actions/runs/3729922740/jobs/6326747184); the other discrepancy with runs was because the expected status checks were `test-romancal/jwst-xdist-cov` and I had `test-romancal/jwst-xdist`. I added the coverage in a commit.

hbushouse:
There seem to be huge number failures of the type:
`FileNotFoundError: [Errno 2] No such file or directory: '/grp/crds/cache/mappings/jwst/jwst_1023.pmap'` in the failing CI tests. Does the CI environment not have access to the CRDS directories or something?

hbushouse (approved):
Well, the CI tests for jwst now pass, so I'm happy. Romancal is still failing ...

PR #138, opened 2022-12-29
add fallback to `scikit-image` for ellipse functions

zacharyburnett:
This PR adds fallback functions for ellipse construction in case `opencv-python` is not installed.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -1,6 +1,7 @@
 import logging
 import multiprocessing
 import time
+import warnings
 
 import numpy as np
 
@@ -13,8 +14,15 @@ try:
 
     ELLIPSE_PACKAGE = 'opencv-python'
 except (ImportError, ModuleNotFoundError):
-    ELLIPSE_PACKAGE_WARNING = '`opencv-python` must be installed (`pip install stcal[opencv]`) ' \
-                              'in order to use ellipses'
+    try:
+        import skimage.draw
+        import skimage.measure
+
+        ELLIPSE_PACKAGE = 'scikit-image'
+        ELLIPSE_PACKAGE_WARNING = f'`opencv-python` not installed; using `scikit-image` for ellipse construction'
+    except (ImportError, ModuleNotFoundError):
+        ELLIPSE_PACKAGE_WARNING = '`opencv-python` must be installed (`pip install stcal[ellipse]`) ' \
+                                  'in order to use ellipses'
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
@@ -402,6 +410,10 @@ def extend_snowballs(plane, snowballs, sat_flag, jump_flag, expansion=1.5):
         color = (0, 0, 4)
         if ELLIPSE_PACKAGE == 'opencv-python':
             image = cv.circle(image, center, extend_radius, color, -1)
+        elif ELLIPSE_PACKAGE == 'scikit-image':
+            warnings.warn(ELLIPSE_PACKAGE_WARNING)
+            disk = skimage.draw.disk(center, extend_radius)
+            image[disk] = color
         else:
             raise ModuleNotFoundError(ELLIPSE_PACKAGE_WARNING)
         jump_circle = image[:, :, 2]
@@ -437,6 +449,10 @@ def extend_ellipses(plane, ellipses, sat_flag, jump_flag, expansion=1.1):
         color = (0, 0, 4)
         if ELLIPSE_PACKAGE == 'opencv-python':
             image = cv.ellipse(image, center, axes, alpha, 0, 360, color, -1)
+        elif ELLIPSE_PACKAGE == 'scikit-image':
+            warnings.warn(ELLIPSE_PACKAGE_WARNING)
+            ellipse = skimage.draw.ellipse(*center, *axes, rotation=alpha)
+            image[ellipse] = color
         else:
             raise ModuleNotFoundError(ELLIPSE_PACKAGE_WARNING)
         jump_ellipse = image[:, :, 2]
@@ -455,7 +471,8 @@ def find_circles(dqplane, bitmask, min_area):
         bigcontours = [con for con in contours if cv.contourArea(con) >= min_area]
         circles = [cv.minEnclosingCircle(con) for con in bigcontours]
     else:
-        raise ModuleNotFoundError(ELLIPSE_PACKAGE_WARNING)
+        # raise ModuleNotFoundError(ELLIPSE_PACKAGE_WARNING)
+        raise ModuleNotFoundError('`opencv-python` required for this functionality')
     return circles
 
 
@@ -470,7 +487,8 @@ def find_ellipses(dqplane, bitmask, min_area):
         # with just 4 points.
         ellipses = [cv.minAreaRect(con) for con in bigcontours]
     else:
-        raise ModuleNotFoundError(ELLIPSE_PACKAGE_WARNING)
+        # raise ModuleNotFoundError(ELLIPSE_PACKAGE_WARNING)
+        raise ModuleNotFoundError('`opencv-python` required for this functionality')
     return ellipses
 
 
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -19,7 +19,8 @@ except (ImportError, ModuleNotFoundError):
         import skimage.measure
 
         ELLIPSE_PACKAGE = 'scikit-image'
-        ELLIPSE_PACKAGE_WARNING = f'`opencv-python` not installed; using `scikit-image` for ellipse construction'
+        ELLIPSE_PACKAGE_WARNING = '`opencv-python` not installed; ' \
+                                  'using `scikit-image` for ellipse construction'
     except (ImportError, ModuleNotFoundError):
         ELLIPSE_PACKAGE_WARNING = '`opencv-python` must be installed (`pip install stcal[ellipse]`) ' \
                                   'in order to use ellipses'
--- a/tests/test_jump.py
+++ b/tests/test_jump.py
@@ -10,12 +10,12 @@ try:
     import cv2  # noqa: F401
 
     ELLIPSE_PACKAGE_INSTALLED = True
-except ImportError:
+except (ImportError, ModuleNotFoundError):
     try:
         import skimage  # noqa: F401
 
         ELLIPSE_PACKAGE_INSTALLED = True
-    except:
+    except (ImportError, ModuleNotFoundError):
         ELLIPSE_PACKAGE_INSTALLED = False
 
 
--- a/tests/test_jump.py
+++ b/tests/test_jump.py
@@ -7,11 +7,16 @@ from stcal.jump.jump import flag_large_events, find_circles, find_ellipses
 DQFLAGS = {'JUMP_DET': 4, 'SATURATED': 2, 'DO_NOT_USE': 1}
 
 try:
-    import cv2 as cv # noqa: F401
+    import cv2  # noqa: F401
 
-    OPENCV_INSTALLED = True
+    ELLIPSE_PACKAGE_INSTALLED = True
 except ImportError:
-    OPENCV_INSTALLED = False
+    try:
+        import skimage  # noqa: F401
+
+        ELLIPSE_PACKAGE_INSTALLED = True
+    except:
+        ELLIPSE_PACKAGE_INSTALLED = False
 
 
 @pytest.fixture(scope='function')
@@ -31,7 +36,7 @@ def setup_cube():
     return _cube
 
 
-@pytest.mark.xfail(not OPENCV_INSTALLED, reason="`opencv-python` not installed")
+@pytest.mark.xfail(not ELLIPSE_PACKAGE_INSTALLED, reason="`opencv-python` not installed")
 def test_find_simple_circle():
     plane = np.zeros(shape=(5, 5), dtype=np.uint8)
     plane[2, 2] = DQFLAGS['JUMP_DET']
@@ -43,7 +48,7 @@ def test_find_simple_circle():
     assert circle[0][1] == pytest.approx(1.0, 1e-3)
 
 
-@pytest.mark.xfail(not OPENCV_INSTALLED, reason="`opencv-python` not installed")
+@pytest.mark.xfail(not ELLIPSE_PACKAGE_INSTALLED, reason="`opencv-python` not installed")
 def test_find_simple_ellipse():
     plane = np.zeros(shape=(5, 5), dtype=np.uint8)
     plane[2, 2] = DQFLAGS['JUMP_DET']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,10 @@ jobs:
             python-version: '3.8'
           - toxenv: test-opencv-xdist
             os: ubuntu-latest
-            python-version: '3.x'
+            python-version: '3.10'
+          - toxenv: test-scikitimage-xdist
+            os: ubuntu-latest
+            python-version: '3.10'
           - toxenv: test-jwst-xdist-cov
             os: ubuntu-latest
             python-version: '3.x'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ test = [
     'pytest-doctestplus',
     'pytest-openfiles >=0.5.0',
 ]
-opencv = [
+ellipse = [
     'opencv-python >=4.6.0.66',
 ]
 
--- a/tox.ini
+++ b/tox.ini
@@ -46,13 +46,13 @@ description =
     run tests
     jwst: of JWST pipeline
     romancal: of Romancal pipeline
-    opencv: requiring opencv-python
+    ellipse: requiring opencv-python
     warnings: treating warnings as errors
     cov: with coverage
     xdist: using parallel processing
 extras =
     test
-    opencv: opencv
+    ellipse: ellipse
 deps =
     xdist: pytest-xdist
     jwst: jwst[test] @ git+https://github.com/spacetelescope/jwst.git
@@ -75,7 +75,7 @@ commands =
     jwst: --pyargs jwst --ignore-glob=timeconversion --ignore-glob=associations \
     romancal: --pyargs romancal \
     cov: --cov=. --cov-config=pyproject.toml --cov-report=term-missing --cov-report=xml \
-    opencv: -- tests/test_jump.py \
+    ellipse: -- tests/test_jump.py \
     {posargs}
 
 [testenv:build-docs]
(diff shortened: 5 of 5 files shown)


zacharyburnett:
Contour finding also needs to be ported:
```python
```

zacharyburnett:
As described in https://github.com/spacetelescope/stcal/pull/126#issuecomment-1324012508, there are some differences in the methods used to construct ellipses between the two packages which result in different image masks. Perhaps a warning should make this more clear.

zacharyburnett:
It looks like the contours found by `scikit-image` are consistently larger, by about half a unit, than those found by `opencv`

comparison

```python
import matplotlib.patches
import numpy
from matplotlib import pyplot

import stcal.jump.jump
from stcal.jump.jump import find_circles, find_ellipses
from tests.test_jump import DQFLAGS

if __name__ == '__main__':
    circle_plane = numpy.zeros(shape=(5, 5), dtype=numpy.uint8)
    circle_plane[2, 2] = DQFLAGS['JUMP_DET']
    circle_plane[3, 2] = DQFLAGS['JUMP_DET']
    circle_plane[1, 2] = DQFLAGS['JUMP_DET']
    circle_plane[2, 3] = DQFLAGS['JUMP_DET']
    circle_plane[2, 1] = DQFLAGS['JUMP_DET']

    ellipse_plane = numpy.zeros(shape=(5, 5), dtype=numpy.uint8)
    ellipse_plane[2, 2] = DQFLAGS['JUMP_DET']
    ellipse_plane[3, 2] = DQFLAGS['JUMP_DET']
    ellipse_plane[1, 2] = DQFLAGS['JUMP_DET']
    ellipse_plane[2, 3] = DQFLAGS['JUMP_DET']
    ellipse_plane[2, 1] = DQFLAGS['JUMP_DET']
    ellipse_plane[1, 3] = DQFLAGS['JUMP_DET']
    ellipse_plane[2, 4] = DQFLAGS['JUMP_DET']
    ellipse_plane[3, 3] = DQFLAGS['JUMP_DET']

    opencv_circles = find_circles(circle_plane, DQFLAGS['JUMP_DET'], 1)
    opencv_ellipses = find_ellipses(ellipse_plane, DQFLAGS['JUMP_DET'], 1)

    stcal.jump.jump.ELLIPSE_PACKAGE = 'scikit-image'

    skimage_circles = find_circles(circle_plane, DQFLAGS['JUMP_DET'], 1)
    skimage_ellipses = find_ellipses(ellipse_plane, DQFLAGS['JUMP_DET'], 1)

    figure, (circle_axis, ellipse_axis) = pyplot.subplots(1, 2)
    figure.set_size_inches(9, 9 / 1.618)

    circle_axis.set_title('circle')
    circle_axis.imshow(circle_plane)
    for circle in opencv_circles:
        circle_axis.add_patch(matplotlib.patches.Circle(*circle, color='b', fill=False, label='opencv'))
    for circle in skimage_circles:
        circle_axis.add_patch(matplotlib.patches.Circle(*circle, color='r', fill=False, label='skimage'))
    circle_axis.legend()

    ellipse_axis.set_title('ellipse')
    ellipse_axis.imshow(ellipse_plane)
    for ellipse in opencv_ellipses:
        ellipse_axis.add_patch(
            matplotlib.patches.Rectangle(
                (ellipse[0][0], ellipse[0][1] - 1.5),
                *ellipse[1],
                ellipse[2],
                color='b',
                fill=False,
                label='opencv',
            )
        )
    for ellipse in skimage_ellipses:
        ellipse_axis.add_patch(
            matplotlib.patches.Rectangle(
                (ellipse[0][0], ellipse[0][1] - 2),
                *ellipse[1],
                ellipse[2],
                color='r',
                fill=False,
                label='skimage',
            )
        )
    ellipse_axis.legend()

    pyplot.show()
```

zacharyburnett:
expanding the ellipse test case to a more oblong shape reveals a few edge cases in the `scikit-image` method; I'll convert this to draft for now and work on that

zacharyburnett:
adding some more complexity to the test shapes:
comparison

tests/test_jump.py
```diff
@@ -40,23 +45,25 @@ def test_find_simple_circle():
     plane[2, 3] = DQFLAGS['JUMP_DET']
     plane[2, 1] = DQFLAGS['JUMP_DET']
     circle = find_circles(plane, DQFLAGS['JUMP_DET'], 1)
+    assert circle[0][0] == pytest.approx((2, 2))
     assert circle[0][1] == pytest.approx(1.0, 1e-3)
```
zacharyburnett, line 49: here is where the fit circle from the `scikit-image` contour differs from the `opencv-python` circle; the radius is 1.5 instead of 1

zacharyburnett:
Currently, the `.minAreaRect()` is replaced by `.minimum_rotated_rectangle` from Shapely; this could be replaced by an like in `circle.py` to remove the need for the additional dependency

kmacdonald-stsci (approved):

zacharyburnett:
should get input from INS on whether this fallback is acceptable

jdavies-st (commented):
This adds a lot of code, a lot of complexity (two libraries attempting to do the same thing), and a future maintenance headache, and when we're trying to solve the headache of adding `opencv` as a dependency.  Best to get rid of `opencv` altogether and pick a single library that works and is low maintenance.  I would recommend using astropy `regions`, as it can create pixel maps from circles and ellipses, which seems to be the only thing `opencv` is being used for here?

Though I feel fitting circles and ellipses in this algorithm is problematic, as the fit can go wrong, spectacularly, as it is poorly constrained.  An ellipse can grow to arbitrary size, and I've seen a quarter of the MIRI detector masked (incorrecty) because of this.  Better to use binary operations such as eroding and dilating the mask, functions that are available in `scipy`, a dependency we already carry, and already demonstrated to work well in `grizli`

https://github.com/gbrammer/grizli/commit/93a570ba5364ca70284ae6d07f4c24ee5a90b1e7#diff-24ba9312ac84f8067ea7213d6d20d9eda4703b229185b25b4a382196514fa4a4R2769

mwregan2:
James,
  Can region detect ellipses? The detection step is key. I didn’t want to reinvent the wheel and CV was the first package I found that met my requirements. Looking at the Regions API I see than it could do the second half of the process of extending and masking pixels but not the first half of finding the best fit ellipse.
  The live version suffers from fitting circles to what are highly elliptical snowballs (a rare feature) in the NIR. My dev version fixes that by always fitting ellipses. If there are incorrect ellipses being fit, please show me an example. I’d like to figure out why that would happen.
  I’ll switch to any package that does what we need.

From: James Davies ***@***.***>
Date: Monday, February 6, 2023 at 4:38 AM
Cc: Michael Regan ***@***.***>, Review requested ***@***.***>
Subject: Re: [spacetelescope/stcal] add fallback to `scikit-image` for ellipse functions (PR #138)


@jdavies-st commented on this pull request.

This adds a lot of code, a lot of complexity (two libraries attempting to do the same thing), and a future maintenance headache, and when we're trying to solve the headache of adding opencv as a dependency. Best to get rid of opencv altogether and pick a single library that works and is low maintenance. I would recommend using astropy regions, as it can create pixel maps from circles and ellipses, which seems to be the only thing opencv is being used for here?

Though I feel fitting circles and ellipses in this algorithm is problematic, as the fit can go wrong, spectacularly, as it is poorly constrained. An ellipse can grow to arbitrary size, and I've seen a quarter of the MIRI detector masked (incorrecty) because of this. Better to use binary operations such as eroding and dilating the mask, functions that are available in scipy, a dependency we already carry, and already demonstrated to work well in grizli

***@***.***#diff-24ba9312ac84f8067ea7213d6d20d9eda4703b229185b25b4a382196514fa4a4R2769<https://urldefense.com/v3/__https:/github.com/gbrammer/grizli/commit/93a570ba5364ca70284ae6d07f4c24ee5a90b1e7*diff-24ba9312ac84f8067ea7213d6d20d9eda4703b229185b25b4a382196514fa4a4R2769__;Iw!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!zO7HXykaGzhYp1chIVVF-1pRaUoy_vuVIrw4DLzHST85ZwVpJFEIF2LhoRlmI-cmytbja9t6rOwe-mg-ezsS6Uk$>

zacharyburnett:
I removed `opencv-python` entirely in this PR. We can look into alternative methods using `scipy` and binary closure / dilation + erosion.

src/stcal/jump/jump.py
```diff
@@ -658,32 +673,30 @@ def extend_saturation(
...
-                -1,
+            ellipse_mask = skimage.draw.ellipse(
+                r=round(ceny),
+                c=round(cenx),
+                r_radius=round(axis1/2),
+                c_radius=round(axis2/2),
+                shape=image.shape,
+                rotation=alpha,
             )
+            image[*ellipse_mask, 2] = 22
```
zacharyburnett, line 684: replaces `cv.ellipse` with a mask derived from `skimage.draw.ellipse`, then applies Blue 22 to the image array (why is the blue channel set to 22 if it's discarded right after? can this just be a simple mask?)

src/stcal/jump/jump.py
```diff
@@ -658,32 +673,30 @@ def extend_saturation(
...
-                -1,
+            ellipse_mask = skimage.draw.ellipse(
+                r=round(ceny),
+                c=round(cenx),
+                r_radius=round(ellipse[1][0] / 2),
+                c_radius=round(ellipse[1][1] / 2),
+                shape=persist_image.shape,
+                rotation=alpha,
             )
+            persist_image[*ellipse_mask, 2] = 22
```
zacharyburnett, line 699: same as above; why is this blue?

src/stcal/jump/jump.py
```diff
@@ -734,16 +747,10 @@ def extend_ellipses(
...
-            alpha,
-            0,
-            360,
-            (0, 0, jump_flag),
-            -1,
-        )
+        center = (round(ceny), round(cenx))
+        axes = (round(axis1 / 2), round(axis2 / 2))
+        rr, cc = skimage.draw.ellipse(*center, *axes, shape=image.shape, rotation=alpha)
+        image[rr, cc, 2] = 4
```
zacharyburnett, line 753: is there a reason why this is rounded to the nearest integer?

src/stcal/jump/jump.py
```diff
@@ -779,26 +811,110 @@ def find_last_grp(grp, ngrps, num_grps_masked):
...
+    best_idx = np.argmin(areas)
+
+    # return the best box
+    x1 = max_x[best_idx]
+    x2 = min_x[best_idx]
+    y1 = max_y[best_idx]
+    y2 = min_y[best_idx]
+    r = rotations[best_idx]
+
+    rval = np.zeros((4, 2))
```
zacharyburnett, line 871: this function replaces `shapely.geometry.Polygon.minimum_bounding_box`

src/stcal/jump/jump.py
```diff
@@ -779,26 +811,110 @@ def find_last_grp(grp, ngrps, num_grps_masked):
...
+
+
+def area_of_polygon(xy: np.ndarray) -> float:
+    """
+    apply shoelace algorithm on collection of xy vertex pairs
+    https://stackoverflow.com/questions/24467972/calculate-area-of-polygon-given-x-y-coordinates
+    """
+    return 0.5 * np.abs(
+        np.dot(xy[:, 0], np.roll(xy[:, 1], 1)) - np.dot(xy[:, 1], np.roll(xy[:, 0], 1))
+    )
```
zacharyburnett, line 887: this function replaces `shapely.geometry.Polygon.area`

src/stcal/jump/jump.py
```diff
@@ -779,26 +811,110 @@ def find_last_grp(grp, ngrps, num_grps_masked):
...
+
+
+def find_circles(dqplane: np.ndarray, bitmask: np.ndarray, min_area: float) -> list[Circle]:
```
zacharyburnett, line 890: added type hints here

src/stcal/jump/jump.py
```diff
@@ -1037,20 +1153,37 @@ def find_faint_extended(
...
+                minimum_bounding_rectangle(con) for con in bigcontours
+            ]
+            ellipses = [
+                (
+                    tuple(np.mean(rectangle[[0, 2], :], axis=0)),
+                    tuple(np.hypot(*np.diff(rectangle[[0, 1, 2], :], axis=0))),
+                    np.degrees(np.arctan2(*np.flip(np.diff(rectangle[[3, 0], :], axis=0)[0]))),
+                )
+                for rectangle in rectangles
+            ]
```
zacharyburnett, line 1171: this is where the axis differences comes from; the minor and major radii are larger than those from `opencv` by 1.5 (?) units

zacharyburnett (commented):
the formatter made some irrelevant edits, so here are the relevant changes to this PR

src/stcal/jump/jump.py
mwregan2: No real reason. It's hack to create a mask of the ellipse. Sorry that's it is confusing.

src/stcal/jump/jump.py
mwregan2: Same reason, a hack.

src/stcal/jump/jump.py
mwregan2: If I remember right the routine is expecting integers.  My guess was this is just an interface to C code.
Eventually we have to more to integers to mask the pixels anyway.

src/stcal/jump/jump.py
mwregan2: Remember that the goal is to generate an ellipse. The ellipse can be specified by the minimum bounding box. I have to say that the format that CV uses for the bounding ellipse is very confusing. Hopefully, your version will be easier to understand.

zacharyburnett:
opencv_flag_large_events_withsnowball
skimage_flag_large_events_withsnowball

differences in `flag_large_events` from `test_flag_large_events_withsnowball`

zacharyburnett:
comparison of results from `test_find_faint_extended` (`opencv` on the left, `skimage` on the right):

zacharyburnett:
to summarize, two tests still show differences between the two methods; `test_flag_large_events_withsnowball` and `test_find_faint_extended`. I've added comparison images, above.

mwregan2:
I don’t understand what I’m looking at


From: Zach Burnett ***@***.***>
Date: Thursday, June 6, 2024 at 12:51 PM
Cc: Michael Regan ***@***.***>, Review requested ***@***.***>
Subject: Re: [spacetelescope/stcal] replace `opencv-python` with `scikit-image` for snowball detection (PR #138)


comparison of results from test_find_faint_extended (opencv on the left, skimage on the right):
opencv_test_find_faint_extended.png (view on web)<https://urldefense.com/v3/__https:/github.com/spacetelescope/stcal/assets/16024299/e20bc052-a1f2-454f-b761-103781f487cc__;!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!zoeP76EdhjR4CrgsLzrhmnUTWGdhHtDYy_1OXPuqbmu5Stz2qu4d-044BXnXDpUu618wwMy2DxTWaDEM612-mKU$> skimage_test_find_faint_extended.png (view on web)<https://urldefense.com/v3/__https:/github.com/spacetelescope/stcal/assets/16024299/2cc4a402-1471-4d38-859c-4739cd95022b__;!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!zoeP76EdhjR4CrgsLzrhmnUTWGdhHtDYy_1OXPuqbmu5Stz2qu4d-044BXnXDpUu618wwMy2DxTWaDEM9oAMH2g$>

zacharyburnett:
> I don’t understand what I’m looking at

Sorry about that! I'll update the comments to be more descriptive

perrygreenfield:
To use your first simple example. I find that if I give the radius for draw.disk as 4.0001 I get the same image as cv2. So it wouldn't surprise me that in one the test is an < test and the other <=. Or something like that. Adding a small epsilon may do it.

zacharyburnett:
a comparison of the `find_contours` functions between `opencv` and `skimage`:
minimum_rotated_rectangle

and the `minimum_bounding_box` analogues:
minimum_rotated_rectangle

zacharyburnett:
started regression test run against this branch at https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST/detail/JWST/2925/pipeline

drlaw1558:
Hm, I'm testing this on jw02123001001_09101_00001_nrs1 and something doesn't seem to be working right.  I started out by merging the latest main branch in to make sure everything was synced, and then tried processing jw02123001001_09101_00001_nrs1

Runtime was significantly longer for the Scikit version; 743 seconds vs 314 seconds for the current opencv

More problematically though, the Scikit version doesn't seem to be flagging snowballs properly.  It's flagging lots of little things that don't need attention, but missing the big snowballs.  Perhaps I'm running the code the wrong way?  In both cases I'm just using strun to execute detector1 pipeline, with the default expand_large_events=True

scikit

pllim:
pyproject.toml needs to be updated to:

* remove opencv-python as a dependency
* unpin max version of python (namely, `python<3.14`) if it is still a thing by the time we circle back to this PR

mwregan2:
Zach,
  An interesting comparison. My first impression is the large crescent moon difference implies to me that the center of the ellipses are not the same. The residual highly elliptical DQ is because the original didn't have a continuous region that was consistent with one ellipse and instead had a series of smaller masks. I'd like to understand what happened there.
  Given that is a subset of one group in an exposure, the overall results will have a lot of differences. They may not matter but we should check.
  Meanwhile, I've been working on improvements for the opencv version. Most of the changes are tweaks to the parameters. I will not be doing any PRs. I'll just open tickets with the suggested changes.
  I'm out until next Thursday and will be able to respond more frequently.
Mike

________________________________

tapastro (approved):
Looks good! I'll get a fresh set of RTs running on jwst/main after merge, for okify purposes.

tests/test_jump.py
```diff
@@ -228,9 +228,9 @@ def test_find_ellipse2():
     ellipse = ellipses[0]
     assert ellipse[0][0] == 2
     assert ellipse[0][1] == 2
-    assert ellipse[1][0] == 2
-    assert ellipse[1][1] == 2
-    assert ellipse[2] == 90.0
+    assert ellipse[1][0] == pytest.approx(2.266, 1e-3)
+    assert ellipse[1][1] == pytest.approx(2.266, 1e-3)
+    assert ellipse[2] == pytest.approx(-0.785, 1e-3)
```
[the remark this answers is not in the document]
braingram: Maybe but isn't that equivalent to what is already here (which matches the other tests in this file that use `pytest.approx`)? Feel free to open up a follow up ticket to update the tests to use `assert_allclose`.

nden (approved):
I left two minor requests inline. Other than that it looks good to me.

PR #141, opened 2023-01-12
JP-3004: Should have NaN's for Suppressed Ramps, but Getting Zeros.

kmacdonald-stsci:
Resolves [JP-3004](https://jira.stsci.edu/browse/JP-3004)

This PR addresses zero values for suppressed ramps that should have been set to NaN in both the rate and rateints product.  There was an unexpected edge case due to the jump step setting JUMP_DET because of the `flag_4_neighbors` being set to true.  This resulted in an effective 2 group ramp with the second ramp being flagged as JUMP_DET, which is nonsense.   Further, in ramp fitting, because the `suppress_one_group` is set to true, this has the result in the first group being marked as DO_NOT_USE, while the second group is only a jump detection.  A one group ramp with the DO_NOT_USE flagged has different results to a one group ramp flagged as JUMP_DET, a situation unanticipated.

The solution was to suppress all groups in a one group ramp, rather than just the first good group.  This, then had unexpected results.  Internally, LARGE_VARIANCE is used to compute variances with bad data.  This allows the inverse sums to be computed, while adding effectively nothing to good variance computations.  However, when all variances are LARGE_VARIANCE, the resultant sum will be large and should be set to zero.  The threshold to check this was set too high, allowing these large variances through, when they should have been set to zero.  Also, this threshold was defined inconsistently in the code, so a top level variable in the `utils.py` was created to use in all instances this variance threshold needs to be used.

Finally, and unrelated to the issue in JP-3004, a warning thrown when the code tried to cast NaN as an integer has been suppressed to make the output cleaner.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,7 @@ Bug Fixes
 ramp_fitting
 ~~~~~~~~~~~~
 
-- Fixed zeros that should be NaNs in rateints product and suppressed
+- Fixed zeros that should be NaNs in rate and rateints product and suppressed
   a cast warning due to attempts to cast NaN to an integer. [#141]
 
 Changes to API
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,12 @@ General
 Bug Fixes
 ---------
 
--
+  
+ramp_fitting
+~~~~~~~~~~~~
+
+- Fixed zeros that should be NaNs in rateints product and suppressed
+  a cast warning due to attempts to cast NaN to an integer. [#141]
 
 Changes to API
 --------------
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -274,4 +274,5 @@ def suppress_one_good_group_ramps(ramp_data):
             # ramp saturated, mark the 0th groups as saturated, too.
             good_index = np.where(ramp_data.groupdq[integ, :, row, col] == 0)
             if ramp_data.groupdq[integ, good_index, row, col] == 0:
-                ramp_data.groupdq[integ, good_index, row, col] = dnu_flag
+                ramp_data.groupdq[integ, :, row, col] = np.bitwise_or(
+                 ramp_data.groupdq[integ, :, row, col], dnu_flag)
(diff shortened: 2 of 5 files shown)


src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -274,4 +274,5 @@ def suppress_one_good_group_ramps(ramp_data):
             # ramp saturated, mark the 0th groups as saturated, too.
             good_index = np.where(ramp_data.groupdq[integ, :, row, col] == 0)
             if ramp_data.groupdq[integ, good_index, row, col] == 0:
-                ramp_data.groupdq[integ, good_index, row, col] = dnu_flag
+                ramp_data.groupdq[integ, :, row, col] = np.bitwise_or(
```
dmggh, line 277: Line 267 - 'flagg'; and should that comment (and the comment that follows that) be changed to state that ALL groups will be suppressed ?

dmggh (approved):
Looks good, minor comment on comments.

Line 267 - 'flagg', and should this comment be changed to reflect that ALL groups will be suppressed ?

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I fixed the comments in this section to reflect the new change in ramp suppression.

hbushouse:
Regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/2168/ was clean, except for 1 unrelated failure (slow machine today resulted in the failure of a timed test).

hbushouse (approved):
LGTM.

CHANGES.rst
```diff
@@ -9,7 +9,12 @@ General
 Bug Fixes
 ---------
 
--
+  
+ramp_fitting
+~~~~~~~~~~~~
+
+- Fixed zeros that should be NaNs in rateints product and suppressed
```
hbushouse, line 16: Minor nit, but doesn't this change affect both rate and rateints products? Or was the result in the rate product already being set to NaN and the problem was just in the results for individual integrations?

kmacdonald-stsci: It affects both.  I updated the change log.

hbushouse:
Clean regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/549/ using this PR's git hash for stcal.

PaulHuwe (approved):
Roman Unit & Regression tests pass with these changes.

hbushouse:
Looks like everyone is in agreement on this, so I'm going to merge.

PR #145, opened 2023-03-01
JP-3075: Updating dynamic_mask function to include inversion functionality

kmacdonald-stsci:
…he 'dq_def' attribute to compress the DQ array.

Resolves [JP-3075](https://jira.stsci.edu/browse/JP-3075)

This PR addresses round tripping issues for datamodels that use `dq_def`.  The `dq_def` table details how DQ flags are compressed for saving and how to be uncompressed when opened.  When opened, datamodels automatically call `dynamic_mask` to properly uncompress DQ arrays.   However, the inverse was not true during save, which meant that if these datamodels got saved, the uncompressed DQ array got saved to disk, so that when opened again the decompression failed.

An `inv` parameter got added to the `dynamic_mask` function to compress uncompressed DQ arrays.  This inversion is used for datamodels that use `dq_def`.  In particular, in `stdatamodels` the inversion is used for ReferenceFileModel during save to ensure the uncompressed DQ array does not get saved, but is properly transformed using the `dq_def` table before saving.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Bug Fixes
 1.3.4 (2023-02-13)
 ==================
 
+- Updated the `dynamic_mask` function to include an inversion. [#145]
+
 ramp_fitting
 ~~~~~~~~~~~~
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,12 +4,13 @@
 Bug Fixes
 ---------
 
--
+Other
+-----
+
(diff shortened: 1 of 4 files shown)


WilliamJamieson:
spacetelescope/stdatamodels#134 and #146 coordinate moving `dynamicdq` out of `stcal` and into `stdatamodels`. I think its best to just wait for these PRs to merge, and then you can add the `dynamicdq` changes in this PR to spacetelescope/stdatamodels#132.

PR #151, opened 2023-03-09
import `tomli` as `tomllib` for `docs` builds before Python 3.11

zacharyburnett:
`tomli` is implemented read-only as `tomllib` in Python 3.11; this change ensures that the documentation build doesn't fail in Python 3.11 and above 

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -4,8 +4,11 @@ from datetime import datetime
 from pathlib import Path
 
 import stsci_rtd_theme
-import tomli
-
+if sys.version_info < (3, 11):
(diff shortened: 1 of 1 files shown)


zacharyburnett:
> I requested @WilliamJamieson for the second review (not sure why this PR said it requires 2 approving reviews).

I think that's a setting under the branch protections; I dunno whether we should enforce 1 or 2 approvals for PRs

WilliamJamieson (approved):

PR #153, opened 2023-03-14
use OpenAstronomy workflows and move MacOS jobs to schedule

zacharyburnett:
This PR abstracts CI workflows to the OpenAstronomy workflows; this should reduce maintenance for updating and maintaining actions. Additionally, these changes move the majority of MacOS jobs to the weekly scheduled workflow. This should alleviate the issue where the limited MacOS runners for the organization are not available for CI jobs.

image

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,85 +16,56 @@ on:
 
 jobs:
   check:
-    name: ${{ matrix.toxenv }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        toxenv: [ check-style, check-security, check-build ]
-        python-version: [ '3.x' ]
-        os: [ ubuntu-latest ]
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - uses: actions/setup-python@v4
-        with:
-          python-version: ${{ matrix.python-version }}
-          cache: 'pip'
-          cache-dependency-path: pyproject.toml
-      - run: pip install tox
-      - run: tox -e ${{ matrix.toxenv }}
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -16,85 +16,56 @@ on:
...
+        - linux: test-numpy120-xdist
+          python-version: '3.9'
+        - linux: test-numpy121-xdist
+          python-version: '3.9'
+        - linux: test-numpy122-xdist
+          python-version: '3.9'
+        - linux: test-numpy121-xdist
+          python-version: '3.10'
+        - linux: test-numpy122-xdist
+          python-version: '3.10'
```
WilliamJamieson, line 61: This seems excessive, it is also incomplete as `numpy` is up to version `1.24` I think.

It the max `numpy` version will be pulled into these tests naturally by `pip` during the `tox` environment creations. I suggest just doing a `minimum_dependencies` job, which will pull in the oldest dependencies (including `numpy`).

WilliamJamieson (changes requested):

pyproject.toml
```diff
@@ -14,7 +14,7 @@ classifiers = [
 dependencies = [
     'astropy >=5.0.4',
     'scipy >=1.6.0',
-    'numpy >=1.17',
+    'numpy >=1.20',
```
WilliamJamieson, line 17: I think a dependency change should probably be announced in the changelog (even if its really because other dependencies have pinned  the minimum value above this).

WilliamJamieson (changes requested):

WilliamJamieson (approved):

PR #154, opened 2023-03-16
JP-3132: Errors During Multiprocessing When More Processors are Requested than Rows in the Data

### LINKED ISSUE #152: max on an empty array raises a valueerror
In some cases when this line encounters a boolean array that has no true values the resulting array will have a 0 length, and then `max` will raise a valueerror.

I have a user which encounters this problem, but I am not fully sure why it happens.

https://github.com/spacetelescope/stcal/blob/6eff679acd6cf382b54fd0559b530d660dc5cc94/src/stcal/ramp_fitting/ols_fit.py#L3454

kmacdonald-stsci:
Resolves [JP-3132](https://jira.stsci.edu/browse/JP-3132)

Closes #152 

This PR addresses errors during processing when invoking multiprocessing requesting more processors than rows.  The data is sliced up by dividing the number of rows by the number of processors requested.  If the number of rows is less than the number of processors requested, this results in processors running on empty data sets with dimensions (nints, ngroups, 0, ncols).

To prevent this from occurring safeguards have been implemented in multiprocessing to prevent the usage of more processors than rows available.  The number of processors actually used will now be no more than the number of rows, regardless of the number of processors requested.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/tests/test_ramp_fitting.py
+++ b/tests/test_ramp_fitting.py
@@ -1073,9 +1073,9 @@ def test_multi_more_cores_than_rows():
                 ramp.data[integ, :, row, col] = bramp
                 bramp = bramp * factor
 
-    algo, save_opt, ncores = "OLS", False, "all"
+    bufsize, algo, save_opt, ncores = 512, "OLS", False, "all"
     slopes, cube, ols_opt, gls_opt = ramp_fit_data(
-        ramp, 512, save_opt, rnoise, gain, algo, "optimal", ncores, dqflags)
+        ramp, bufsize, save_opt, rnoise, gain, algo, "optimal", ncores, dqflags)
     # This part of the test is simply to make sure ramp fitting
     # doesn't crash.  No asserts are necessary here.
 
--- a/tests/test_ramp_fitting.py
+++ b/tests/test_ramp_fitting.py
@@ -1034,6 +1034,40 @@ def test_dq_multi_int_dnu():
     np.testing.assert_allclose(cerr, check, tol, tol)
 
 
+def test_multi_more_cores_than_rows():
+    """
+    This tests a (1, 10, 1, 2) dimensional dataset with multi-
+    processing using "all" to force all available processors to
+    be selected.  The data is divided by row.  Since there is
+    only one row, the number of processors actually used should
+    only be one.  Otherwise, there would be a crash as an empty
+    slice of the data would be sent through ramp fitting.
+    """
+    nints, ngroups, nrows, ncols = 2, 10, 1, 2
+    rnval, gval = 10., 5.
+    frame_time, nframes, groupgap = 10.736, 5, 0
+
+    dims = nints, ngroups, nrows, ncols
+    var = rnval, gval
+    tm = frame_time, nframes, groupgap
+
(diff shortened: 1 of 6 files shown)


PaulHuwe (approved):
This passes all RomanCAL unit and regression tests.

tests/test_ramp_fitting.py
```diff
@@ -1034,6 +1034,53 @@ def test_dq_multi_int_dnu():
...
+    NOTE: The following is useful only on computers with more than
+          one available processor.  Running this test on one
+          processor does NOT test the safety features of multi-
+          processing preventing the number of processors used
+          being no more than the number of processors requested.
+    """
+    ramp, gain, rnoise = create_blank_ramp_data(dims, var, tm)
+    bramp = np.array([ 150.4896,  299.7697,  449.0971,  600.6752,  749.6968,
+                       900.9771, 1050.1395, 1199.9658, 1349.9163, 1499.8358])
+    factor = 1.05
```
dmggh, line 1069: Why a factor of 1.05 ? Just to make the next ramp slightly higher ?

tests/test_ramp_fitting.py
```diff
@@ -1034,6 +1034,53 @@ def test_dq_multi_int_dnu():
...
+    for integ in range(nints):
+        for row in range(nrows):
+            for col in range(ncols):
+                ramp.data[integ, :, row, col] = bramp
+                bramp = bramp * factor
+
+    algo, save_opt, ncores = "OLS", False, "all"
+    slopes, cube, ols_opt, gls_opt = ramp_fit_data(
+        ramp, 512, save_opt, rnoise, gain, algo, "optimal", ncores, dqflags)
```
dmggh, line 1078: Maybe a note of what 512 signifies.

tests/test_ramp_fitting.py
```diff
@@ -1034,6 +1034,53 @@ def test_dq_multi_int_dnu():
...
+    # This part of the test is simply to make sure ramp fitting
+    # doesn't crash.  No asserts are necessary here.
```
dmggh, line 1080: How would a failure here be reported ? If not in the same format as other test's failures, perhaps an assert would be preferable for consistency.

dmggh (approved):
Looks good to me. Just a few questions.

hbushouse (approved):
LGTM

tests/test_ramp_fitting.py
kmacdonald-stsci: No reason.  Just a random number to use to set up data, same as the ramp data.

tests/test_ramp_fitting.py
kmacdonald-stsci: I updated this test.

tests/test_ramp_fitting.py
kmacdonald-stsci: The comment explains why there is no assert.  The ticket JP-3132 reported crashes during the conditions ramp fit is run in the during this test.  This test is an assurance that it doesn't crash.  Running `ramp_fit_data` with no problems is the test.

PR #162, opened 2023-04-14
drop support for Python 3.8

zacharyburnett:
Numpy is imminently dropping support for Python 3.8

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,9 +31,7 @@ jobs:
     with:
       envs: |
         - linux: test-oldestdeps-cov-xdist
-          python-version: 3.8
-        - linux: test-xdist
-          python-version: '3.8'
+          python-version: 3.9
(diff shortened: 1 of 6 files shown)


braingram:
roman (which still supports 3.8) uses stcal and tests against the development version.
https://github.com/spacetelescope/romancal/blob/18546a1592803eaed7fd6f5e91e3cbfab74abbbc/.github/workflows/roman_ci.yml#L82
It doesn't define a python version for these which I think means it uses '3.X'. @zacharyburnett do you know what version of python that becomes? I'm failing to find it in the open astronomy and associated docs:
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python

hbushouse (approved):
Needs approval from Roman.

PaulHuwe:
I'll do the Roman testing.

PR #163, opened 2023-04-19
JP-3221: Fixing Segment Length Computation Needed for Optimal Weighting

kmacdonald-stsci:
Correcting the way the number of groups in a segment gets calculated for the optimal weighting during the weighted OLS fitting for each segment.

Resolves [JP-3221](https://jira.stsci.edu/browse/JP-3221)

This PR addresses the computation of segment lengths for the purposes of optimal weighting.  Originally, this was done by counting the non-zero groups in an integration ramp of the masked data.  The masked data was created by multiplying a boolean array where contiguous `True` values defined the current segment being computed.  This multiplication had the side effect that `np.nan * False = np.nan`, so NaN groups were incorrectly being included in the computation of the number of groups in a segment.

Specifically, a test ramp with the group DQ flags as follows:

GDQ: [0, 0, 0, 4, 0, 0, 0, 0, 2, 2] (where 4 is JUMP_DET and 2 is SATURATED)

resulted in two segments, [0, 2] and [3, 7].  Because groups flagged as SATURATED have SCI data set to NaN, the last two groups are erroneously included in the computation of the length of each group.  Instead of computing two segments of length 3 and 5, the segment lengths were computed as 5 and 7.

The computation of the segment length is now changed to consider only the mask, which, as mentioned, is a boolean array with contiguous `True` values denoting the groups of a segment.

This PR will affect regression tests, since there are pixels with saturated groups in ramps.  JWST regression tests have been run locally using the `-k rate` filter and the tests showing differences in outputted data and the truth table show differences in pixels that have saturated ramps, as expected.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,13 @@
 Bug Fixes
 ---------
 
--
+ramp_fitting
+~~~~~~~~~~~~
+
+- Correctly compute the number of groups in a segment to properly compute the
+  optimal weights for the OLS ramp fitting algorithm.  Originally, this
+  computation had the potential to include groups not in the segment being
+  computed. [#163]
 
 Changes to API
 --------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,16 +4,6 @@
 Bug Fixes
 ---------
 
-Changes to API
---------------
-
-- drop support for Python 3.8 [#162]
-
-Other
------
-
--
-
 ramp_fitting
 ~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
Looks OK to me, other than the question about the code style error for `c_mask_2d == True`. If it really needs to work that way, we can always just override/ignore the CI code style error.

kmacdonald-stsci:
> Looks OK to me, other than the question about the code style error for `c_mask_2d == True`. If it really needs to work that way, we can always just override/ignore the CI code style error.

It doesn't need to work that way.  I made the change you suggested and it worked.

hbushouse:
Looks like the one and only CI failure is due to a ramp_fit test over on the jwst side that simply needs updating due to this change in computation.

PaulHuwe (approved):
This did nto trip up anything in the unit or regression tests of Roman.

PR #169, opened 2023-05-22
Issue-168: Updating Usage of 'np.where' Function

### LINKED ISSUE #168: Optimization: use binary masks instead of indices during ramp fitting
While looking at the ramp fitting code I noticed several instances where a binary mask was created, then`numpy.where` used to generate indices that were then used to set values in an array like in the following example:
https://github.com/spacetelescope/stcal/blob/3321069e833ce61d39a86b88ef4630517b22b286/src/stcal/ramp_fitting/utils.py#L1501-L1505

The binary mask can be used directly:
```python
data_sect[np.bitwise_and(gdq_sect, ramp_data.flags_saturated).astype(bool)] = np.NaN
```
which will result in faster code (the modified code takes ~15% of the original code using a test case, included below, on my machine).

```python
import time
import numpy

bad_flag = 0b1
n_bad = 100
shape = (10, 512, 512)

# generate a fixed number of bad pixels at random
# locations
size = numpy.prod(shape)
flat_arr = numpy.zeros(size, dtype='uint16')
flat_arr[:n_bad] |= bad_flag
numpy.random.shuffle(flat_arr)
dq = flat_arr.reshape(shape)
data = numpy.zeros(shape, dtype='float32')


def f1(data, dq):
    ws = numpy.where(numpy.bitwise_and(dq, bad_flag))
    data[ws] = numpy.NaN
    del ws
    return data


def f2(data, dq):
    data[numpy.bitwise_and(dq, bad_flag).astype(bool)] = numpy.NaN
    return data

def timeit(f):
    t0 = time.perf_counter()
    f()
    t1 = time.perf_counter()
    return t1 - t0

print("f1")
data[:] = 0
t = timeit(lambda: f1(data, dq))
print(f"f1 took {t}")

f1_data = data.copy()

print("f2")
data[:] = 0
t = timeit(lambda: f2(data, dq))
print(f"f2 took {t}")

numpy.testing.assert_array_equal(f1_data, data)
```

kmacdonald-stsci:
Closes #168 

This PR addresses the usage of `np.where` and removes it where possible for performance reasons in the ramp fitting step.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,8 @@ ramp_fitting
 - Correct the "averaging" of the final image slope by properly excluding
   variances as a part of the denominator from integrations with invalid slopes.
   [#167]
+- Removing the usage of ``numpy.where`` where possible for perfomance
+  reasons. [#169]
 
 Changes to API
 --------------
--- a/src/stcal/ramp_fitting/ols_fit.py
+++ b/src/stcal/ramp_fitting/ols_fit.py
@@ -943,10 +943,7 @@ def ramp_fit_slopes(ramp_data, gain_2d, readnoise_2d, save_opt, weighting):
             gain_sect = gain_2d[rlo:rhi, :]
 
             # Reset all saturated groups in the input data array to NaN
-            where_sat = np.where(np.bitwise_and(gdq_sect, ramp_data.flags_saturated))
-
-            data_sect[where_sat] = np.NaN
-            del where_sat
+            data_sect[np.bitwise_and(gdq_sect, ramp_data.flags_saturated).astype(bool)] = np.NaN
 
             # Calculate the slope of each segment
             # note that the name "opt_res", which stands for "optional results",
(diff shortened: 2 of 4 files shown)


braingram (approved):
Excellent!

I took a brief look at other uses last week but never finished updating the branch to a point where I could open a PR:
https://github.com/braingram/stcal/tree/no_where

I see some uses in this PR that I missed in my branch and also some other uses in my branch which are not in this PR.

The changes included in this PR appear to address the bulk of the uses of `np.where`. Thanks!

Was the test added to increase coverage? I'm not familiar enough with ramp fitting to review the test.

Also, a bit off-topic but while looking at the contribution of these lines of code to the runtime of ramp fitting. I noticed that roughly 40% of a test call to `ols_ramp_fit_single` was spent on these two lines of code:
https://github.com/spacetelescope/stcal/blob/main/src/stcal/ramp_fitting/ols_fit.py#L1657-L1658
which appears to be iterating through all pixels. I opened a separate issue #170 for that to discuss if this can be converted to a vector operation or in another way optimized. I wanted to bring it up here in case there is an easy fix.

kmacdonald-stsci:
> https://github.com/spacetelescope/stcal/blob/main/src/stcal/ramp_fitting/ols_fit.py#L1657-L1658

The added test was to increase code coverage and add a case that needs testing.

hbushouse:
> > @PaulHuwe can you (or your designee) run a Roman regtest against this PR branch?
> 
> I think I set this up correctly to run the Roman devdeps regtest with the source branch for this PR: https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/295/

There were a number of failures, with only 1 in the ramp_fitting step itself. It appears to be an unrelated error that is common to all the failures. Should we call this a success?

braingram:
> There were a number of failures, with only 1 in the ramp_fitting step itself. It appears to be an unrelated error that is common to all the failures. Should we call this a success?

I would call this a success but @PaulHuwe might know more about the existing failures which all appear to be unit related.

The previous devdeps run (started by timer with no overrides) had 19 failures:
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-devdeps/294/

The run that used the source branch for this PR had 15 failures. All of these appear to match failures from the previous run:
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FRoman-devdeps/detail/Roman-devdeps/295/tests/

hbushouse (approved):

PR #175, opened 2023-06-22
RCAL-511 Implement Uneven ramp fitting

stscieisenhamer:
Resolves [RCAL-511](https://jira.stsci.edu/browse/RCAL-511)

This PR implements the Casertino, et.al., 2021 algorithm for fitting unevenly spaced ramps. Initial implementation based primarily on code from `romanisim`.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


stscieisenhamer:
Developer note: None of the model and reference file handling has yet to be added. Currently planned on handling this in `romancal`. Part of the reason is to get the basic code out for further development of the jump algorithm, which is closely related

stscieisenhamer:
Just realized there was a commit that was neglected. Will be re-applying it shortly.

kmacdonald-stsci (commented):
Many functions are too long and complicated.  

If reasonable, attempts should be made to keep variable names across the code base the same.  For example, `nreads` and variations were often used interchangeably with group and frame time, making the code confusing to follow.  Also, not sure why use the variable name `resultant`, rather than `group`, since they seem to be used for the same thing.

The code appears to properly implement the algorithm from Casertano's paper.

schlafly:
FWIW, some of the nomenclature answers:
- read_time: I think this is frame time in JWST.  It's one read of the array and is ~3.04 s for Roman.  The reads/frames get averaged into resultants/groups.
- resultant vs. group: yes, these are the same things.  No, I don't know why Roman adopted the "resultant" language and JWST the "group" language, but "Roman" is pretty insistent on using resultants as names in place of groups.  I don't feel strongly about global search and replace resultant for group in this codebase, but more broadly we will just have to live with the fact that the two missions went with different words for the same quantities.
- par and var were short for parameters and variances, but these could be given other names.

ddavis-stsci (approved):
Modulo the given comments, LGTM. 
Since I don't see JWST using this for their ramp analysis so the Roman specific nomenclature should not be too confusing.

src/stcal/ramp_fitting/ols_cas21.pyx
```diff
@@ -0,0 +1,153 @@
...
+
+    This implementation fits all ramp segments between bad pixels
+    marked in the dq image with values not equal to zero.  So the
+    number of fit ramps can be larger than the number of pixels.
+    The derived slopes, corresponding variances, and the locations of
+    the ramps in each pixel are given in the returned dictionary.
+
+    Parameters
+    ----------
+    resultants : np.ndarry[nresultants, npixel]
```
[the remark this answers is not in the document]
WilliamJamieson: Roman and JWST are using different terminology for this. Roman terms this "resultant", so its not clear how to name this.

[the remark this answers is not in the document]
kmacdonald-stsci: I understand that, but this is STCAL.  It's supposed to be telescope agnostic.  Developing parallel code bases with different nomenclature for the same thing undermines being agnostic.

stscieisenhamer:
Finally addressing comments. @kmacdonald-stsci : Agree with most everything.

Driver of not doing too much refactoring/naming conventions are primarily that this algorithm is still undergoing algorithmic/science discussion. Kept the implementation close to the original so that discussion would not be confused by refactor/rename issues. Also, the JWST use is still down-the-road so the terminology issue, though definitely needing to be addressed, will be so later on.

Of course, this is all modulo what RCAL would like to see in the current PR.

schlafly:
Hi Jonathan,

This looks good.  Thinking about next steps:
- I would go ahead and get rid of the utility code supporting the ramp fit interpolator that has been removed.  I'll keep it in the simulator if we ever need it.  I think removing it would mean dropping
  - construct_covar
  - construct_ramp_fitting_matrices
  - construct_ki_and_variances
  - ki_and_variance_grid
  - resultants_to_differences
  - associated tests
- Let's either pull in simulate_many_ramps for test_simulated_ramps or drop that test.  I think I'm in favor of pulling it in; that is a pretty good test.
- I'm happy to sign on to s/resultants/r/groups to improve terminology consistency, etc..

stscieisenhamer:
Some external feedback on the terminology issue: At the 20230720 TIPS, the terminology issue was brought up, of which Casertano defended the current uneven/Roman centric form. Unless there is strong reason to change in this PR, I propose to deal with it in a later PR.

WilliamJamieson:
> Some external feedback on the terminology issue: At the 20230720 TIPS, the terminology issue was brought up, of which Casertano defended the current uneven/Roman centric form. Unless there is strong reason to change in this PR, I propose to deal with it in a later PR.

IMOP I think terminology should follow whatever the reference paper uses if possible as we may have to refer back to the paper in the future. It is unfortunate that the two telescopes could not agree on terminology.

stscieisenhamer:
@schlafly The `simulate_many_ramps` has further `romanisim` and `galsim` dependencies. The current solution allows full testing if the relevant packages are present. If there is strong desire to have this test execute on CI, `romanisim` can be a test dependency. Otherwise, `galsim.GaussianDeviate(seed)` will need to be dealt with, plus copy of at least two other functions.

schlafly:
Oh no, you're right, I don't want to bring in l1.apportion_counts_to_resultants.  That's more detailed than we need for this.  Let me rig up a simulate_many_ramps replacement that would be appropriate here; it's not so bad.

src/stcal/ramp_fitting/ols_cas22_util.py
```diff
@@ -0,0 +1,140 @@
+"""Utility routines for Mutli-Accum Ramp Fitting
+"""
+import numpy as np
+
+# Read Time in seconds
+#   For Roman, the read time of the detectors is a fixed value and is currently
+#   backed into code. Will need to refactor to consider the more general case.
+#   Used to deconstruct the MultiAccum tables into integration times.
+READ_TIME = 3.04
```
schlafly, line 9: Oops, this is Roman specific.  We need to pull this out as an argument.

schlafly:
With the caveat that I haven't tested anything, so I'm sure it's full of bugs, here's a simpler version of simulate_many_ramps...
```
def simulate_many_ramps(ntrial=100, flux=100, readnoise=5, ma_table=None):
    if ma_table is None:
        ma_table = [[1, 4], [5, 1], [6, 3], [9, 10], [19, 3], [22, 15]]
    nread = np.array([x[1] for x in ma_table])
    tij = ma_table_to_tij(ma_table)
    resultants = np.zeros((len(ma_table), ntrial), dtype='f4')
    buf = np.zeros(ntrial, dtype='i4')
    for i, ti in enumerate(tij):
        for t0 in ti:
            buf += np.random.poisson(READ_TIME * flux, ntrial)
        resultants[i] = (buf / len(ti)).astype('f4')
    resultants += np.random.randn(len(ma_table), ntrial) * (
        readnoise / np.sqrt(nread)).reshape(len(ma_table), 1)
    return (ma_table, flux, readnoise, resultants)
```
This is a good simulation of a ramp in the context of the assumptions inherent in the algorithm.  That's better than trying to bring over stuff from l1.apportion_counts_to_resultants, sorry for the bad pointer.

src/stcal/ramp_fitting/ols_cas22_util.py
stscieisenhamer: Seems there was some recent discussion of this value, potentially coming from PPS. I am not seeing anything like this in the schema. What is the process to retrieve from PPS and add as a meta value to the Level 1 file?

stscieisenhamer: also note: this is already an argument. the above is the default. However, I will move this into romancal and make the argument required.

schlafly: This is the value to use: https://github.com/spacetelescope/rad/blob/main/src/rad/resources/schemas/exposure-1.0.0.yaml#L268
I don't know if we're populating it yet usually.

schlafly: Sorry I missed that you were just using FRAME_TIME as a default.  I agree with you that it makes more sense to have that on the romancal side.

stscieisenhamer:
Due to illness, I haven't been able to address the following:
- replace the romanisism-dependent test

If there is need to get this merged sooner rather than later, I would suggest merging as-is and handle the dependency in a separate PR.

schlafly:
Okay, let's make a ticket and move forward.  You're able to test locally that those tests pass?  Sorry about your illness!

stscieisenhamer:
Got the romanisim test dependency completed. However, I want to point out a C-ism that was causing issues. In
ols_cas22.pyx:85, which is
```
fabs((tbar[start + i] - tbarmid) / tscale) ** weight_power)
```
generates an error concerning not being able to coerce a complex number to double. This error appeared after taking in the memory enhancement changes Eddy made in the romanisim version. The condition when this occurs seems to be when `tbar[start + i] == tbarmid`. I can only guess that the floating math is creating a `-0.0` situation. Note that this is not reproducible with pure Python.

The error does suggest setting of a cython parameter, `cython.cpow(True)`, to mitigate the situation. Doing so, as best as I can tell, does remediate the exception, and  produces the expected results.

schlafly:
Oops, I hit this one too.  This happened when cython 3 came out a week or two ago.  I tried pinging you here but should have followed up.  https://github.com/spacetelescope/romanisim/pull/67#issuecomment-1640430712

PR #177, opened 2023-06-29
Implement stcal.alignment sub-package.

mairanteodoro:
This PR addresses the implementation of the `alignment` sub-package for `stcal`.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,7 +86,7 @@ ipython_config.py
 # pyenv
 #   For a library or package, you might want to ignore these files since the code is
 #   intended to run in multiple environments; otherwise, check them in:
-# .python-version
+.python-version
 
 # pipenv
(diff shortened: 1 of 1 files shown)


nden:
@mairanteodoro, @s-goldman Do you mind if this work is kept on a separate branch until ready? Then we can merge into main. In this way it won't interfere with releases. 

Also I think we should add modules only when we need them, to avoid adding too much clutter. So perhaps keep only `utils.py` and `resample_utils.py`.

mairanteodoro:
> 

I don't mind keeping a separate branch for the new sub-package work.

mairanteodoro:
Moving changes to a separate branch (`stcal-alignment`).

PR #190, opened 2023-08-17
RCAL-632: add unit tests for additional methods.

mairanteodoro:
Resolves [RCAL-632](https://jira.stsci.edu/browse/RCAL-632)

This PR addresses adding new methods required by the resampling step for RomanCal and its corresponding unit tests.

**Notes**
- For JWST, the `update_s_region_imaging` is calling `wcs.footprint` with `center=True`, which means the footprint coordinates will be those of the center of the pixels. However, since we're using a bounding box to calculate the footprint, it would make sense to use the *vertices* of the pixels instead of the center, which is the default behavior of the `wcs.footprint` method. Therefore, we added the option to pass the `center` parameter to `update_s_region_imaging` in order to make the unit tests pass the correct way of using `wcs.footprint`.

- This PR is blocking the [PR for the resample step](https://github.com/spacetelescope/romancal/pull/787).

**Checklist**
- [X] updated relevant tests
- [X] added relevant label(s)


tests/test_alignment.py
```diff
@@ -169,10 +181,15 @@ def test_wcs_from_footprints():
 
     wcs = wcs_from_footprints([dm_1, dm_2])
 
-    # check that center of calculated WCS matches the
-    # expected position onto wcs_1 and wcs_2
-    assert all(np.isclose(wcs(2, 2), wcs_1(1, 1)))
-    assert all(np.isclose(wcs(2, 2), wcs_2(2, 2)))
+    # check that all elements of footprint match the *vertices* of the new combined WCS
+    assert all(np.isclose(wcs.footprint()[0], wcs(0, 0)))
```
kmacdonald-stsci, line 185: Just a suggestion: there is built in `np.testing.assert_allclose` that does the same thing as this function composition.

kmacdonald-stsci (approved):

tests/test_alignment.py
mairanteodoro: Thanks for the tip, @kmacdonald-stsci! I'll adopt that method moving forward.

PR #199, opened 2023-08-25
Add c99 flag for gcc

WilliamJamieson:
The Jenkins build for JWST is now getting the error:
```
error: ‘for’ loop initial declarations are only allowed in C99 mode
```
when building `stcal`. It looks like someone turned of that flag for gcc there. This should add the flag in.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


WilliamJamieson:
The `jwst` CI failure looks unrelated to me

hbushouse (approved):

PR #205, opened 2023-08-28
set `build_platform_wheels` to `true`

### LINKED ISSUE #200: Make sure to build system specific wheels for next release
`stcal` now has `cython` code that needs to be built which will come out in the next release. We should be prepared to build and upload all of the different wheels to PyPi

zacharyburnett:
Closes #200 

This PR attempts to build wheels using `cibuildwheel` for specific platforms

If we would like to use the [OpenAstronomy action](https://github-actions-workflows.openastronomy.org/en/stable/publish.html), we'll need the PyPI token (instead of the username and password)

**Checklist**
- [ ] ~added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)~
- [ ] ~updated relevant tests~
- [ ] ~updated relevant documentation~
- [x] added relevant label(s)

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -9,7 +9,7 @@ jobs:
     uses: spacetelescope/action-publish_to_pypi/.github/workflows/workflow.yml@master
     with:
       test: false
-      build_platform_wheels: false # Set to true if your package contains a C extension
+      build_platform_wheels: true # Set to true if your package contains a C extension
     secrets:
       user: ${{ secrets.PYPI_USERNAME_STSCI_MAINTAINER }}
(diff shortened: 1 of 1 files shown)


hbushouse (approved):

WilliamJamieson:
> If we would like to use the [OpenAstronomy action](https://github-actions-workflows.openastronomy.org/en/stable/publish.html), we'll need the PyPI token (instead of the username and password)
> 
For security reasons the `STSCI_MAINTAINER` PyPi account really ought to be 2FA (if it is not that should be considered). According to https://blog.pypi.org/posts/2023-06-01-2fa-enforcement-for-upload/, API tokens or Trusted Publish configurations must be in place instead.

@zanecodes do you know what the status is of the PyPi account used for this sort of automation is? I thought the "password" was really a generated token in this case.

kmacdonald-stsci (commented):
Looks fine

PR #207, opened 2023-08-30
Snowball flagging of grouped data

mwregan2:
This PR addresses the case of snowballs occurring within a group. When this happens the center of the snowball does not saturate until the next group. This causes the snowball to not be recognized.
The change is to look at the next group to see if it is saturated.

**Checklist**
- [x] updated relevant tests
- [x] added relevant label(s)


src/stcal/jump/jump.py
```diff
@@ -609,25 +607,38 @@ def make_snowballs(gdq, integration, group, jump_ellipses, sat_ellipses,
...
     # of the saturation circle within the enclosing jump rectangle.
     snowballs = []
+    num_groups = gdq.shape[1]
     for jump in jump_ellipses:
-        if near_edge(jump, low_threshold, high_threshold):
+        # center of jump should be saturated
+        jump_center = jump[0]
+        # if center of the jump ellipse is not saturated in this group and is saturated in
+        # the next group add the jump ellipse to the snowball lisr
+        if group < (num_groups - 1) and \
```
kmacdonald-stsci, line 616: Use parentheses, instead of '\' to capture multi-line conditional.

src/stcal/jump/jump.py
kmacdonald-stsci, line 1: A lot of functions in this file are missing docstrings.

src/stcal/jump/jump.py
```diff
@@ -609,25 +607,38 @@ def make_snowballs(gdq, integration, group, jump_ellipses, sat_ellipses,
...
-                            gdq[integration, :, :, :] = \
-                                extend_saturation(gdq[integration, :, :, :],
-                                                  group, [sat], sat_flag,
-                                                  min_sat_radius,
-                                                  expansion=expansion,
-                                                  max_extended_radius=max_extended_radius)
+    # extend the saturated ellipses that are larger than the min_sat_radius
+    gdq[integration, :, :, :] = \
+        extend_saturation(gdq[integration, :, :, :],
+        group, sat_ellipses, sat_flag,
```
kmacdonald-stsci, line 637: Indent function parameters to align with first function parameter.

kmacdonald-stsci (approved):
The `jump.py` file has many functions missing docstring.  This should be corrected, but is not necessary for this PR.

src/stcal/jump/jump.py
mwregan2: If every function has to have full doc strings, a large fraction of the lines in the file will be redundant listings of parameters being passed down to lower level functions.

src/stcal/jump/jump.py
mwregan2: fixed

hbushouse (commented):
Code looks OK to me. Will give final approval once we've done a regtest run.

hbushouse:
regtest started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/920/

hbushouse:
OK, I guess none of our test datasets in the regression test suite contain any snowballs that fit this description, because I see no changes at all from this PR. But at least it didn't crash or produce unwanted changes, so I guess it's good.

hbushouse (approved):

PR #214, opened 2023-09-19
use PyPI upload workflow from OpenAstronomy

zacharyburnett:
Resolves [SCSB-108](https://jira.stsci.edu/browse/SCSB-108)

This PR changes the PyPI upload process to use the OpenAstronomy workflow instead of the STScI one.

There is an analogous PR in `stpipe` using the "pure Python" workflow; however, as `stcal` now includes a C extension, we need to use the workflow that uses `cibuildwheel`. The workflow in this PR is configured to test the build with each pull request push, and then publish upon an official release.

~Additionally, we can't use this workflow until we make a PyPI token (with write / upload permission to packages) for the STScI maintainer account. Currently, we use a username and password to PyPI; however, this method will be deprecated at some point in the future in favor of tokens.~

**Checklist**
- [x] ~added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)~
- [x] ~updated relevant tests~
- [x] ~updated relevant documentation~
- [x] added relevant label(s)


zacharyburnett:
encountered this error when building:
https://github.com/spacetelescope/stcal/actions/runs/6262073971/job/17003453952#step:7:580
```
../../numpy/meson.build:207:4: ERROR: Problem encountered: No BLAS library detected! Install one, or use the `allow-noblas` build option (note, this may be up to 100x slower for some linear algebra operations).
```

kmacdonald-stsci (approved):

WilliamJamieson:
The wheel build failure is because of it trying to build PyPy wheels. Unless we want to support PyPy wheels (which numpy and astropy do not), cibuildwheel will need to be configured to not build PyPy wheels.

PR #215, opened 2023-09-27
Add Roman jump detection to ramp fitting

WilliamJamieson:
This PR adds the Roman jump detection to the roman ramp fitting algorithm.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


WilliamJamieson:
This is very much in draft form right now. The conversion from using MA tables to read_patterns needs to be completed within the supporting python code.

There also seems to be a bug in the code somewhere because its not detecting jumps in the right places sometimes.

src/stcal/ramp_fitting/ols_cas22/_fixed.pyx
```diff
@@ -0,0 +1,187 @@
+"""
```
WilliamJamieson, line 1: I need to remove the use of negative indices here so I can remove some types of checks so that its faster. This is relatively simple.

tests/test_jump_cas22.py
```diff
@@ -0,0 +1,444 @@
...
+            assert set(ramp_indicies).union(fit['jumps']) == set(range(len(read_pattern)))
+
+    # Check that the slopes have been estimated reasonably well
+    #   There are not that many pixels to test this against and many resultants
+    #   have been thrown out due to the jumps. Thus we only check the slope is
+    #   "fairly close" to the expected value. This is purposely a loose check
+    #   because the main purpose of this test is to verify that the jumps are
+    #   being detected correctly, above.
+    for fit in fits:
+        assert_allclose(fit['average']['slope'], FLUX, rtol=3)
```
schlafly, line 444: If you were to write this test like the chi^2 test, what tolerance would you need?  It is very reassuring when the chi^2/dof comes out close to 1.

tests/test_ramp_fitting_cas22.py
```diff
@@ -14,22 +14,22 @@
...
-def test_readpattern_to_matable():
+def test_read_pattern_to_ma_table():
     """Test conversion from read pattern to multi-accum table"""
     pattern = [[1], [2, 3], [4], [5, 6, 7, 8], [9, 10], [11]]
     expected = [[1, 1], [2, 2], [4, 1], [5, 4], [9,2], [11,1]]
 
-    result = ols_cas22_util.readpattern_to_matable(pattern)
+    result = ols_cas22_util.read_pattern_to_ma_table(pattern)
 
     assert result == expected
```
schlafly, line 34: Let's delete these.

tests/test_ramp_fitting_cas22.py
```diff
@@ -38,14 +38,18 @@ def test_simulated_ramps():
...
-        resultants, resultants * 0, read_noise, ROMAN_READ_TIME, ma_table=ma_table)
+        resultants, dq, read_noise, ROMAN_READ_TIME, ma_table=ma_table)
+
     chi2dof_slope = np.sum((par[:, 1] - flux)**2 / var[:, 2]) / ntrial
     assert np.abs(chi2dof_slope - 1) < 0.03
 
     # now let's mark a bunch of the ramps as compromised.
     bad = np.random.uniform(size=resultants.shape) > 0.7
-    dq = resultants * 0 + bad
+    dq += bad
```
schlafly, line 52: For the little it's worth, I expect |= rather than += for these kinds of bitwise operations.

schlafly (commented):
This looks good to me.  I made a bunch of minor comments.  Some summary:
* Let's get rid of all of the _ma_table routines.  We only ever need read_pattern; the ma_table pattern was something I made up for the simulator before read_pattern existed, and it doesn't need to continue existing.
* It will be good to see that this matches Sanjib's results, but the current tests look good and I'm optimistic.
* The final test in test_find_jumps would be more convincing if it were phrased in terms of the chi^2/dof.    That's slightly annoying in that case since the degrees of freedom are 8 resultants - ~2 flagged - (2, 3, or 4 depending on where the jump landed).  But I think it's not that bad and would make for a strong demo that both the ramps and their variances were good.

src/stcal/ramp_fitting/ols_cas22/_fixed.pyx
```diff
@@ -0,0 +1,197 @@
...
+    Fast constructor for Fixed class
+"""
+import numpy as np
+cimport numpy as np
+cimport cython
+
+from stcal.ramp_fitting.ols_cas22._core cimport Thresh, DerivedData, Diff
+from stcal.ramp_fitting.ols_cas22._fixed cimport Fixed
+
+cdef class Fixed:
```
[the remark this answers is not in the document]
WilliamJamieson: These are not constants. Calling them "constants" implies that they never change between entirely different runs. They are "fixed" relative to the set of pixels being fit to the same read pattern.

src/stcal/ramp_fitting/ols_cas22/_pixel.pxd
```diff
@@ -0,0 +1,23 @@
...
+from stcal.ramp_fitting.ols_cas22._core cimport RampFit, RampFits, RampIndex
+from stcal.ramp_fitting.ols_cas22._fixed cimport Fixed
+
+cdef class Pixel:
+    cdef Fixed fixed
+    cdef float read_noise
+    cdef float [:] resultants
+
+    cdef float[:, :] delta
+    cdef float[:, :] sigma
```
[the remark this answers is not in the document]
WilliamJamieson: Header files cannot contain comments.

src/stcal/ramp_fitting/ols_cas22/_fixed.pyx
```diff
@@ -0,0 +1,197 @@
...
+    recip : float[:, :]
+        single sum of reciprocal n_reads:
+            recip[0, :] = ((1/n_reads[i+1]) + (1/n_reads[i]))
+        double sum of reciprocal n_reads:
+            recip[1, :] = ((1/n_reads[i+2]) + (1/n_reads[i]))
+    slope_var : float[:, :]
+        single of slope variance term:
+            slope_var[0, :] = ([tau[i] + tau[i+1] - min(t_bar[i], t_bar[i+1]))
+        double of slope variance term:
+            slope_var[1, :] = ([tau[i] + tau[i+2] - min(t_bar[i], t_bar[i+2]))
```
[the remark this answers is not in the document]
schlafly: Yes, I like both slope_var_val, var_slope_coeff, slope_var_coeff, etc. more than slope_var.

tests/test_jump_cas22.py
```diff
@@ -0,0 +1,444 @@
...
+
+from stcal.ramp_fitting.ols_cas22._wrappers import read_data
+from stcal.ramp_fitting.ols_cas22._wrappers import init_ramps
+from stcal.ramp_fitting.ols_cas22._wrappers import run_threshold, make_fixed, make_pixel
+
+from stcal.ramp_fitting.ols_cas22 import fit_ramps, Parameter, Variance, Diff
+
+
+RNG = np.random.default_rng(619)
+ROMAN_READ_TIME = 3.04
```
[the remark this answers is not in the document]
WilliamJamieson: This can be any (reasonable) float. I just chose that time as that was the time it needs to work with. I can change it if you want?

tests/test_jump_cas22.py
WilliamJamieson: See last test, it has been modified to use enough points to get good statistical convergence.

tests/test_jump_cas22.py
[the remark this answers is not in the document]
schlafly: Really, I just feel like we'd like this to be mission independent, so I think I'm just asking to delete ROMAN_.  But obviously this is minor and I don't feel strongly.

src/stcal/ramp_fitting/ols_cas22/_fixed.pyx
```diff
@@ -0,0 +1,213 @@
...
+            double sum of reciprocal n_reads:
+                recip[Diff.double, :] = ((1/n_reads[i+2]) + (1/n_reads[i]))
+    var_slope_coeffs : float[:, :]
+        Coefficients for the slope portion of the variance used to compute the
+        jump detection statistics, which happend to be fixed for any given ramp
+        fit.
+            single of slope variance term:
+                slope_var[Diff.single, :] = ([tau[i] + tau[i+1] - min(t_bar[i], t_bar[i+1]))
+            double of slope variance term:
+                slope_var[Diff.double, :] = ([tau[i] + tau[i+2] - min(t_bar[i], t_bar[i+2]))
```
schlafly, line 68: I think you still need to change the slope_var in these comments to var_slope_coeffs.

src/stcal/ramp_fitting/ols_cas22/_fit_ramps.pyx
```diff
@@ -0,0 +1,138 @@
...
+    # Use list because this might grow very large which would require constant
+    #    reallocation. We don't need random access, and this gets cast to a python
+    #    list in the end.
+    cdef cpp_list[RampFits] ramp_fits
+
+    cdef np.ndarray[float, ndim=2] parameters = np.zeros((n_pixels, 2), dtype=np.float32)
+    cdef np.ndarray[float, ndim=2] variances = np.zeros((n_pixels, 3), dtype=np.float32)
+
+    # Copy the dq array so we can modify it without fear
+    cdef np.ndarray[int, ndim=2] fit_dq = dq.copy()
```
schlafly, line 117: Would it make sense to modify dq in place?  I have been trying to avoid making copies of any of the things that are n_pixel * n_resultant in size.

tests/test_jump_cas22.py
```diff
@@ -0,0 +1,539 @@
...
+    if not use_dq:
+        assert okay.all()
+
+    output = fit_ramps(resultants, dq, read_noise, ROMAN_READ_TIME, read_pattern, use_jump=use_jump)
+    assert len(output.fits) == N_PIXELS  # sanity check that a fit is output for each pixel
+
+    chi2 = 0
+    for fit, use in zip(output.fits, okay):
+        if not use_dq:
+            assert len(fit['fits']) == 1  # only one fit per pixel since no dq/jump in this case
```
schlafly, line 359: I'm trying to make sure I understand this case.  I would have said if use_jump=True and use_dq=False, there should be some falsely detected jumps from noise among the 100k pixels which would give a few cases where len(fit['fits']) > 1.  What am I missing?

schlafly (approved):
This looks great.  Two questions:
- Can we modify dq in place?  I have been trying to avoid making any new things that are nresultants x npixel in size elsewhere in the pipeline.
- I am missing some bit in one of the tests where I expect some false positives.  Are there really none or am I confused?

tests/test_jump_cas22.py
WilliamJamieson: In absolute terms read noise is so low compared with the expected signal that one would not expect there to be an "significant" deviations even over such a large data set. That would require the Poisson process to "significantly" deviate enough over a resultant to impact the average combined with the read noise process aligning aligning "significantly" in the same direction. This combination of events is extremely unlikely given that the read noise is quite small.

Basically the `FLUX` and `READ_NOISE` test constants were chosen purposefully to avoid extraneous jump detections so that we would only detect jumps when we specifically insert them. The test-case you highlighted here is a demonstration that the data used for testing does not trigger false positives when we don't purposefully insert jumps. Note that because `detector_data` is a fixture scoped for the entire module it only gets computed once for the purposes of the tests in question. Moreover, the `jump_data` fixture depends on the `detector_data` fixture and is also scoped for the model making it also constant with respect to the tests in question. Effectively this means that no false positives for jump detection means that the jump detection tests do not have to be concerned with dealing with false positives.

schlafly: Thanks.  We've had this conversation in a side channel, but agreeing here as well---the threshold is large enough that we only expect a false positive one in every few million pixels or so, so it's not surprising that this test with 100k ramps passes.  Sounds good, thanks.

schlafly (approved):
LGTM, thanks.

src/stcal/ramp_fitting/ols_cas22/_wrappers.pyx
kmacdonald-stsci, line 1: No function has any docstring.

kmacdonald-stsci (approved):

src/stcal/ramp_fitting/ols_cas22/_core.pyx
```diff
@@ -0,0 +1,238 @@
...
+    s: float
+        signal from the resultants
+
+    Returns
+    -------
+    signal power from Table 2
+    """
+    cdef int i
+    for i in range(6):
+        if s < PTABLE[0][i]:
```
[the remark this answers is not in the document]
WilliamJamieson: Done

src/stcal/ramp_fitting/ols_cas22/_wrappers.pyx
WilliamJamieson: I merged the wrappers back into the main code using some `cpdef` and made some notes about them.

stscieisenhamer (approved):
not knowing the details of jump detection, lgtm

PR #226, opened 2023-11-01
RCAL-702: Fix memory use issue for Cas22 jump detection

WilliamJamieson:
Resolves [RCAL-702](https://jira.stsci.edu/browse/RCAL-702)

This PR resolves a massive memory spike as part of the solution to RCAL-702

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ ramp_fitting
 
 - Refactor Casertano, et.al, 2022 uneven ramp fitting and incorporate the matching
   jump detection algorithm into it. [#215]
+- Fix memory issue with uneven ramp fitting [#226]
 
 Changes to API
 --------------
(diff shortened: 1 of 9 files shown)


schlafly (approved):
Looks good to me, thanks!

schlafly:
@stscieisenhamer , mind checking to see if this resolves your issue?  Thanks!

kmacdonald-stsci (approved):

stscieisenhamer (approved):
Dramatic change in memory usage: From 150GB to not even breaking 1GB.

And also, there is an actual run-time difference in running with and without `--use_jump_detection`. Without the step runs ~1.5 minutes. With jump detection, the task runs about as long as before this fix, ~6 minutes. As a reminder, before the runtime was no different between the two.

LGTM

schlafly:
I think I see that with include_diagnostic=False we don't get any information back about where the jumps are and we don't end up updating the DQ array.  I think we need that.  Ideally that could happen in place on the DQ array, so no additional memory were needed; I'm not sure how convenient that ends up being here.

PR #227, opened 2023-11-06
Resolve some bugs with the jump fitting algorithm

WilliamJamieson:
This PR makes some adjustments to the jump detection algorithm so that it exactly detects the jumps detected in the simulated data.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/tests/test_jump_cas22.py
+++ b/tests/test_jump_cas22.py
@@ -425,7 +425,6 @@ def test_fit_ramps(detector_data, use_jump, use_dq):
     assert len(output.fits) == N_PIXELS  # sanity check that a fit is output for each pixel
 
     chi2 = 0
-    excess = 0  # counts the number of jumps erroneously detected
     for fit, use in zip(output.fits, okay):
         if not use_dq and not use_jump:
             ##### The not use_jump makes this NOT test for false positives #####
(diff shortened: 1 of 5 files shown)


tests/test_jump_cas22.py
```diff
@@ -24,7 +24,7 @@
 READ_NOISE = np.float32(5)
 
 # Set a value for jumps which makes them obvious relative to the normal flux
-JUMP_VALUE = 10_000
+JUMP_VALUE = 1_000
```
schlafly, line 27: Looks good to me.  I wasn't sure if this one was intentional.

WilliamJamieson: This had no visible effect on the results

tests/test_jump_cas22.py
```diff
@@ -447,7 +452,7 @@ def test_fit_ramps(detector_data, use_jump, use_dq):
 
             assert use_dq # sanity check that this branch is only encountered when use_dq = True
 
-    chi2 /= np.sum(okay)
+    chi2 /= (np.sum(okay) + excess)
```
schlafly, line 455: I don't think it makes a difference, but I don't think I expected the "excess" to make an appearence here, since those cases don't contribute to chi^2 above on line 444.

schlafly (approved):

tests/test_jump_cas22.py
WilliamJamieson: It actually should be be a difference not a sum, but you are right, it makes no difference

schlafly: Yup, would make sense subtracted, thanks.

kmacdonald-stsci (approved):

PR #228, opened 2023-11-07
JP-3458: Moving ramp_fit CI tests from JWST to STCAL.

kmacdonald-stsci:
Resolves [JP-3458](https://jira.stsci.edu/browse/JP-3458)

This PR addresses moves CI tests from JWST to STCAL.

This PR is in conjunction with https://github.com/spacetelescope/jwst/pull/8060, which removed these tests from JWST.

**Checklist**
- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,9 +10,10 @@ ramp_fitting
   jump detection algorithm into it. [#215]
 
 - Fix memory issue with uneven ramp fitting [#226]
+
 - Fix some bugs in the jump detection algorithm [#227]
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):

WilliamJamieson (approved):
These look fine to me. I would suggest adding the commit hash from jwst, but that probably is not necessary.

WilliamJamieson (approved):
These look fine to me. I would suggest adding the commit hash from jwst, but that probably is not necessary.

PR #229, opened 2023-11-09
Speed up Jump detection

WilliamJamieson:
This PR dramatically speeds up Roman's jump detection (and ramp fitting to a lesser extent). Empirical testing locally showed me a speed up from ~33sec to ~6sec for fitting a test data set using jump detection.

The main culprit for this problem was the fact that to use `numpy` one has to jump out of the C code temporally. This overhead was orders of magnitude larger (relative to the expected data sizes, number of resultants per pixel in this case) than just running the computation directly in C using cython commands despite the fact that the numpy implementations of those numerical operations are much more highly optimized. Steps have also been taken to remove an unnecessary calls out of C to python which also has lead to a modest speed improvement.

~Note that there maybe additional speed improvements by explicitly aligning and indicating the contiguous array dimensions in the memory views.~
> I performed this experiment but found essentially no improvements. In fact it might be slower because our expected input data is not going to be contiguous in the dimension that would result in speed improvements. In order to make this contiguous we would have to copy the input data which is quite expensive.

**Checklist**
- [ ] updated relevant tests
- [ ] added relevant label(s)


schlafly (approved):
Looks good to me.  A major refactor!

PR #232, opened 2023-11-17
Convert ramp fitting Cython .pyx files into annotated .py files

WilliamJamieson:
Cython 3+ supports compiling directly from annotated .py files instead of its own .pyx files. Indeed, Cython [recommends](https://cython.readthedocs.io/en/latest/src/userguide/faq.html#do-i-need-to-rename-my-py-file-to-pyx) doing this if there is no need to interface directly into C files, which is exactly the case for ramp fitting. This is because it enables the full suite of standard python tooling without the need to seek out tooling which specifically supports Cython. Moreover, it also reduces human context switching when reading the code as the "Cython" code now reads exactly like (verbose) python code.

In addition to this switch, I made several other small tweaks to ramp fitting which operationally make no difference (the code exactly reproduces the previous version's results). These tweaks net a modest ~20%-30% speed up over the current ramp fitting code. Other changes include some refactoring to make parts of the code more readable.

Note that when consulting the Cython docs, I discovered that code tends to be faster if all of it is contained within a single Cython module as the Cython C-transpiler can perform some optimizations such as inlining (i.e. Cython cannot inline a function, even if it is marked to be inlined, if the function is called from a different module). This makes the code slightly less readable but it did result in a small but noticeable performance gain.

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)


src/stcal/ramp_fitting/_ols_cas22/_fit.pxd
```diff
@@ -0,0 +1,32 @@
+# cython: language_level=3str
```
WilliamJamieson, line 1: Unfortunately, I was unable to eliminate this file because Cython's `enum` structures do not have a Python syntax equivalent at this time.

schlafly:
I agree it's nice to keep this "pure python" from a maintainability perspective.  This _exactly_ reproduces the previous results, so no changes to regression tests, etc., are needed?

schlafly:
This code looks good to me and it's nice to be more pure-python-y.  I think I understood once that this gives you literally the exact same results as before (i.e., no regression tests would need to change).  Is that the case?

kmacdonald-stsci (approved):

stscieisenhamer:
Is this PR still relevant?

schlafly:
Ah, we lost track of this one.  I'm just going to close.

PR #238, opened 2024-01-05
JP-3502: Extend snowball flags to next integration

mwregan2:
Resolves [JP-3502](https://jira.stsci.edu/browse/JP-3502)

Closes spacetelescope/jwst#8174

This PR addresses the problem that the saturated core of snowballs shows up in the next integration with an excess count rate that takes a long time to decay. The solution is to mark as jump a number of groups passed in from JWST Jump. 

**Checklist**

- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,7 @@
 
 - 
 
-1.5.1 (2023-10-16)
+1.5.1 (2023-11-16)
 ==================
 
 - re-release to publish source distribution
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,41 @@
 1.5.3 (unreleased)
 ==================
 
+Changes to API
+--------------
+
 jump
-----
+~~~~
+
 - Add in the flagging of groups in the integration after a snowball
   occurs. The saturated core of the snowball gets flagged as jump
   for a number of groups passed in as a parameter [#238]
+
+Bug Fixes
+---------
+
+jump
+~~~~
+
+- Fixed the computation of the number of rows per slice for multiprocessing, which
+  was causing different results when running the step with multiprocess [#239]
+
+- Fix the code to at least always flag the group with the shower and the requested
+  groups after the primary shower. [#237]
+
+Other
+-----
+
+jump
+~~~~
+
+- Reorganize jump docs between the jwst and stcal repos. [#240]
+
+ramp_fitting
+~~~~~~~~~~~~
+
+- Reorganize ramp_fitting docs between the jwst and stcal repos. [#240]
+
 
 1.5.2 (2023-12-13)
 ==================
@@ -28,7 +58,7 @@ Other
 - Enable automatic linting and code style checks [#187]
 
 ramp_fitting
-------------
+~~~~~~~~~~~~
 
 - Refactor Casertano, et.al, 2022 uneven ramp fitting and incorporate the matching
   jump detection algorithm into it. [#215]
@@ -88,13 +118,16 @@ jump
   within a group. [#207]
 
 - Added more allowable selections for the number of cores to use for
-  multiprocessing [#183].
+  multiprocessing [#183]
+
+- Fixed the computation of the number of rows per slice for multiprocessing,
+  which caused different results when running the step with multiprocess [#239]
 
 ramp_fitting
 ~~~~~~~~~~~~
 
 - Added more allowable selections for the number of cores to use for
-  multiprocessing [#183].
+  multiprocessing [#183]
 
 - Updating variance computation for invalid integrations, as well as
   updating the median rate computation by excluding groups marked as
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@
 ==================
 
 jump
-------------
+----
 - Add in the flagging of groups in the integration after a snowball
   occurs. The saturated core of the snowball gets flagged as jump
   for a number of groups passed in as a parameter [#238]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,12 @@
-1.5.2 (unreleased)
+1.5.3 (unreleased)
 ==================
 
 - 
+
+1.5.2 (2023-12-13)
+==================
+
+- non-code updates to testing and development infrastructure
 
 1.5.1 (2023-11-16)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,11 @@
 1.5.3 (unreleased)
 ==================
 
-- 
+jump
+------------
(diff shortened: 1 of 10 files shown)


src/stcal/jump/jump.py
```diff
@@ -54,6 +54,8 @@ def detect_jumps(
     minimum_groups=3,
     minimum_sigclip_groups=100,
     only_use_ints=True,
+    mask_persist_grps_next_int=True,
+    persist_grps_flagged=25,
```
kmacdonald-stsci, line 58: This function now has roughly 30 arguments.  Would it be better served by using a class to contain the parameters, rather than continue to add parameters to this function?

src/stcal/jump/jump.py
```diff
@@ -537,6 +550,7 @@ def flag_large_events(
     total_snowballs = 0
     nints = gdq.shape[0]
```
kmacdonald-stsci, line 551: This could be done as 

`nints, ngroups, nrows, ncols = gdq.shape`

Which would make the next few lines more descriptive in the use of variables.  For example, `gdq.shape[2]` isn't very descriptive.

src/stcal/jump/jump.py
```diff
@@ -610,10 +637,26 @@ def extend_saturation(
...
+                (round(ceny), round(cenx)),
+                (round(ellipse[1][0] / 2), round(ellipse[1][1] / 2)),
+                alpha,
+                0,
+                360,
+                (0, 0, 22),
+                -1,
+            )
+            persist_ellipse = persist_image[:, :, 2]
+            persist_saty, persist_satx = np.where(persist_ellipse == 22)
```
kmacdonald-stsci, line 657: What is special about 22 here?  A comment or declaring a descriptive variable with the value 22 would be good here.

kmacdonald-stsci (approved):

CHANGES.rst
```diff
@@ -1,7 +1,11 @@
 1.5.3 (unreleased)
 ==================
 
-- 
+jump
+------------
```
hbushouse, line 5:
```suggestion
----
```

src/stcal/jump/jump.py
```diff
@@ -525,7 +533,12 @@ def flag_large_events(
         required for a snowball to be created
     max_extended_radius : int
         The largest radius that a snowball or shower can be extended
-
+    mask_persist_grps_next_int: Boolean
```
hbushouse, line 536:
```suggestion
    mask_persist_grps_next_int : bool
```

src/stcal/jump/jump.py
```diff
@@ -525,7 +533,12 @@ def flag_large_events(
...
+        The flag to turn on the extension of the flagging of the saturated cores of
+        snowballs.
+    persist_grps_flagged : int (grps)
```
hbushouse, line 539:
```suggestion
    persist_grps_flagged : int
```

src/stcal/jump/jump.py
mwregan2: I agree that it has a lot of parameters but users need to be able to tweak the parameters. Is there any cost to the large number of parameters?

src/stcal/jump/jump.py
mwregan2: I made the change.

src/stcal/jump/jump.py
mwregan2: I updated the comments in this section. Let me know, if you think it needs more details.

src/stcal/jump/jump.py
kmacdonald-stsci: The cost is readability and maintainability.  Grouping various parameters into a class or classes makes the code easier to read, as well as easier to maintain, especially when adding a parameter that needs to be passed to subroutines.  If you simply pass parameters, the function signature needs to be changed for all functions that need that parameter.  If you use classes, those classes will be passed to the subroutines, so when you add a parameter to a class, no subroutine signature needs to be edited, since the parameter is contained in the class.

kmacdonald-stsci (approved):

hbushouse:
Lots of CI failures that appear to be due to API changes in function calls.

src/stcal/jump/jump.py
mwregan2: No real reason that I can remember. There needs to be a number that I can use np.where to find the pixels that are affected. I'm sure there's a more elegant way to do it.

kmacdonald-stsci: I think it would be fine to make a comment saying this number could change and that this number was determined though trial and error.  But more importantly what this number is supposed to represent.  Numbers with no reference or description leaves future programmers to try to figure out what that number represents and why that number was chosen.

src/stcal/jump/jump.py
mwregan2: While it is true that passing objects instead of primitives  has advantages there are also disadvantages. APIs with only primitive parameters are easier to understand where the primitives are coming from. At the extreme, a function that has a single class as input in a black box. Someone browsing the code has more work to do than with primitive parameters. 
  Also, If we want to consolidate the parameters passed to detect_jump.py, I feel that the change should be in a separate PR. That would give us a working master to compare against. 
  What do you think?

src/stcal/jump/jump.py
mwregan2: The value of 22 is arbitrary. It just needs to be something that can be tested for to create the mask of pixels within the ellipse.

hbushouse (approved):
Given that the only outstanding (unresolved) comments are related to stylistic issues, and not actual problems with the code, I'm going to approve in order to get this moving.

hbushouse:
@mwregan2  Doesn't this need a corresponding change in the jump_step.py module in the jwst package, in order to add the 2 new step arguments that've been implemented here and pass them to the `detect_jumps` function? I don't see an existing PR against jwst that contains such changes.

mwregan2:
I’ve got the branch ready. I do the PR today


From: Howard Bushouse ***@***.***>
Date: Wednesday, February 14, 2024 at 8:16 PM
Subject: Re: [spacetelescope/stcal] JP-3502: Extend snowball flags to next integration (PR #238)


@mwregan2<https://urldefense.com/v3/__https:/github.com/mwregan2__;!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!1NXfDtfYo8HSoZ6v0E7RmpmHHIKvFdjtrsWx0pLRSHeSgrttPx_cnJlh4CvFGoGwNYB0Aav6ilvQIzGqcsMHQ_E$> Doesn't this need a corresponding change in the jump_step.py module in the jwst package, in order to add the 2 new step arguments that've been implemented here and pass them to the detect_jumps function? I don't see an existing PR against jwst that contains such changes.

PR #239, opened 2024-01-25
JP-3521: Fix jump multiprocessing error

mwregan2:
Resolves [JP-3521](https://jira.stsci.edu/browse/JP-3521)

Closes spacetelescope/jwst#8221

This PR addresses the different results return from the Jump step when switching from single core processing to multiprocessing. The main problem was caused by calculating the number of rows per slice using floating point division. This lead to some rows not getting processed. 
In addition, the 4 neighbor flagging in multiprocessing was adding the jump bit to pixels that were already flagged as do_not_use or saturated. 

**Checklist**

- [x] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -86,6 +86,9 @@ jump
 - Added more allowable selections for the number of cores to use for
   multiprocessing [#183].
 
+- Fixed the numer of rows per slice in multiprocessing that lead to
+  different results when running the step with multiprocess [#239]
+
 ramp_fitting
 ~~~~~~~~~~~~
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -86,8 +86,8 @@ jump
 - Added more allowable selections for the number of cores to use for
   multiprocessing [#183].
 
-- Fixed the numer of rows per slice in multiprocessing that lead to
-  different results when running the step with multiprocess [#239]
+- Fixed the computation of the number of rows per slice for multiprocessing,
+  which caused different results when running the step with multiprocess [#239]
 
 ramp_fitting
 ~~~~~~~~~~~~
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -427,6 +427,13 @@ def detect_jumps(
             # save the neighbors to be flagged that will be in the next slice
             previous_row_above_gdq = row_above_gdq.copy()
             k += 1
+            # remove redundant bits in pixels that have jump flagged but were
+            # already flagged as do_not_use or saturated.
+            gdq[gdq == np.bitwise_or(dqflags['DO_NOT_USE'], dqflags['JUMP_DET'])] = \
+                dqflags['DO_NOT_USE']
+            gdq[gdq == np.bitwise_or(dqflags['SATURATED'], dqflags['JUMP_DET'])] = \
+                dqflags['SATURATED']
+
         #  This is the flag that controls the flagging of snowballs.
         if expand_large_events:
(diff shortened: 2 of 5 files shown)


src/stcal/jump/jump.py
```diff
@@ -427,6 +427,13 @@ def detect_jumps(
             # save the neighbors to be flagged that will be in the next slice
             previous_row_above_gdq = row_above_gdq.copy()
             k += 1
+            # remove redundant bits in pixels that have jump flagged but were
+            # already flagged as do_not_use or saturated.
+            gdq[gdq == np.bitwise_or(dqflags['DO_NOT_USE'], dqflags['JUMP_DET'])] = \
```
kmacdonald-stsci, line 432: Could this be simplified in checking only for `JUMP_DET` since if the `DO_NOT_USE` flag is already set, there's no need to check for it?

src/stcal/jump/jump.py
```diff
@@ -427,6 +427,13 @@ def detect_jumps(
...
+                dqflags['DO_NOT_USE']
+            gdq[gdq == np.bitwise_or(dqflags['SATURATED'], dqflags['JUMP_DET'])] = \
```
kmacdonald-stsci, line 434: Could this be simplified in checking only for `JUMP_DET` since if the `SATURATED` flag is already set, there's no need to check for it?

kmacdonald-stsci (approved):

src/stcal/jump/jump.py
mwregan2: The pixels have already had the JUMP_DET added when the slices are combined using the rows above and below the slice. This addition does not check for existing flags. It is cleaner to fix the redundant flagging that to check each pixel before add the jump flag. The redundant flagging happens about 1 in a million times.

src/stcal/jump/jump.py
mwregan2: The pixels have already had the JUMP_DET added when the slices are combined using the rows above and below the slice. This addition does not check for existing flags. It is cleaner to fix the redundant flagging that to check each pixel before adding the jump flag. The redundant flagging happens about 1 in a million pixels.

CHANGES.rst
```diff
@@ -86,6 +86,9 @@ jump
 - Added more allowable selections for the number of cores to use for
   multiprocessing [#183].
 
+- Fixed the numer of rows per slice in multiprocessing that lead to
```
hbushouse, line 89:
```suggestion
- Fixed the computation of the number of rows per slice for multiprocessing, which was leading to
```

src/stcal/jump/jump.py
hbushouse, line 432: Do you really want to use `bitwise_or` here, as opposed to `bitwise_and`? Using `or` will result in pixels that have the JUMP_DET flag set, but not DO_NOT_USE, getting reset to DO_NOT_USE. I think you only want to operate on pixels that have *both* flags set. Same down in line 434, for saturation.

mwregan2: The bitwise_or combination is done first. This combines JUMP_DET with either SATURATION or DO_NOT_USE. So, this is looking for pixels in the gdq that have values of exactly 5 or 6. 
I did test with and without these changes and with the changed code the gdq cubes are identical. Without these two lines there were 5 and 6 gdq values that in the single processing results were 1 and 4 respectively.

hbushouse (approved):
Everything seems to work now.

PR #240, opened 2024-02-01
JP-3525: Merge jump and ramp_fitting docs content with jwst docs

hbushouse:
This PR partially resolves [JP-3525](https://jira.stsci.edu/browse/JP-3525) by making updates to the docs stored in stcal, in order to merge and sort out the info between the docs in jwst and stcal. Related updates will be made in jwst/docs. The vast majority of the high-level step information, especially those things that are JWST-specific, is being moved to jwst/docs, while the stcal/docs retain only low-level algorithmic information.

**Checklist**

- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,19 @@
 1.5.3 (unreleased)
 ==================
 
-- 
+Other
+-----
+
+jump
+~~~~
+
+- Fixed the computation of the number of rows per slice for multiprocessing, which
+  was causing different results when running the step with multiprocess [#239]
+
+ramp_fitting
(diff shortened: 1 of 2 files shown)


hbushouse:
> I understand the hierarchy and linking of the documentation for the different repos is difficult, but much of the documentation being removed from the `STCAL` repo is about how ramp fitting works in `STCAL`. In the future, if someone wants to use `STCAL` for a new telescope, they need to know the information in the documentation being deleted.

I thought about this some more and decided to move a lot more of the basic content from jwst to stcal, with a lot more links/references from the jwst side.

PR #236, opened 2023-12-28
JP-3463: Dark current noise floor

mwregan2:
Resolves [JP-3463](https://jira.stsci.edu/browse/JP-3463)

Closes [spacetelescope/jwst#8071]

This PR addresses the issue described in JP-3463 that the Poisson Noise from the dark current in the pixels is not included in the rate uncertainty.  This change adds the dark rate to the estimated source rate to determine the total Poisson noise.

**Checklist**

- [x] added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)
- [X] updated relevant tests
- [ ] updated relevant documentation
- [ ] updated relevant milestone(s)
- [x] added relevant label(s)

PR #243, opened 2024-02-22
JP-3463: Utilize average_dark_current in variance calculations

tapastro:
Resolves [JP-3463](https://jira.stsci.edu/browse/JP-3463)

Duplicates work in #236 

This PR passes the newly added `RampModel.average_dark_current` to the RampData class, then uses it during poisson variance calculations.

**Checklist**

- [ ] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -54,9 +54,10 @@ def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
     ramp_data = ramp_fit_class.RampData()
 
     if isinstance(model.data, u.Quantity):
-        ramp_data.set_arrays(model.data.value, model.err.value, model.groupdq, model.pixeldq)
+        ramp_data.set_arrays(model.data.value, model.err.value, model.groupdq,
+                             model.pixeldq, model.average_dark_current)
     else:
-        ramp_data.set_arrays(model.data, model.err, model.groupdq, model.pixeldq)
+        ramp_data.set_arrays(model.data, model.err, model.groupdq, model.pixeldq, model.average_dark_current)
 
     # Attribute may not be supported by all pipelines.  Default is NoneType.
     drop_frames1 = model.meta.exposure.drop_frames1 if hasattr(model, "drop_frames1") else None
--- a/src/stcal/ramp_fitting/ramp_fit_class.py
+++ b/src/stcal/ramp_fitting/ramp_fit_class.py
@@ -6,6 +6,7 @@ class RampData:
         self.err = None
         self.groupdq = None
         self.pixeldq = None
+        self.average_dark_current = None
 
         # Meta information
         self.instrument_name = None
@@ -41,7 +42,7 @@ class RampData:
 
         self.current_integ = -1
 
-    def set_arrays(self, data, err, groupdq, pixeldq):
+    def set_arrays(self, data, err, groupdq, pixeldq, average_dark_current):
         """
         Set the arrays needed for ramp fitting.
 
@@ -68,6 +69,7 @@ class RampData:
         self.err = err
         self.groupdq = groupdq
         self.pixeldq = pixeldq
+        self.average_dark_current = average_dark_current
 
     def set_meta(self, name, frame_time, group_time, groupgap, nframes, drop_frames1=None):
         """
--- a/docs/stcal/ramp_fitting/description.rst
+++ b/docs/stcal/ramp_fitting/description.rst
@@ -195,7 +195,7 @@ time in seconds (from the keyword TGROUP).
 The variance of the slope in a segment due to Poisson noise is:
 
 .. math::
-   var^P_{s} = \frac{ slope_{est} }{  tgroup \times gain\ (ngroups_{s} -1)}  \,,
+   var^P_{s} = \frac{ slope_{est} + darkcurrent}{  tgroup \times gain\ (ngroups_{s} -1)}  \,,
 
 where :math:`gain` is the gain for the pixel (from the GAIN reference file),
 in e/DN. The :math:`slope_{est}` is an overall estimated slope of the pixel,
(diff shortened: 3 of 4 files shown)


src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -54,9 +54,10 @@ def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
     ramp_data = ramp_fit_class.RampData()
 
     if isinstance(model.data, u.Quantity):
-        ramp_data.set_arrays(model.data.value, model.err.value, model.groupdq, model.pixeldq)
+        ramp_data.set_arrays(model.data.value, model.err.value, model.groupdq,
+                             model.pixeldq, model.average_dark_current)
     else:
-        ramp_data.set_arrays(model.data, model.err, model.groupdq, model.pixeldq)
+        ramp_data.set_arrays(model.data, model.err, model.groupdq, model.pixeldq, model.average_dark_current)
```
kmacdonald-stsci, line 60: Is `avg_dark_current` an array?  I thought it was a single value, which means it should passed to the `RampData` class using the `set_meta` method.

kmacdonald-stsci (approved):
Is `avg_dark_currrent` an array or a single value?  If it's a single value, it should be passed to the `RampData` class using a different method.  I approved this PR in case it is an array, but if it's not an array, please change the method usage to `set_meta`.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I just read through the tests and see it is an array.  This PR is fine.

src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -41,7 +42,7 @@ def __init__(self):
 
         self.current_integ = -1
 
-    def set_arrays(self, data, err, groupdq, pixeldq):
+    def set_arrays(self, data, err, groupdq, pixeldq, average_dark_current):
```
kmacdonald-stsci, line 45: New parameter needs to be added to docstring.

kmacdonald-stsci (approved):

src/stcal/ramp_fitting/ramp_fit_class.py
tapastro: Added ✅

tapastro:
Looks like I need a concurring review - @hbushouse ?

hbushouse:
Do any of these changes involve changes to API for functions that are called from jwst/ramp_fitting? That'll determine whether we make the new stcal release 1.6.1 or 1.7.0.

tapastro:
I think the answer would be yes, it does. The API arguments aren't changing, in that the number and type of objects passed are not changing; however, the RampModel passed between jwst/ramp_fitting <-> stcal/ramp_fitting is expected to have certain attributes that won't be present if the versions are not in sync.

hbushouse (approved):
LGTM

hbushouse:
> I think the answer would be yes, it does. The API arguments aren't changing, in that the number and type of objects passed are not changing; however, the RampModel passed between jwst/ramp_fitting <-> stcal/ramp_fitting is expected to have certain attributes that won't be present if the versions are not in sync.

OK, that makes sense. Even though function calls aren't changing, the fact that we must have matching versions of code between stcal and jwst does warrant the stcal release being bumped up to 1.7.0. So I think that means you need to update the change log entry (I think it said "1.6.1")?

schlafly:
Thanks, yes, one of us should run the Roman regtests with this PR; I can look at that tomorrow.  I don't expect this to cause issues since we use the other ramp fitting approach, though I could imagine some of the ramp fit class API changes causing issues.

braingram:
I started the romancal regtests here:
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/618/

but then noticed that the downstream jwst and romancal jobs are failing in the CI (so I expect the romancal regtests to also fail).

@tapastro can you confirm that the failures in the jwst downstream job are fixed by:
https://github.com/spacetelescope/stdatamodels/pull/265

@schlafly is it expected to see failures like:
`E       AttributeError: No such attribute (average_dark_current) found in node`
in the romancal downstream job for this PR?

tapastro:
I have a proposed fix for the roman side - I have a question on implementation, though. I didn't realize this code was used by both missions, so I wrote it assuming roman datamodels would not see it. So, I need to supply a dummy array if the input model to the RampData does not have an `average_dark_current` attribute. Would a Roman datamodel (RampModel?) allow me to assign a new attribute, e.g. `roman_model.average_dark_current = np.zeros_like(roman_model.pixeldq)`? The first version of the fix initializes a new array instead.

braingram:
Unfortunately the data models do not allow arbitrary attribute assignment.
```python
>> m.average_dark_current = np.zeros((30, 30), dtype='f4')
AttributeError: No such attribute (average_dark_current) found in node
```
It is *possible* to use a key:
```python
>> m['average_dark_current'] = np.zeros((30, 30), dtype='f4')
```
but I don't know if that's recommended.

Most of the failures appear in:
https://github.com/spacetelescope/romancal/blob/main/romancal/ramp_fitting/tests/test_ramp_fit_ols.py
which is testing the `ols` method that is listed as an option (but not the default) for the `RampFitStep`:
https://github.com/spacetelescope/romancal/blob/main/romancal/ramp_fitting/ramp_fit_step.py#L29

If there are no plans to use `ols` perhaps a fix would be to remove that option (and the tests) in romancal.

braingram:
It looks like all the romancal regtest failures match the ones in the CI so it's likely that getting the CI to go "green" will result in a clean regtest run for romancal.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -53,10 +53,16 @@ def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
     """
     ramp_data = ramp_fit_class.RampData()
 
+    if not hasattr(model, 'average_dark_current'):
```
braingram, line 56: Nice! This (and your other changes) appear to have appeased `romancal`. Let me know when you're ready for another regtest run.

schlafly:
Thank you both!  Yes, I think a mode where, if not set, average_dark_current were treated as zero would be the best for us.  Yes, we also intend to eventually remove support for the ols mode (since we need to be able to support uneven ramps), but we weren't planning on doing that now.

PR #251, opened 2024-03-20
JP-3583: Long integrations truncation in ramp fitting

mwregan2:
Resolves [JP-3583](https://jira.stsci.edu/browse/JP-3583)

Closes spacetelescope/jwst#8373

This PR addresses an issue in the ramp_fitting routine where when finding the cosmic ray free segments of integrations. The code accumulates the non-CR groups using a counter. The problem is that this counter is a uint8 data type. Thus, once we get above 256 groups, it rolls over leading to only using a subset of the groups in the slope fit.

The solution was to change three variables from UINT8 to UINT16. This is big enough given the overall limit on the number of frames.

**Checklist**

- [X] added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)
- [X] updated relevant tests
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,7 +16,13 @@ ramp_fitting
 
 Bug Fixes
 ---------
+ramp_fitting
+~~~~~~~~~~~~
 
+-- Changed the data type of three variables that are used in measuring
+   the jump free segments of integrations. The variables were uint8 and
+   they would yield wrong results for integrations with more than 256
+   groups. [#251]
 Other
 -----
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,13 +16,14 @@ ramp_fitting
 
 Bug Fixes
 ---------
+
 ramp_fitting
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -16,7 +16,13 @@ ramp_fitting
 
 Bug Fixes
 ---------
+ramp_fitting
+~~~~~~~~~~~~
 
+-- Changed the data type of three variables that are used in measuring
+   the jump free segments of integrations. The variables were uint8 and
+   they would yield wrong results for integrations with more than 256
+   groups. [#251]
```
hbushouse, line 25:
```suggestion

ramp_fitting
~~~~~~~~~~~~

- Changed the data type of three variables that are used in measuring
  the jump free segments of integrations. The variables were uint8 and
  they would yield wrong results for integrations with more than 256
  groups. [#251]
```

hbushouse (approved):

PR #252, opened 2024-03-21
use sqrtf to avoid cast to double

braingram:
The use of `sqrt` from libc results in a cast to/from `double` in the ols_cas22 ramp fitting (with jump detection). This can result in small numerical differences that lead to romancal regtests that passed on jenkins but failed locally (due to a single extra jump in one pixel).

This PR switches `sqrt` for `sqrtf` to keep the intermediate values as `float` which leads to (at the moment) the same failure locally and on jenkins (where `test_rampfit_step[spec_full]` fails with a single pixel difference):
```
    {'arrays_differ': {"root['roman']['data']": {'abs_diff': <Quantity 1.9270487 DN / s>,
                                                 'n_diffs': 1,
                                                 'worst_abs_diff': {'index': (2798,
                                                                              2758),
                                                                    'value': <Quantity 1.9270487 DN / s>},
```
Regtest run here: 
https://plwishmaster.stsci.edu:8081/job/RT/job/Roman-Developers-Pull-Requests/681/
The `test_resample` failure is unrelated and also occurs on main.

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,11 @@ ramp_fitting
   they would yield wrong results for integrations with more than 256
   groups. [#251]
 
+- Use ``sqrtf`` instead of ``sqrt`` in ols_cas22 ramp fitting with
+  jump detection to avoid small numerical errors on different systems
+  due to a cast to/from double. [#252]
+
+
 Other
 -----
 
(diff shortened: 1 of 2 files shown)


schlafly (approved):
Looks good to me, thank you!

braingram:
@zacharyburnett can we squeeze this into 1.6.2? If not I'll need to update the changelog entry for this PR.

zacharyburnett (approved):
wow what a simple fix! I'll run these through the romancal regtests on github actions too when it's merged

braingram:
Thanks! I'd hold off re-running them until @schlafly can update the truth files. The change here only breaks jenkins in the same way as the local tests were failing (so this PR will break 2 regtests in romancal until the truth files can be updated).

PR #259, opened 2024-05-21
JP-3593: Saturation step should increase flagging for group 3 saturation

penaguerrero:
Resolves [JP-3593](https://jira.stsci.edu/browse/JP-3593)

Closes spacetelescope/jwst#8413

This PR adds a check for saturation bias in group 2 for nframes > 1.

**Checklist**

- [x] updated relevant tests
- [x] added relevant label(s)


kmacdonald-stsci (approved):
The "resolves" links in the description of the PR should be changed to JP-3593, instead of JP-nnnn, so it links directly to the JP ticket.  The RCAL-nnnn should be removed.

Other than that, everything looks good to me.

braingram:
Is there a RCAL ticket? The romancal tests show some failures:
https://github.com/spacetelescope/stcal/actions/runs/9195566204/job/25291696183?pr=259#step:10:1072

@schlafly is there someone from roman who can look at these changes?

schlafly:
Thanks, yes, romancal will definitely need to be changed to accommodate this.  "number of frames per group" isn't really a good concept for Roman since it can vary from group to group.  I'm having trouble immediately understanding how group 3 is important here.  It's not immediately clear to me that we would want this feature in Roman.  Can you provide a little more context as to what the purpose is?  If we wanted to not change the API, would an option be setting nframes = None by default, and only doing the new block if nframes is not None?

schlafly:
Sorry, now reading the linked https://jira.stsci.edu/browse/JP-3593, I understand the motivation here.

If I have understood correctly, the issue is that the ramp actually is saturating in group 3 due to a very large CR.  In Roman we address this via this dilution factor approach here:
https://github.com/spacetelescope/stcal/blob/main/src/stcal/saturation/saturation.py#L87-L93
That basically turns down the saturation threshold for all groups according to how many reads were in the group to account for dilution of the saturated read among the other unsaturated reads within each group.  I think you could use that approach productively for Webb though it's different than the approach suggested in JP-3953.  I'd suggest we either make this extra nframes argument optional and Roman wouldn't use it, or we follow up in 3953 about whether the dilution factor approach would work there.

src/stcal/saturation/saturation.py
```diff
@@ -85,6 +84,20 @@ def flag_saturated_pixels(
...
+                    mask = data[ints, 0, ...] / np.mean(read_pattern[0]) * read_pattern[2][-1] < sat_thresh
+
+                    # Identify groups with suspiciously large values in the second group
+                    mask &= data[ints, 1, ...] > sat_thresh / len(read_pattern[1])
+
+                    # Identify groups that are saturated in the third group
+                    mask &= np.where(gdq[ints, 2, :, :] & saturated, True, False)
+
+                    # Flag the 2nd group for the pixels passing that gauntlet
+                    gdq[ints, 1][mask] |= saturated
```
hbushouse, line 99: Does the "|" (or) operator really work as desired in this case? Because we track flags in individual bits, don't we need to use "np.bitwise_or()" here?

src/stcal/saturation/saturation.py
```diff
@@ -85,6 +84,20 @@ def flag_saturated_pixels(
             plane = data[ints, group, :, :]
 
             if read_pattern is not None:
```
hbushouse, line 86: I don't see anywhere in the subsequent calls to `flag_saturated_pixels` that `read_pattern` is provided as an argument, so isn't it always going to be `None`?

Also, we already use the "readout pattern" nomenclature to refer to the *name* of a readout pattern, such as RAPID, DEEP8, IRS2, etc. Naming this variable "read_pattern" seems confusing, since as far as I can tell it's a list of readout times (which is totally different than what we normally think of when seeing "read pattern").

hbushouse (commented):
I was under the impression that this new flagging of group 2 (when group 3 is saturated) would only occur when nframes>1. It appears to me that nframes is only ever used to compute the values in "read_pattern", but is not used anywhere in this code to trigger the new algorithm. So does that mean it's going to get applied to all exposures, including those with nframes=1?

src/stcal/saturation/saturation.py
schlafly: Re the naming scheme, this is the name for this quantity in Roman, where of course there's another name for the Webb readout pattern.  Unfortunate collision.

schlafly:
Re only triggering when nframes > 1, the reason this works is that when nframes = 1, len(read_pattern[1]) = 1, so the bit on line 93 becomes just a cut against the normal saturation level.  As you say, it runs even when nframes = 1, but won't flag anything extra in that case.  You could code around it to avoid the checks entirely if wanted.

src/stcal/saturation/saturation.py
hbushouse: Gotcha. Since it's existing code, we on the JWST side can live with that. ;-)

hbushouse:
> Re only triggering when nframes > 1, the reason this works is that when nframes = 1, len(read_pattern[1]) = 1, so the bit on line 93 becomes just a cut against the normal saturation level. As you say, it runs even when nframes = 1, but won't flag anything extra in that case. You could code around it to avoid the checks entirely if wanted.

Thanks for explanation. Makes sense now. Just wasn't obvious from a cursory look at the code.

src/stcal/saturation/saturation.py
```diff
@@ -85,6 +84,20 @@ def flag_saturated_pixels(
...
+                if group == 2:
```
drlaw1558, line 87: I don't think this code will work here;  it's looking for where SATURATED is set in groupdq for group==2, but the flags aren't passed from flag_temp into groupdq until after this code block executes.

src/stcal/saturation/saturation.py
```diff
@@ -85,6 +84,20 @@ def flag_saturated_pixels(
...
+                    # Identify groups which we wouldn't expect to saturate by the third group,
+                    # on the basis of the first group
+                    mask = data[ints, 0, ...] / np.mean(read_pattern[0]) * read_pattern[2][-1] < sat_thresh
+
+                    # Identify groups with suspiciously large values in the second group
+                    mask &= data[ints, 1, ...] > sat_thresh / len(read_pattern[1])
```
drlaw1558, line 93: I think we also need something to grow this mask by n_pix_grow_sat using the adjacent_pixels routine. For pixels bordering pixels that saturate in group 3, the saturation flag is set too. Those pixels might not be intrinsically bright enough to trigger the 'suspiciously large' criterion, but will likewise be affected by the CR and restricted to only fitting a ramp to groups 1+2.

src/stcal/saturation/saturation.py
hbushouse: flag_temp? I don't see that used anywhere in here. Where is it?

drlaw1558: Sorry, copy-paste issue from my review of the jwst PR.  Same general idea though- gdq for the third group isn't set until the code block after all the (if group==2) stuff, so it can never fulfill the criterion of identifying things where saturation was set in the third group.

src/stcal/saturation/saturation.py
```diff
@@ -115,6 +101,25 @@ def flag_saturated_pixels(
...
+
+                    # Identify groups that are saturated in the third group
+                    mask &= np.where(gdq[ints, 2, :, :] & saturated, True, False)
+
+                    # now, flag any pixels that border saturated pixels
+                    if n_pix_grow_sat > 0:
+                        mask = adjacent_pixels(mask, saturated, n_pix_grow_sat)
```
drlaw1558, line 118: I don't think we can call adjacent_pixels directly; it takes an array with dq flag values, and mask is an array of True/False.

penaguerrero: yes, I thought I changed this but now it should be ok. @drlaw1558 can you please verify?

src/stcal/saturation/saturation.py
```diff
@@ -102,6 +101,25 @@ def flag_saturated_pixels(
...
+                    mask = data[ints, 0, ...] / np.mean(read_pattern[0]) * read_pattern[2][-1] < sat_thresh
+
+                    # Identify groups with suspiciously large values in the second group
+                    mask &= data[ints, 1, ...] > sat_thresh / len(read_pattern[1])
+
+                    # Identify groups that are saturated in the third group
+                    mask &= np.where(gdq[ints, 2, :, :] & saturated, True, False)
+
+                    # Flag the 2nd group for the pixels passing that gauntlet
+                    gdq[ints, 1][mask] |= saturated
```
drlaw1558, line 117: I think we need to use a temporary array here; if we set SATURATED for things passing the gauntlet above in the gdq array, and then grow the gdq array, then we'll be growing many pixels that we didn't intend to which didn't pass the gauntlet.  I.e., perhaps define a zero-valued gdq_temp, set SATURATED for things that pass the gauntlet, grow gdq_temp, then add gdq_temp into gdq[ints,1, ...]

penaguerrero:
closing this PR as a new branch is dealing with the issues found for NIRISS and NIRCam data https://github.com/drlaw1558/stcal/tree/jp3593_take2

PR #266, opened 2024-06-18
Revert "pin numpy<2 (#265)" use 2.0 (non-rc) for build

braingram:
This reverts commit 80b842b6cef214a18ff99642ce72da24af18a8e6, reversing changes made to 2b257dff64f9ee54f15afd0c0a600270e2242975.

OpenCV 4.10.0.84 (with numpy 2.0 support) is on pypi. This PR removes the numpy upper pin (and the changelog entry since the change was not released).

This PR also changes the `build-system` requires to use numpy 2.0 instead of the rc.

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)


braingram:
Both jwst and romancal have numpy upper pins (at the moment) which can't be released until other dependencies also remove pins (including `scikit-image`).

kmacdonald-stsci (approved):

hbushouse:
> Both jwst and romancal have numpy upper pins (`<2` at the moment) which can't be released until other dependencies also remove pins (including `scikit-image`).

So you're saying that we don't need an upper pin on numpy in stcal, because for the purpose of the JWST calibration pipeline it'll already be taken care of via the upper pin in the jwst repo?

braingram:
> So you're saying that we don't need an upper pin on numpy in stcal, because for the purpose of the JWST calibration pipeline it'll already be taken care of via the upper pin in the jwst repo?

Unfortunately we still need an upper pin in jwst and romancal due to (at least) scikit-image. All of stcal's dependencies are now up-to-date (with the opencv release) so we can remove the pin in stcal.

hbushouse:
> > So you're saying that we don't need an upper pin on numpy in stcal, because for the purpose of the JWST calibration pipeline it'll already be taken care of via the upper pin in the jwst repo?
> 
> Unfortunately we still need an upper pin in jwst and romancal due to (at least) scikit-image. All of stcal's dependencies are now up-to-date (with the opencv release) so we can remove the pin in stcal.

OK, thanks. In addition to needing numpy<2.0 for compatibility with scikit-image, I think we also don't want to start using numpy 2.0 yet until we've had more time to test and verify any differences in results.

hbushouse (approved):

PR #267, opened 2024-07-01
AL-850: adding tweakreg into stcal

emolter:
Resolves [AL-850](https://jira.stsci.edu/browse/AL-850)
Partially resolves [JP-3668](https://jira.stsci.edu/browse/JP-3668)

This PR moves the majority of the tweakreg functionality from JWST into stcal, such that it can be used by both Roman and JWST without needing to be maintained in both places.  This relates to the alignment between images and to an absolute source catalog only, but not to finding sources within images - this is handled differently by the two observatories, and is split into its own step for Roman.

Several WCS-related utilities have also been moved into `alignment` or modified therein.  These should also eventually be called by JWST and Roman from stcal and removed in those repositories.

**Checklist**

- [x] updated relevant tests
- [x] added relevant label(s)


src/stcal/alignment/util.py
```diff
@@ -20,16 +20,19 @@
...
+    "update_s_region_keyword",
     "wcs_from_footprints",
     "reproject",
+    "SupportsDataWithWcs"
 ]
 
 
+@runtime_checkable
 class SupportsDataWithWcs(Protocol):
     _asdf: AsdfFile
```
mcara, line 37: In `util.py`, `ref_model` is used in only one way: `refmodel.meta.wcs`. Here it can definitely be replaced by `ref_wcs` so that we do not need an interface to data models.

docs/api/stcal.alignment.calc_rotation_matrix.rst
braingram, line 1: Did these docs sneak into a commit? They look auto-generated so perhaps the gitignore is missing en entry to ignore them.

src/stcal/alignment/util.py
emolter: This is mostly true; it's used only for `model.meta.wcs` and `model.meta.wcsinfo`.  Plus there is a singular instance where `model.meta.coordinates.reference_frame` is checked, in the function `wcsinfo_from_model()`.  I agree with you that these instances can all be changed to take `wcs`, `wcsinfo`, and `referece_frame` directly as inputs.

There are some consequences to doing that.  The call structure of the `update_s_region_imaging`, `update_s_region_keyword`, and `wcs_from_footprints` functions have to change, and these are all used directly in the JWST repository.  These changes maybe slightly decrease ease-of-use: e.g. for `wcs_from_footprints()` there are now three required arguments instead of just one.  But I think that's okay.

I'm still a bit torn about whether something like this protocol should be implemented for stcal/tweakreg.  At present, that module uses `model.meta.wcs`, `model.meta.wcsinfo`, `model.meta.group_id`, and `model.meta.observation.date`.  However, after doing the refactor of `util.py` it seems a lot more tractable to remove the dependencies on datamodels entirely... I'm going to see what that looks like now.

docs/api/stcal.alignment.calc_rotation_matrix.rst
emolter: It appears so.  I did run `make html` and `make clean html` to test the docs build, but I never touched the .gitignore.  Can you advise what is needed in there?

emolter: ok, will do. confirmed, none of the API was manually generated

src/stcal/alignment/util.py
emolter: How do the new changes look to you @mcara ?  The corresponding jwst PR incorporates the changes too and is found [here](https://github.com/spacetelescope/jwst/pull/8624).

src/stcal/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,203 @@
...
+
+    return radius, fiducial
+
+
+def get_catalog(ra, dec, epoch=2016.0, sr=0.1, catalog="GAIADR3"):
+    """Extract catalog from VO web service.
+
+    Parameters
+    ----------
+    ra : float
```
kmacdonald-stsci, line 166: I suggest changing the variable names of `ra` and `dec` to something more verbose.

src/stcal/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,203 @@
...
+        Right Ascension (RA) of center of field-of-view (in decimal degrees)
+
+    dec : float
+        Declination (Dec) of center of field-of-view (in decimal degrees)
+
+    epoch : float, optional
+        Reference epoch used to update the coordinates for proper motion
+        (in decimal year). Default: 2016.0
+
+    sr : float, optional
```
kmacdonald-stsci, line 176: I suggest making this variable name more verbose.

tests/test_tweakreg.py
```diff
@@ -0,0 +1,304 @@
...
+    relative_align,
+    absolute_align,
+    TweakregError,
+    _parse_refcat
+)
+
+# Define input GWCS specification to be used for these tests
+WCS_NAME = "mosaic_long_i2d_gwcs.asdf"  # Derived using B7.5 Level 3 product
+EXPECTED_NUM_SOURCES = 2469
+EXPECTED_RADIUS = 0.02564497890604383
```
kmacdonald-stsci, line 28: This is used in only one place, so doesn't need to be a global variable.

kmacdonald-stsci (approved):

src/stcal/tweakreg/tweakreg.py
```diff
@@ -0,0 +1,342 @@
...
+    if nelem == 1:
+        return str_list
+    if nelem == 2:
+        return str_list[0] + " or " + str_list[1]
+    return ", ".join(map(repr, str_list[:-1])) + ", or " + repr(str_list[-1])
+
+
+_SQRT2 = math.sqrt(2.0)
+SINGLE_GROUP_REFCAT = ["GAIADR3", "GAIADR2", "GAIADR1"]
+_SINGLE_GROUP_REFCAT_STR = _oxford_or_str_join(SINGLE_GROUP_REFCAT)
```
braingram, line 38:
```suggestion
_SINGLE_GROUP_REFCAT_STR = ",".join(SINGLE_GROUP_REFCAT)
```
I don't think we need `_oxford_or_str_join` (and can remove the function with this change) here and can accept not having "or"s.

src/stcal/tweakreg/tweakreg.py
```diff
@@ -0,0 +1,342 @@
...
+                    "adjust 'tolerance' and/or 'separation' parameters. "
+                    "Alignment to an absolute reference catalog will "
+                    "not be performed.")
+            raise TweakregError(msg) from None
+        raise
+
+    return correctors
+
+
+def _parse_refcat(abs_refcat: str | Path,
```
braingram, line 225:
```suggestion
def _parse_refcat(abs_refcat: str,
```
The "lower" usage below won't work for a `Path`. This was found by manually running mypy on this branch. I don't think the type hints are tested in the CI. Would you run mypy to check these?

braingram:
Once the unit tests are sorted would you run the jwst and romancal regtests with this PR?

The romancal unit tests are currently showing failures in part due to the change in signature for `wcs_from_footprints`:
https://github.com/spacetelescope/stcal/actions/runs/9975888017/job/27566761372?pr=267#step:10:10078

emolter:
> Once the unit tests are sorted would you run the jwst and romancal regtests with this PR?
> 
> The romancal unit tests are currently showing failures in part due to the change in signature for `wcs_from_footprints`: https://github.com/spacetelescope/stcal/actions/runs/9975888017/job/27566761372?pr=267#step:10:10078

Yeah, I definitely expect failures in both jwst and romancal because of the call signature changes that were necessary to avoid adding `datamodel` as a dependency to stcal.  I am not sure whether I have permission to start regtest runs for Roman, and I'm not sure whether it's my responsibility or the responsibility of someone else to modify romancal to account for these changes in stcal.  But I will go ahead and start another JWST regtest run based on this PR branch and the [corresponding JWST PR branch](https://github.com/spacetelescope/jwst/pull/8624).

src/stcal/tweakreg/tweakreg.py
mairanteodoro: `lower()` should work with `Path` if it is wrapped with `str()`: `str(abs_refcat).lower()`

src/stcal/tweakreg/tweakreg.py
```diff
@@ -0,0 +1,342 @@
...
+                  wcsinfo: dict,
+                  epoch: str | astropy.time.Time,
+                  save_abs_catalog: bool = False,
+                  output_dir: str | None = None) -> Table:
+    """
+    Figure out if abs_refcat is an input filename or
+    the name of a GAIA catalog. If the former, load it,
+    and if the latter, retrieve that catalog from the Web.
+    If desired, save the reference catalog in the specified directory.
+    """
```
mairanteodoro, line 237: Maybe we should clear the issue with `abs_refcat` being a Path object here, by setting `abs_refcat = str(abs_refcat)`.

src/stcal/tweakreg/tweakreg.py
```diff
@@ -0,0 +1,342 @@
...
+                        ref_wcsinfo=wcsinfo,
+        )
+
+        return create_astrometric_catalog(
+            combined_wcs, epoch,
+            catalog=gaia_cat_name,
+            output=output_name,
+        )
+
+    if Path.is_file(Path(abs_refcat)):
```
mairanteodoro, line 263: If `abs_refcat` is a `str`, then we can do `if Path(abs_refcat).is_file()`.

src/stcal/tweakreg/tweakreg.py
```diff
@@ -0,0 +1,342 @@
...
+    If desired, save the reference catalog in the specified directory.
+    """
+    if save_abs_catalog:
+        root = f"fit_{abs_refcat.lower()}_ref.ecsv"
+        output_name = Path(root) if output_dir is None \
+            else Path(output_dir) / root
+    else:
+        output_name = None
+
+    abs_refcat = str(abs_refcat).strip()
```
mairanteodoro, line 245: If `abs_refcat` is already a `str`, we can remove the parsing.

tests/.gitignore
```diff
@@ -0,0 +1,4 @@
+/*.fits
```
mairanteodoro, line 1: Are we saving files in the working directory that need to be removed (for example, after running the unit tests)?

mairanteodoro (commented):
Looks good to me! Thanks, @emolter! I left a few comments.

emolter:
The mypy checking suggested by @braingram raised an additional issue, which is that `wcsinfo` isn't actually a dict, but an `stdatamodels.properties.ObjectNode`, and its attributes are called by many of these functions.  I see two ways to fix this: 

1. change all instances of `wcsinfo.property` to `wcsinfo["property"]` and then pass `wcsinfo.instance` into all the functions that currently require `wcsinfo`, since `wcsinfo.instance` is a dict.
2. write an adapter class that takes in an `Any` type and attempts to turn it into a `wcsinfo` dict.

I'm tempted to go with the former.  It will require even more changes to the jwst and romancal repositories, but those changes are already needed to do away with the `SupportsDataWithWCS` protocol.  Thoughts?

mairanteodoro:
> The mypy checking suggested by @braingram raised an additional issue, which is that `wcsinfo` isn't actually a dict, but an `stdatamodels.properties.ObjectNode`, and its attributes are called by many of these functions.

@emolter I'm not sure about JWST but for Roman `meta.wcsinfo` is of type roman_datamodels.stnode.Wcsinfo and it has a method `.to_flat_dict()` to parse it into a flat dict.

emolter:
> > The mypy checking suggested by @braingram raised an additional issue, which is that `wcsinfo` isn't actually a dict, but an `stdatamodels.properties.ObjectNode`, and its attributes are called by many of these functions.
> 
> @emolter I'm not sure about JWST, but for Roman `meta.wcsinfo` is of type `roman_datamodels.stnode.Wcsinfo` and it has a method `.to_flat_dict()` to parse it into a flat dict.

thanks, that exists in jwst as well.  Before, the `SupportsDataWithWCS` class required the `to_flat_dict` method, but it sounded like people were interested in removing it.  Instead of doing that, I guess we could go for the option of only applying `SupportsDataWithWCS` to `wcsinfo` objects, not to (for lack of a better word) "datamodel-like" objects

tests/.gitignore
emolter: Yes, `test_tweakreg.py` is creating `ref_cat.ecsv` in a unit test I moved from the jwst repo.  `test_jump.py` was already creating `data.fits` and `gdq.fits`, which is likely (I didn't test this) happening on main.

Admittedly, adding the .gitignore is a lazy band-aid.  Any suggestions for how to ensure the tests clean up after themselves?

emolter:
Ok, I put a proposed fix to the `wcsinfo` typing issue into the most recent push.  In short, I wrote a protocol for the `wcsinfo` object specifying that it needs a `.instance` attribute that converts it to a dict, as well as attributes corresponding to the expected metadata fields (e.g. `roll_ref`, `v3yangle`, etc.).  All the methods that take in a `wcsinfo` object now also accept a `dict`.  

I chose this design because I think the protocol is specific enough that a different observatory outside the stsci umbrella could straightforwardly adapt it to their needs (That is the original idea of stcal, right?) while the romancal and jwst repos won't need to pass in `wcsinfo.instance` or similar every time they want to use one of the utility functions.

I was wrong that `wcsinfo` in the jwst repo has a `to_flat_dict()` method, so I am using `.instance` instead - is that going to work for Roman?

I also think all other comments have been addressed.

Let me know what you think

braingram:
> Yeah, I definitely expect failures in both jwst and romancal because of the call signature changes that were necessary to avoid adding `datamodel` as a dependency to stcal. I am not sure whether I have permission to start regtest runs for Roman, and I'm not sure whether it's my responsibility or the responsibility of someone else to modify romancal to account for these changes in stcal. But I will go ahead and start another JWST regtest run based on this PR branch and the [corresponding JWST PR branch](https://github.com/spacetelescope/jwst/pull/8624).

I think it's best if we address the known failures first.

As the changes will constitute a breaking change we'll need to consider:
- updating pins in jwst/romancal (more on this below)
- should this trigger a stcal 2.x.x

The currently release romancal does not have an upper pin for stcal:
https://github.com/spacetelescope/romancal/blob/008e9ad726be3c84aa54c6979060d1793e7a5667/pyproject.toml#L31
jwst has one set to "<1.8.0":
https://github.com/spacetelescope/jwst/blob/d4bd628cf527ae3b0bd33a191cdb778f0f532293/pyproject.toml#L37

This PR should at least trigger 1.8.0 (to prevent breaking jwst). romancal is in a worse state (I think that's more a sign that we should always pin stcal in romancal). Working with what we have I'm wondering if it makes more sense to deprecate the old functions if the old signatures can't be supported.

Do you think the api changes that are breaking romancal are required?

mcara (approved):
I think this looks great. Thanks!

mcara:
I also think that the original approach with the interface was not bad actually. Many years ago the thought was that `NDData` would be a data structure used by astronomical data software and provide a common interface definition that would hold data/error/variance/etc. arrays and WCS information. I think having a common base class (not necessarily `NDData`) for JWST and Roman that would provide a common interface would be great.

src/stcal/alignment/util.py
```diff
@@ -14,32 +20,39 @@
...
     "wcs_from_footprints",
     "reproject",
+    "Wcsinfo"
 ]
 
 
-class SupportsDataWithWcs(Protocol):
-    _asdf: AsdfFile
+@runtime_checkable
+class Wcsinfo(Protocol):
```
braingram, line 39:
```suggestion
class Wcsinfo(Protocol):
```
I don't see any uses of this outside of the type hints. The `runtime_checkable` can be removed.

src/stcal/alignment/util.py
```diff
@@ -115,12 +131,14 @@ def _generate_tranform(
     transform : ~astropy.modeling.Model
         An :py:mod:`~astropy` model containing the transform between frames.
     """
+    if not isinstance(wcsinfo, dict):
+        wcsinfo = wcsinfo.instance
```
braingram, line 135: I just tested a roman `cal` file from artifactory that has a wcsinfo.

The wcsinfo object fails an `isinstance(wcsinfo, dict)` check (it is of type `roman_datamodels.stnode.Wcsinfo`).

If I switch this to `isinstance(wcsinfo, WcsInfo)` it fails for 2 reasons:
- missing `instance` method
- missing `wcsaxes`

pyproject.toml
```diff
@@ -19,6 +19,8 @@ dependencies = [
     "opencv-python-headless >=4.6.0.66",
     "asdf >=2.15.0",
     "gwcs >= 0.18.1",
+    "tweakwcs >=0.8.6",
```
braingram, line 22: Would you add these to the devdeps?
https://github.com/spacetelescope/stcal/blob/eb8528c21e0ffbbdec89f5e1d7323425850d3b69/tox.ini#L56

src/stcal/alignment/util.py
```diff
@@ -3,7 +3,13 @@
...
 import functools
 import logging
-from typing import TYPE_CHECKING, Protocol
+from typing import TYPE_CHECKING, Protocol, runtime_checkable
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Sequence
+
+    import astropy
+    import numpy  # noqa: ICN001 this is the only way Sphinx can find ndarray
```
braingram, line 12: What's the sphinx error that required this? Is it due to the type hints?

src/stcal/tweakreg/tweakreg.py
```diff
@@ -0,0 +1,340 @@
...
+        Dictionary containing WCSreference angles.
+    """
+    if not isinstance(refang, dict):
+        refang = refang.instance
+    catalog = filter_catalog_by_bounding_box(
+        catalog, wcs.bounding_box)
+
+    return JWSTWCSCorrector(
+        wcs=wcs,
+        wcsinfo={"roll_ref": refang["roll_ref"],
```
braingram, line 330: I'm seeing a mypy error here:
```
Value of type "dict[Any, Any] | Wcsinfo" is not indexable  [index]
```

braingram (commented):
I did a bit of testing with a roman file and left a few comments.

Also I ran mypy with this PR and it reports 50 errors (it reports 34 on main). Some of these are due to missing type stubs, some are due to code on main, and some due to code in this PR.

Would you fix the type hint errors added in this PR? (I left a comment about one but after seeing the number figured I'd leave a general comment).

I think this PR would benefit from mypy or some other tool checking these types in the CI. Providing incorrect type hints seems worse than none. I think adding a type checker is beyond the scope of this PR and perhaps makes sense to add in a separate PR that is merged first (so that the type hints in this PR can be checked).

src/stcal/alignment/util.py
emolter: The error is `/Users/emolter/repositories/stcal/src/stcal/alignment/util.py:docstring of stcal.alignment.util.compute_fiducial:: WARNING: py:class reference target not found: np.ndarray`

It seems to be due to a sphinx bug, but the problem, strangely, only seems to occur when the _output_ type is abbreviated to `np.ndarray`, not when any of the _inputs_ are `np.ndarray`

emolter:
@braingram sorry for making you ask twice about mypy and type hits.  It may not look like it (apparently it doesn't), but I did look through those, I just didn't see many of them as problems.  However I now realize that some small changes could fix these:

- _Skipping analyzing "astropy": module is installed, but missing library stubs or py.typed marker._  These, and similar, are (to my understanding) beyond my control and have to do with the way those packages are themselves set up.  This represents a large fraction of the errors.  If there's a way to get mypy to ignore those, that would probably be helpful.
- _Value of type "dict[Any, Any] | Wcsinfo" is not indexable._  These were due to the fact that I didn't type hint the `.instance` function definition in the `Wcsinfo` protocol.  This has been fixed.
- _alignment/util.py:196: error: Incompatible types in assignment (expression has type "tuple[tuple[Any, Any], ...]", variable has type "list[tuple[Any, Any]]")._ This was due to an unnecessary explicit assignment of `output_bounding_box` to type `tuple` when it seems to be fine for all purposes to have either a `list` or `tuple`.  There were other similar errors that could be fixed by making all type hints consistently Sequence.
- _alignment/util.py:278: error: "None" object is not iterable._ and some similar errors related to allowing None as a type.  These were caused by problems in function logic that could theoretically allow a NoneType to pass through `if var is None` kinds of checking (although in some cases mypy seemed not to follow the logic quite right)

In any case, the only mypy issues left now should be due to missing library stubs.

src/stcal/alignment/util.py
emolter: Good to know.  What do you think is the best way to handle this?  I agree maybe checking `isinstance(wcsinfo, WcsInfo)` is a good plan, then checking `isinstance(wcsinfo, dict)` and if it's neither, then perhaps raise an error.  But that doesn't solve the issue that jwst and roman have different ways of accessing `wcsinfo` as a dict. (as an aside, I find this a bit strange).  I could check for either `.instance` or `.to_flat_dict`, but I'm not sure how to set up a protocol to do that, or if that's even a good idea.

Regarding `wcsaxes`, that attribute is used only in the function `wcsinfo_from_model`.  Looking again, I am not sure why `wcsinfo_from_model()` ended up in stcal in the first place... I don't see it imported in romancal nor jwst, and jwst has its own version of it inside `pointing.py` that looks identical (except the changes I made that remove dependence on the datamodel structure).  This was introduced in [this commit](https://github.com/spacetelescope/stcal/commit/0a1fa458d1a9dd6b6ba5dcd392e258ec79509383) related to resample. @mairanteodoro can you confirm whether or not this is still needed?  If not, I will go ahead and remove the function entirely.

src/stcal/tweakreg/tweakreg.py
emolter: should be fixed- see my comment below

src/stcal/alignment/util.py
braingram: Thanks. I can also see that error in the current docs build. I'm not sure what's going here either. The sphinx version is pinned:
https://github.com/spacetelescope/stcal/blob/eb8528c21e0ffbbdec89f5e1d7323425850d3b69/pyproject.toml#L34
since:
https://github.com/spacetelescope/stcal/commit/3baee3ba572e55ef77c33b00ae45625a775468b1
I tried a newer version and saw the same issue.

Do we perhaps need a sphinx extension or something to allow the docs to use the `ndarray` type hint? Changes to the code to fix the docs seem less than ideal.

emolter: I agree, although the code change only has to do with type hints, so maybe it's okay?

braingram: I'm not familiar enough with type hints and the typical docs tools to know what's going on or if there are ways to fix it that don't involve code changes. I'll mark this as resolved since it seems to fix the issue and hasn't raised any other objections.

braingram:
No worries and thanks for rerunning them. I'm also seeing 34 errors (the same as main) after the changes you've added. Thanks for the details on the failures. I'm certainly new to type hints and this PR has taught me a lot.

src/stcal/alignment/util.py
braingram: At the moment I'm not sure what to recommend and would have to dig into the code to come up with one.

Ideally this PR wouldn't break romancal. Is that possible?

emolter: @braingram The root cause of all the ways in which this will break romancal (and jwst for that matter) is that the call structure of certain `stcal.alignment.util` functions was changed to remove the need for the `SupportsDataWithWCS` protocol.  Some of these functions are used in resample as well as tweakreg, namely `wcs_from_footprints()` and `update_s_region_imaging()`.

My takeaway from Thursday's meeting was that people were satisfied with those changes, as that protocol was not very well-defined and we would like to remove dependency on the datamodel structure.  It sounded from yesterday's meeting like there was no problem with making a patch of romancal that pinned an upper dependency of stcal and then modifying those function calls as part of @mairanteodoro's work to make the romancal tweakreg step rely on stcal.  Is it no longer the case that this solution is okay?

Of course other ways of doing this are are possible: for example, to @mcara's point, I could write basically a datamodel-to-wcs-and-wcsinfo interface, and then revert all the call structure changes.  I view that solution as a clunky and non-permanent fix, and it seems to me it would have to be changed again as soon as we start moving some of these functions to tweakwcs and gwcs.

Making it so the Wcsinfo object I introduced into stcal can use either the `to_flat_dict()` or `.instance()` method to return a dictionary I think is a good idea, and I'll go ahead with that.  However, I still think it's strange that they are different between jwst and romancal, and I wonder if we should consider changing one of them to make them consistent.

schlafly: FWIW, we did decide yesterday at the romancal tag to do a new romancal release with an upper pin on stcal to enable this.

emolter:
I put some lines of mypy configuration into pyproject.toml, and there are now zero failures with mypy using that configuration.  

For some reason, mypy wasn't picking up the configuration unless I specified explicitly `mypy filename --config-file path/to/pyproject.toml` but I assume that if a mypy check is ever added to our CI workflow it will pick up the proper configuration in pyproject.toml

src/stcal/alignment/util.py
braingram: > My takeaway from Thursday's meeting was that people were satisfied with those changes, as that protocol was not very well-defined and we would like to remove dependency on the datamodel structure. In the meeting it sounded like there was no problem with making a patch of romancal that pinned an upper dependency of stcal and then modifying those function calls as part of @mairanteodoro's work to make the romancal tweakreg step rely on stcal. Is it no longer the case that this solution is okay?

It's my understanding that solution is still ok. However the use of `wcsinfo.instance` meant that to fix romancal would involve either:
- removing the `wcsinfo.instance` use here in stcal (which is being introduced with this PR)
- fundamental changes to roman datamodels (which would be no small undertaking)

I see that you've updated the PR to remove the `instance` use. Thanks for updating the PR. I only brought it up because I tested this with a roman file and encountered an issue that would come up when attempting to update romancal.

emolter: Ok great, glad we are on the same page.  I got confused and assumed you were still concerned about `wcs_from_footprints`.

There should hopefully be no need to change roman datamodels now, and I think the `Wcsinfo` protocol is now as simple as possible while still being specific about its needs.

braingram:
> I put some lines of mypy configuration into pyproject.toml, and there are now zero failures with mypy using that configuration.
> 
> For some reason, mypy wasn't picking up the configuration unless I specified explicitly `mypy filename --config-file path/to/pyproject.toml` but I assume that if a mypy check is ever added to our CI workflow it will pick up the proper configuration in pyproject.toml

Thanks! I opened a PR against your branch to run mypy in the CI:
https://github.com/emolter/stcal/pull/1

tox.ini
```diff
@@ -15,6 +15,14 @@ envlist =
 #     tox -l -v
 #
 
+[testenv:check-types]
+description = check type hints, e.g. with mypy
+deps =
+    mypy
```
braingram, line 21:
```suggestion
    mypy
    types-requests
```

tox.ini
```diff
@@ -15,6 +15,14 @@ envlist =
...
+commands =
+    mypy --install-types --non-interactive
```
braingram, line 23:
```suggestion
```

braingram: It looks like `--install-types` requires a previous run so let's try just installing the requests type stubs: https://github.com/spacetelescope/stcal/pull/267/files#r1684549714

emolter: looks to be working, thanks!

braingram (approved):
Thanks for all your work on this.

braingram (changes requested):
One more change. @emolter would you change the `CHANGES.rst` version to 1.8.0 for this PR?

emolter:
Do not merge at this time. somehow one of the last small changes broke s_region values in jwst. looking into it

emolter:
Long story short, `isinstance(wcsinfo, Wcsinfo)` returns `False` for JWST wcsinfo objects, of type ObjectNode.  However, if I do
```
for attr in ["ra_ref", "dec_ref", ...]:
    assert hasattr(wcsinfo, attr)
```
all of the assert statements pass.  The [documentation](https://typing.readthedocs.io/en/latest/spec/protocol.html#runtime-checkable-decorator-and-narrowing-types-by-isinstance) and other snooping I've done seem to indicate that a runtime_checkable protocol _should_ be basically shorthand for all the `hasattr` statements, so I'm pretty confused about what's happening there.  My only idea (based on [PEP 544](https://peps.python.org/pep-0544/#support-isinstance-checks-by-default)) is that these ObjectNode objects are not initialized until called explicitly, but I don't know enough about those objects to be sure.

I tried updating all the JWST code to pass in `wcsinfo` as a `dict` in all cases (using `wcsinfo.instance`), but this leads to the issue that updating keywords in the `wcsinfo` object via e.g. `update_s_region_keyword` becomes cumbersome and IMO unnecessarily verbose (it becomes necessary to reassign all the values from the dict to the wcsinfo object after the fact).

While perhaps not perfect, changing back to `if not isinstance(wcsinfo, dict):` achieves the desired behavior.  While technically the `not` would make possible passing any other type through, the type hint (still using the custom Wcsinfo protocol but no longer runtime checkable) specifies what attributes a `wcsinfo` object needs to have for these routines to run properly.  Let me know if that solution sounds ok to you @braingram.

emolter:
Ok, all the regression tests for the JWST repo are now passing.  It may be prudent to wait for more approvals of [that PR](https://github.com/spacetelescope/jwst/pull/8624), just in case changes are requested that require updating the stcal side for some reason.  Other than that, this is ready to merge.  @zacharyburnett what is the status of the romancal release - is it now safe to merge into stcal?  @braingram are you happy with the changes now, and if so, can you remove the "changes requested" status?

tox.ini
```diff
@@ -15,6 +15,14 @@ envlist =
 #     tox -l -v
 #
 
+[testenv:check-types]
+description = check type hints, e.g. with mypy
+deps =
+    mypy
+    types-requests
+commands =
+    mypy src/stcal --config-file pyproject.toml
```
zacharyburnett, line 24: nice addition!

tests/test_alignment.py
```diff
@@ -71,6 +71,9 @@ def _create_wcs_and_datamodel(fiducial_world, shape, pscale):
 
 
 class WcsInfo:
+    """
```
braingram, line 74: This object (and `DataModel`) isn't added by this PR but I think could be removed with the changes in this PR. I'm not suggesting that we remove these as part of this PR. I opened an issue to track their removal: https://github.com/spacetelescope/stcal/issues/273

braingram (approved):
Thanks!

src/stcal/alignment/util.py
```diff
@@ -733,17 +721,17 @@ def update_s_region_keyword(model, footprint):
...
-    else:
-        model.meta.wcsinfo.s_region = s_region
-        log.info("Update S_REGION to %s", model.meta.wcsinfo.s_region)
+        return None
+    log.info("Update S_REGION to %s", s_region)
+    return s_region
 
 
-def reproject(wcs1, wcs2):
+def reproject(wcs1: gwcs.wcs.WCS, wcs2: gwcs.wcs.WCS) -> Callable:
```
nden, line 734: I think this function needs to replicate what the[ jwst version](https://jira.stsci.edu/browse/RSOCDR-145) does, which is the most general way. The use case that this is missing is passing `sliced_wcs.SlicedLowLevelWCS` object. It's a JWST use case, that was requested by external users. There are tests for this in jwst:
https://github.com/spacetelescope/jwst/blob/master/jwst/resample/tests/test_utils.py#L191
Were they not moved to stcal or am I not seeing them?

emolter: @nden Thanks for the comment.  I think the first link goes to something unrelated, unless I'm missing something.  There are two `reproject` functions in `jwst`, one in `assign_wcs.util` and one in `resample.resample_utils`.  The one in `resample` is not affected by this PR, and is the one that those tests cover.

In my opinion, consolidating these functions falls under the scope of [this issue](https://github.com/spacetelescope/stcal/issues/271) and does not need to be done here.  But I can work on changing it if you like

nden (approved):

mairanteodoro (commented):
Looks good to me. Thanks, @emolter!

PR #269, opened 2024-07-05
JP-3678: Fix C poisson variance calculation

drlaw1558:
Resolves [JP-3678](https://jira.stsci.edu/browse/JP-3678)

This PR fixes the C implementation of the poisson variance calculations when provided an average dark current.  It ensures appropriate weighting of the average dark when the average ramp is negative (due to influence of very small segments on the simple average ramp calculation) and ensures that the variance is not set to zero whenever the average ramp is below zero.  Largely, it follows the same kind of modifications made to the python routine previously in https://github.com/spacetelescope/stcal/pull/255

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,12 @@ Changes to API
 Bug Fixes
 ---------
 
-- 
+ramp_fitting
+~~~~~~~~~~~~
+
+- Fix bugs in the C algorithm Poisson variance calculation when provided with
+  an average dark current. [#269]
+
 
 1.7.2 (2024-06-12)
 ==================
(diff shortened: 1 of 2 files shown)


drlaw1558:
Adding a note that I've tested this on jw01047002001_02101_00001_mirifushort_uncal, jw01236004001_02103_00001_mirifulong_uncal, jw01523003001_03102_00001_mirifushort_uncal, and jw01970002001_02101_00003_nrs1_uncal

SCI, ERR, and VAR_POISSON for these now all match the results from the python code as expected, with and without providing an average_dark_current input.

nden:
Regression tests starting here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/

kmacdonald-stsci (approved):

PR #270, opened 2024-07-08
Move outlier detection utility functions from jwst to stcal

braingram:
This PR moves a number of utility functions related to outlier detection from jwst to stcal.

This PR is required for https://github.com/spacetelescope/jwst/pull/8613

As this PR adds a few new dependencies (and these are added to the devdeps) the `run devdeps tests` label was added. It's expected that the jwst devdeps will fail as they are failing on main: https://github.com/spacetelescope/stcal/actions/runs/9836681142/job/27152917937 

**Checklist**

- [x] updated relevant tests
- [x] added relevant label(s)


braingram:
Closing and opening to re-trigger CI since the `run devdeps tests` label didn't seem to take.

src/stcal/outlier_detection/utils.py
```diff
@@ -0,0 +1,337 @@
...
+    # Sigma-clip the unmasked data
+    weight_masked = sigma_clip(weight_masked, sigma=3, maxiters=5)
+    mean_weight = np.mean(weight_masked)
+    # Mask pixels where weight falls below maskpt percent
+    weight_threshold = mean_weight * maskpt
+    return weight_threshold
+
+
+def _abs_deriv(array):
+    # but fixing the bugs will likely change the output
```
emolter, line 88: This comment looks deprecated? or at least isolated from any context about what the bug is

src/stcal/outlier_detection/utils.py
```diff
@@ -0,0 +1,337 @@
...
+
+def flag_crs(
+    sci_data,
+    sci_err,
+    blot_data,
+    snr,
+):
+    """
+    Straightforward detection of outliers for non-dithered data since
+    err_data includes all noise sources (photon, read, and flat for baseline).
```
emolter, line 120:
```suggestion
    sci_err includes all noise sources (photon, read, and flat for baseline).
```

src/stcal/outlier_detection/utils.py
```diff
@@ -0,0 +1,337 @@
...
+    # what we've been doing up until now, so more investigation is needed
+    # before a change is made.  Preferably, fix tblot in drizzle.
+    pixmap[np.isnan(pixmap)] = -1
+    tblot(median_data, pixmap, outsci, scale=pix_ratio, kscale=1.0,
+          interp='linear', exptime=1.0, misval=0.0, sinscl=1.0)
+
+    return outsci
+
+
+def calc_gwcs_pixmap(in_wcs, out_wcs, in_shape):
```
emolter, line 270: is this identical to (or combine-able with) the existing `calc_pixmap` function in `resample_utils`? https://github.com/spacetelescope/stcal/blob/eb8528c21e0ffbbdec89f5e1d7323425850d3b69/src/stcal/alignment/resample_utils.py#L12

src/stcal/outlier_detection/utils.py
```diff
@@ -0,0 +1,337 @@
...
+        Computed pixmap.
+    """
+    bb = wcs_bbox_from_shape(in_shape)
+    log.debug("Bounding box from data shape: {}".format(bb))
+
+    grid = gwcs.wcstools.grid_from_bounding_box(bb)
+    return np.dstack(reproject(in_wcs, out_wcs)(grid[0], grid[1]))
+
+
+def reproject(wcs1, wcs2):
```
emolter, line 297: Similar to a comment you made to me, I wonder where this utility and similar ones, like `calc_gwcs_pixmap`, should go - tweakwcs? stcal alignment utils? here?  I'm fine merging this with it here, but perhaps a discussion we should open sooner rather than later.

src/stcal/outlier_detection/utils.py
```diff
@@ -0,0 +1,337 @@
...
+        raise TypeError("Input should be a WCS") from err
+
+    def _reproject(x, y):
+        sky = forward_transform(x, y)
+        flat_sky = []
+        for axis in sky:
+            flat_sky.append(axis.flatten())
+        # Filter out RuntimeWarnings due to computed NaNs in the WCS
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", RuntimeWarning)
```
emolter, line 331: at some point I updated all the warnings in the `jwst` repo to use `warnings.filterwarnings(whatever, whatever, message="specific message")` to be absolutely sure we don't accidentally suppress unintended warnings.  Could you do the same here?

emolter (commented):
Looking good, just a few small things.

src/stcal/outlier_detection/utils.py
emolter: One more thing about this function I just realized when looking through open JWST PRs: it looks like the reproject function is getting either substantial changes or removal in the near future: https://github.com/spacetelescope/jwst/pull/8554

braingram: Unfortunately the `calc_gwcs_pixmap` in ailgnment.utils uses a `reproject` function that isn't compatible with gwcs. Here's the line that failed when I tried to use it:
https://github.com/spacetelescope/stcal/blob/eb8528c21e0ffbbdec89f5e1d7323425850d3b69/src/stcal/alignment/util.py#L821
@nden mentioned that the function is using the FITS convention (which in this case fails for the gwcs.WCS).

Consolidating the many `reproject`s (and other duplicated functions) would certainly be an improvement. I think the `calc_gwcs_pixmap` in alignment.utils is unused (romancal has it's [own copy](https://github.com/spacetelescope/romancal/blob/459e8e345bbb1ba6bfd76d0d8a90f2058089791e/romancal/resample/resample_utils.py#L205)).

I guess this PR is a step in the wrong direction by adding another one but perhaps it's within scope of this PR to also delete the one in alignment utils or do you think that would make more sense in a separate PR?

src/stcal/outlier_detection/utils.py
braingram: See: https://github.com/spacetelescope/stcal/pull/270#discussion_r1674651695

src/stcal/outlier_detection/utils.py
braingram: Good catch! I removed the mentioned bugs but not the comment. I removed this in: f32eef6

src/stcal/outlier_detection/utils.py
braingram: I applied this fix in: 30355b0

src/stcal/outlier_detection/utils.py
braingram: Good suggestion.
I don't see a message for the copied jwst code:
https://github.com/spacetelescope/jwst/blob/bc255f4cffa7a887ab3d1eed7ebfe82b856a4c67/jwst/resample/resample_utils.py#L158-L159
and running this (with the test data) doesn't generate any warnings. @nden any tips on how to trigger a warning here to see the message?

braingram: I'm wondering if we could just remove the warning filtering. It was added in:
https://github.com/spacetelescope/jwst/pull/3785
and I can't seem to reproduce those issues (I did not try running the old jwst code, just setting up transforms with nans and performing comparisons against nans). I wonder if numpy no longer emits these warnings (I tried back to numpy 1.21 and I don't get a warning using `nan` in a less operation).

src/stcal/outlier_detection/utils.py
emolter: It would make sense to me to have a separate PR focused specifically on evaluating the use of all these utils here an in both jwst and romancal, and figure out if they should be moved to gwcs, consolidated and remain here, or removed entirely.  This could probably happen after this PR and my tweakreg PR are both merged, since both of them add more utils

src/stcal/outlier_detection/utils.py
emolter: Thanks for trying these things. I agree it can just be removed if you aren't able to reproduce it.

emolter (approved):
All my comments have been addressed, looks good to me

src/stcal/outlier_detection/utils.py
braingram: Thanks! I opened a issue for this: https://github.com/spacetelescope/stcal/issues/271

mcara (approved):
LGTM

tox.ini
```diff
@@ -60,10 +60,14 @@ deps =
     oldestdeps: minimum_dependencies
     devdeps: numpy>=0.0.dev0
     devdeps: scipy>=0.0.dev0
+    devdeps: scikit-image>=0.0.dev0
     devdeps: pyerfa>=0.0.dev0
     devdeps: astropy>=0.0.dev0
-    devdeps: tweakreg>=0.0.dev0
     devdeps: requests>=0.0.dev0
+    devdeps: tweakreg>=0.0.dev0
```
emolter, line 67:
```suggestion
    devdeps: tweakwcs>=0.0.dev0
```

I am sorry!! I typo'd this and now it's merged

emolter:
looks like mypy needs an override in pyproject.toml for drizzle

src/stcal/outlier_detection/utils.py
```diff
@@ -0,0 +1,333 @@
...
+    # Sigma-clip the unmasked data
+    weight_masked = sigma_clip(weight_masked, sigma=3, maxiters=5)
+    mean_weight = np.mean(weight_masked)
+    # Mask pixels where weight falls below maskpt percent
+    weight_threshold = mean_weight * maskpt
+    return weight_threshold
+
+
+def _abs_deriv(array):
+    tmp = np.zeros(array.shape, dtype=np.float64)
```
kmacdonald-stsci, line 88: Missing docstring.

src/stcal/outlier_detection/utils.py
```diff
@@ -0,0 +1,333 @@
...
+    tmp[:, 1:] = array[:, :-1]
+    tmp, out = _absolute_subtract(array, tmp, out)
+    tmp[:, :-1] = array[:, 1:]
+    tmp, out = _absolute_subtract(array, tmp, out)
+
+    return out
+
+
+def _absolute_subtract(array, tmp, out):
+    tmp = np.abs(array - tmp)
```
kmacdonald-stsci, line 105: Missing docstring.

kmacdonald-stsci (approved):

src/stcal/outlier_detection/utils.py
braingram: I added short docstrings in [2112773](https://github.com/spacetelescope/stcal/pull/270/commits/2112773e603fac985c28f603eb6cc1bdd168adf3)

This function (and `_absolute_subtract`) should likely not be used since they appear to introduce erroneous cr flags due to treating off-edge pixels as 0. However, fixing this bug would result in changes to regression test results and I think makes more sense in a separate PR (where the new results can be verified). There is a JP ticket for this https://github.com/spacetelescope/jwst/issues/8636 and candidate new code here https://github.com/spacetelescope/jwst/pull/8635/files

For the scope of this PR I added "Do not use this function" to the docstrings as moving this code to stcal did not introduce the bug.

src/stcal/outlier_detection/utils.py
braingram: I added a brief docstring to this function (see https://github.com/spacetelescope/stcal/pull/270#discussion_r1693048629 for more details).

I attempted to add more but given the number of lines of code and the array "switch-a-roo" that this function is performing I think reading the code is the best bet for trying to understand what's happening.

Also this function will be removed if the candidate code for fixing https://github.com/spacetelescope/jwst/issues/8636 is acceptable.

PR #275, opened 2024-07-26
JP-3669: Updating the C Extension to do CHARGELOSS Read Noise Recalculations

kmacdonald-stsci:
Resolves [JP-3669](https://jira.stsci.edu/browse/JP-3669)

Closes #8601

This PR addresses the read noise variance recalculation based on the CHARGELOSS flag.  This recalculation was in the JWST ramp fit step code, but that came with a performance hit and was unnecessarily complicated.  The recalculation has been implemented in the C extension code, simplifying the ramp fitting, as well as improving performance.

There is a corresponding JWST PR that removes the CHARGELOSS recalculation from the step code, as well as removes the corresponding test. https://github.com/spacetelescope/jwst/pull/8678

The regression tests pass, except for one test.  The test failure has been confirmed to be incorrect.  The regression test truth file for this test needs to be updated:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1623/

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,12 @@ General
 
 - Add TweakReg submodule. [#267]
 
+ramp_fitting
+~~~~~~~~~~~~
+
+- Move the CHARGELOSS read noise variance recalculation from the JWST step
+  code to the C extension to simply the code and improve performance.[#275]
+
 Changes to API
 --------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,7 @@ ramp_fitting
 ~~~~~~~~~~~~
 
 - Move the CHARGELOSS read noise variance recalculation from the JWST step
-  code to the C extension to simply the code and improve performance.[#275]
+  code to the C extension to simplify the code and improve performance.[#275]
 
 Changes to API
 --------------
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -30,7 +30,7 @@ log.setLevel(logging.DEBUG)
 BUFSIZE = 1024 * 300000  # 300Mb cache size for data section
 
 
-def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
+def create_ramp_fit_class(model, algorithm, dqflags=None, suppress_one_group=False):
     """
     Create an internal ramp fit class from a data model.
 
@@ -59,10 +59,11 @@ def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
         dark_current_array = model.average_dark_current
 
     orig_gdq = None
-    wh_chargeloss = np.where(np.bitwise_and(model.groupdq.astype(np.uint32), dqflags['CHARGELOSS']))
-    if len(wh_chargeloss[0]) > 0:
-        orig_gdq = model.groupdq.copy()
-    del wh_chargeloss
+    if algorithm.upper() == "OLS_C":
+        wh_chargeloss = np.where(np.bitwise_and(model.groupdq.astype(np.uint32), dqflags['CHARGELOSS']))
+        if len(wh_chargeloss[0]) > 0:
+            orig_gdq = model.groupdq.copy()
+        del wh_chargeloss
 
     if isinstance(model.data, u.Quantity):
         ramp_data.set_arrays(model.data.value, model.err.value, model.groupdq,
@@ -182,7 +183,7 @@ def ramp_fit(
     # Create an instance of the internal ramp class, using only values needed
     # for ramp fitting from the to remove further ramp fitting dependence on
     # data models.
-    ramp_data = create_ramp_fit_class(model, dqflags, suppress_one_group)
+    ramp_data = create_ramp_fit_class(model, algorithm, dqflags, suppress_one_group)
 
     if algorithm.upper() == "OLS_C":
         ramp_data.run_c_code = True
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -91,6 +91,7 @@ def create_ramp_fit_class(model, algorithm, dqflags=None, suppress_one_group=Fal
     if "zero_frame" in model.meta.exposure and model.meta.exposure.zero_frame:
         ramp_data.zeroframe = model.zeroframe
 
+    ramp_data.algorithm = algorithm
     ramp_data.set_dqflags(dqflags)
     ramp_data.start_row = 0
     ramp_data.num_rows = ramp_data.data.shape[2]
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -58,11 +58,22 @@ def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
     else:
         dark_current_array = model.average_dark_current
 
+    orig_gdq = None
+    wh_chargeloss = np.where(np.bitwise_and(model.groupdq.astype(np.uint32), dqflags.group['CHARGELOSS']))
+    if len(wh_chargeloss[0]) > 0:
+        orig_gdq = model.groupdq.copy()
+
     if isinstance(model.data, u.Quantity):
         ramp_data.set_arrays(model.data.value, model.err.value, model.groupdq,
                              model.pixeldq, dark_current_array)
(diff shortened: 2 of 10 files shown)


CHANGES.rst
```diff
@@ -6,6 +6,12 @@ General
 
 - Add TweakReg submodule. [#267]
 
+ramp_fitting
+~~~~~~~~~~~~
+
+- Move the CHARGELOSS read noise variance recalculation from the JWST step
+  code to the C extension to simply the code and improve performance.[#275]
```
melanieclarke, line 13:
```suggestion
  code to the C extension to simplify the code and improve performance. [#275]
```

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -58,11 +58,23 @@ def create_ramp_fit_class(model, dqflags=None, suppress_one_group=False):
     else:
         dark_current_array = model.average_dark_current
 
+    orig_gdq = None
+    wh_chargeloss = np.where(np.bitwise_and(model.groupdq.astype(np.uint32), dqflags['CHARGELOSS']))
```
melanieclarke, line 62: I have noticed in testing the previous version of ramp fit that this `where` call takes a ton of memory to complete, maybe because of the forced casting to np.uint32.  Since this is only used to decide whether to copy the groupdq here, can we find a less expensive substitute?  And maybe only call it if algorithm is OLS_C?

src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -41,13 +45,15 @@ def __init__(self):
...
+        self.run_chargeloss = True
+        # self.run_chargeloss = False
 
         self.one_groups_locs = None  # One good group locations.
         self.one_groups_time = None  # Time to use for one good group ramps.
 
         self.current_integ = -1
 
-    def set_arrays(self, data, err, groupdq, pixeldq, average_dark_current):
+    def set_arrays(self, data, err, groupdq, pixeldq, average_dark_current, orig_gdq=None):
```
melanieclarke, line 56: Add orig_gdq to docstring

melanieclarke (commented):
I'm not great at C, but this looks plausible to me.  Just a couple comments on the Python parts and some typos.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I can research this.  Maybe `np.any` can be used in conjunction with `np.bitwise_and` to be faster than `np.where`.  Also, I will see if this is necessary.  The copy is done because the arrays get modified in during ramp fitting.  However, the way the C-extension is written, the code operates on local copies, possibly making this copy unnecessary.  In the mean time I conditioned it on running the `OLS_C` algorithm.

drlaw1558 (approved):
Linked to https://github.com/spacetelescope/jwst/pull/8697

Looks good to me!  I've successfully tested that:

1) Runtime improves as expected in test case jw02079004003_03101_00001_nis.  In the old version it takes about 2.5 minutes to run the relevant piece of the RN recalculation, while with this PR it takes about 2.5 seconds.

2) RN variances stay as expected in test case jw01510001001_02108_00001_nis using both the OLS and OLS_C algorithms.

3) Checked that the SCI and variance arrays did not change for other instruments that don't use charge migration.

CHANGES.rst
```diff
@@ -6,6 +6,12 @@ General
 
 - Add TweakReg submodule. [#267]
 
+ramp_fitting
+~~~~~~~~~~~~
+
+- Move the CHARGELOSS read noise variance recalculation from the JWST step
+  code to the C extension to simplify the code and improve performance.[#275]
+
```
zacharyburnett, line 14:
```suggestion
ramp_fitting
~~~~~~~~~~~~

- Move the CHARGELOSS read noise variance recalculation from the JWST step
  code to the C extension to simplify the code and improve performance.[#275]

```
#284 switched change log handling to `towncrier`, so this change log entry should be a file in `changes/`:
```shell
echo "[ramp_fitting] Move the CHARGELOSS read noise variance recalculation from the JWST step code to the C extension to simplify the code and improve performance." > changes/275.general.rst
```

(new PRs will have these instructions in the checklist)

melanieclarke (approved):
All my comments were addressed, thanks!

I think this should be good to go, once the changelog format is fixed and conflicts are resolved.  It might be a good idea to run the regression tests one last time before merging.

The corresponding JWST PR can be merged as soon as this one is in.

tapastro:
This needs an update to move the changelog entry to a towncrier fragment, then we'll get this merged ASAP in preparation for an stcal build.

tapastro:
Tests are showing a number of failures. It looks like the gls_fit tests have out of date dqflags, and the downstream tests show the same failures wrt. level 2 pipelines across PRs. We can get this PR in, but we should clean up these failures. There isn't much value gained by having tests if we ignore them due to repeated failures.

kmacdonald-stsci:
> Tests are showing a number of failures. It looks like the gls_fit tests have out of date dqflags, and the downstream tests show the same failures wrt. level 2 pipelines across PRs. We can get this PR in, but we should clean up these failures. There isn't much value gained by having tests if we ignore them due to repeated failures.

I added the CHARGELOSS flag keyword to the GLS flags for testing.  They now pass.

schlafly:
Is there an option to turn on/off this behavior?  For Roman we don't currently have this flag but we could add it; it does seem to me that we will have charge migration issues.  On the other hand, this code isn't about flagging or addressing the charge migration per se.  My understanding is that the "old" behavior gives a more accurate estimation of the read noise contribution to the variance of these pixels, but that we've decided that we don't want that more accurate estimation because it leads to high flux pixels being weighted downward in resample for IVM weighting.  I might have approached this via something like:
https://github.com/spacetelescope/jwst/pull/7563
where in resample one replaces the weights of saturated (or now also chargeloss'd) pixels with their unaffected neighbors, reasoning that one doesn't "really" care about the read noise variance for these pixels, where the poisson noise is dominant.

src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -131,6 +144,7 @@ def set_dqflags(self, dqflags):
         self.flags_saturated = dqflags["SATURATED"]
         self.flags_no_gain_val = dqflags["NO_GAIN_VALUE"]
         self.flags_unreliable_slope = dqflags["UNRELIABLE_SLOPE"]
+        self.flags_chargeloss = dqflags["CHARGELOSS"]
```
braingram, line 147: Is `CHARGELOSS` only supported by `OLS_C` and not `OLS`? If so, I think (and I tested this locally) passing in `algorithm` to this method and only setting `flags_chargeloss` if `algorithm == OLS_C` should fix the romancal tests.

kmacdonald-stsci: The `CHARGELOSS` flag is used when selecting either `OLS_C` or `OLS`.  Currently, the use of the `CHARGELOSS` flag is handled in the `JWST` step code.  Unfortunately, there is a bug in the segmentation computation (that occurs very rarely on an edge case) and with a computation hit.  

When running the `OLS` code, there is a message that the code returned from the `STCAL` code, but there can be a noticeable lag between that and the completion of the ramp fitting step.  To fix the bug and to improve computational performance, I moved the read noise variance re-calculation from the `JWST` step code into the C-extension.  This required some accounting changes in `ramp_fit_class.py` and `ramp_fit.py` to get the proper information to the C-extension.

nden:
I got confirmation from RTB that they do not intend to use OLS_C or OLS for Roman. 
If necessary, changes related to Roman will be done later. This can be merged  for JWST.

kmacdonald-stsci:
> Happy to merge this ASAP, but it looks like one of the new unit tests is failing:
> 
> ```
> FAILED tests/test_ramp_fitting.py::test_cext_chargeloss - assert np.float32(0.000556237) != np.float32(0.000556237)
> ```

I"ll check it out.

kmacdonald-stsci:
> Happy to merge this ASAP, but it looks like one of the new unit tests is failing:
> 
> ```
> FAILED tests/test_ramp_fitting.py::test_cext_chargeloss - assert np.float32(0.000556237) != np.float32(0.000556237)
> ```

It's fixed.

tapastro (approved):
LGTM - merging this momentarily.

PR #279, opened 2024-08-08
Add common resample code to stcal

mcara:
This PR add the common resample code used by both JWST and Roman pipelines to stcal. Also, for the first time, this PR adopts the new `drizzle` API from https://github.com/spacetelescope/drizzle/pull/134 for the resample code used in the pipelines.

This work is related to https://jira.stsci.edu/browse/AL-835

At this moment this is a very rough draft for illustration purpose. It should run with default arguments (except input_models and output file name can be specified; everything else is not guaranteed to work). There are no unit/regression tests and documentation may not match the code.

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)


src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1122 @@
...
+            result = exptime * dqmask
+
+        else:
+            result = np.ones(model.data.shape, dtype=model.data.dtype) * dqmask
+
+        return result.astype(np.float32)
+
+    @abc.abstractmethod
+    def run(self):
+        ...
```
kmacdonald-stsci, line 483: Instead of using ellipses, would it be better to raise and exception with a message saying "method not implemented"?

src/stcal/resample/utils.py
kmacdonald-stsci, line 1: Is this file necessary?  It is imported by only one other file.  It is not a set of common functions shared among many files.  It's two short functions.

kmacdonald-stsci (approved):

braingram:
Would you rebase this and add a drizzle dependency with the branch for https://github.com/spacetelescope/drizzle/pull/134 so we can see the CI run?

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+from stdatamodels.dqflags import interpret_bit_flags
```
braingram, line 2:
```suggestion
from stdatamodels.dqflags import interpret_bit_flags
```
Is there a non-stdatamodels way to handle this. We can't make stdatamodels a dependency of this code.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+            {
+                "weight_type": self.weight_type,
+                "pointings": len(self._input_models.group_names),
+            }
+        )
+        # TODO: also store the number of images added in total: ncoadds?
+
+        self.final_post_processing()
+
+        self._output_model.write(self._output_filename, overwrite=True)
```
braingram, line 785: This won't work for roman since `write` does not exist for `roman_datamodels.DataModel` and `overwrite` is not a valid keyword argument.

I think it's best if we leave all file-IO up to the pipeline and not include it here in stcal.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+    def new_model(self, image_shape=None, file_name=None, copy_meta_from=None):
+        """ Return a new model for the resampled output """
+        ...
+
+
+class OutputTooLargeError(RuntimeError):
+    """Raised when the output is too large for in-memory instantiation"""
+
+
+def output_wcs_from_input_wcs(input_wcs_list, pixel_scale_ratio=1.0,
```
emolter, line 149: I might put this somewhere other than `resample.py`, e.g. alignment/utils or alignment/resample_utils.  I would vote (eventually) for having a separate submodule for WCS-related utility functions, but that's probably beyond the PR scope

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+
+    def preload_input_meta(self, wcs1, filename, s_region):
+        # set-up lists for WCS and file names
+        self._input_img1_wcs = None
+        self._input_s_region_list = []
+        self._input_filename_list = []
+
+        # loop over only science exposures in the ModelLibrary
+        # sci_indices = self._input_models.ind_asn_type("science")
+        with self._input_models:
```
emolter, line 427: We seem to be in the same infinite loop as before here, about whether or not to make certain datamodels dependencies of stcal, either implicitly or explicitly.  It seems to me that with this and other code lines, ResampleBase can only be used if the input models are assumed to be ModelLibrary instances.  Are we okay with this?  If so, should we just make this explicit in some way by specifying the input model type inside the `ResampleBase` abstract base class, and make stpipe a dependency?

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+            for model in self._input_models:
+                # model = self._input_models.borrow(idx)
+
+                try:
+                    if self.get_model_attr_value(model, "exptype").upper() != "SCIENCE":
+                        self._input_models.shelve(model, modify=False)
```
emolter, line 433: usually when there's a `try...except` clause inside a ModelLibrary context where the `try` could fail while models are borrowed from the library, it's necessary to put the `shelve` into a `finally` block so it executes regardless of the path the `try...except` takes

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+    def check_memory_requirements(self, allowed_memory):
+        """ Called just before '_pre_run_callback()' is called to verify
+        that there is enough memory to hold the output. """
+        if allowed_memory is None and "DMODEL_ALLOWED_MEMORY" not in os.environ:
+            return
+
+        allowed_memory = float(allowed_memory)
+
+        # get the available memory
+        available_memory = psutil.virtual_memory().available + psutil.swap_memory().total
```
emolter, line 389: if `psutil` is a dependency here, why is `bytes2human` repeated here, instead of being inherited from `psutil`?

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+         weight type, exposure time (if relevant), and kernel, and merges
+         them with any user-provided values.
+      2. Creates output WCS based on input images and define mapping function
+         between all input arrays and the output array.
+      3. Updates output data model with output arrays from drizzle, including
+         a record of metadata from all input models.
+    """
+    resample_suffix = 'i2d'
+    resample_file_ext = '.fits'
+    n_arrays_per_output = 2  # #flt-point arrays in the output (data, weight, var, err, etc.)
```
emolter, line 181: why is this hard-coded?  Will it be different for co-add and single resamples?  Don't we have multiple types of variance that need to be in memory simultaneously in the current version of resample, i.e., doesn't this underestimate the memory usage when it's used in `check_memory_requirements`?

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+                self._input_models.shelve(model, modify=False)
+
+        # store first model's entire meta (except for WCS and data):
+        if self._first_model_meta is None:
+            self._first_model_meta = self.new_model(copy_meta_from=model)
+
+    def build_driz_weight(self, model, weight_type=None, good_bits=None):
+        """Create a weight map for use by drizzle
+        """
+        data = self.get_model_array(model, "data")
```
emolter, line 467: again here certain attributes of the model are assumed to exist, and IMO this should be made explicit either by having a model base class or by adding stcal dependencies (although we've been trying hard to avoid the latter)

src/stcal/resample/resample.py
emolter: by marking it as an `abstractmethod` this is done automatically if you try to call it directly

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1506 @@
...
+        return self.create_output_model(driz_data)
+
+
+class ResampleSingle(ResampleBase):
+    """
+    This is the controlling routine for the resampling process.
+
+    Notes
+    -----
+    This routine performs the following operations::
```
emolter, line 1164: this looks identical to ResampleCoAdd.  Some doc updates should be made to clarify how they are different

emolter (commented):
my main question is still the extent to which we assume the structure and attributes of datamodels, and now also of ModelLibrary

src/stcal/resample/resample.py
mcara: that was an omission. I'll fix this.

src/stcal/resample/utils.py
mcara: yes, there is. I'll fix it (I forgot to switch to `astropy` although `stdatamodels.dqflags.interpret_bit_flags()` for unknown to me reasons does things slightly different. I think _that_ function should be removed altogether from `stdatamodels.dqflags`.

mcara: oh, I did it in `resample.py` but not in utils.

src/stcal/resample/resample.py
mcara: > We seem to be in the same infinite loop as before here, about whether or not to make certain datamodels dependencies of stcal, either implicitly or explicitly. It seems to me that with this and other code lines, ResampleBase can only be used if the input models are assumed to be ModelLibrary instances.

True. It is difficult to maximize the common code ported to `stcal` with the purpose of reducing maintenance burden in the pipelines given the rigid usage imposed by the `ModelLibrary` (by contrast, `ModelContainer` was behaving like a standard list). I can try to see how to bury `ModelLibrary` into the "IO" class, but it's not going to be pretty.

> Are we okay with this? If so, should we just make this explicit in some way by specifying the input model type inside the `ResampleBase` abstract base class, and make stpipe a dependency?

I don't know what are the plans for these packages but I do not think they are intended to be general purpose algorithms like those in `astropy` (or even `drizzle`) so I do not see an issue with using structures shared by all pipelines.

src/stcal/resample/resample.py
mcara: because `psutil` does not have this function.

src/stcal/resample/resample.py
mcara: Agree. It was temporary here. This will be fixed once we have a function that computes output WCS from `s_region` of input data models and other parameters.

src/stcal/resample/resample.py
emolter: whoops, sorry

src/stcal/resample/resample.py
mcara: actually, if an exception is triggered then there is some other processing after the `try-except` block and then the model is closed. So, I think, the code is fine: this was intentional.

src/stcal/resample/resample.py
mcara: > again here certain attributes of the model are assumed to exist, and IMO this should be made explicit either by having a model base class or ... 

Seems like this is not going to happen. 

> ... by adding stcal dependencies (although we've been trying hard to avoid the latter)

I don't think this is necessary as this PR illustrates how to avoid imports.

Yes, it is true thst we assume some attributes to exist. I think for now we are safe but if you would like to harden the code, I could add `try-except` blocks around these data retrieval statements.

perrygreenfield (commented):
Minor comments so far.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1310 @@
...
+
+        """
+        if allowed_memory is None and "DMODEL_ALLOWED_MEMORY" not in os.environ:
+            return
+
+        allowed_memory = float(allowed_memory)
+
+        # get the available memory
+        available_memory = (
+            psutil.virtual_memory().available + psutil.swap_memory().total
```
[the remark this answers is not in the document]
emolter: I think this is ported over from current jwst main branch, but I don't think it does anything useful at present.  See https://github.com/spacetelescope/jwst/issues/8775

I wouldn't mind if it were removed as part of this PR, but the JIRA ticket is still assigned to David Law which means it hasn't been officially approved for work by the pipeline team yet

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,125 @@
...
+        tmeasure = model["measurement_time"]
+    except KeyError:
+        return model["exposure_time"], False
+    if tmeasure is None:
+        return model["exposure_time"], False
+    else:
+        return tmeasure, True
+
+
+# FIXME: temporarily copied here to avoid this import:
```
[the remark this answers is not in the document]
mcara: yes. I think that would be the right place.

PR #281, opened 2024-08-17
JP-3698: Fixed Memory Leak in Ramp Fitting C-Extension

### LINKED ISSUE #276: Memory leak in `OLS_C` ramp fitting.
xref for the jwst issue: https://github.com/spacetelescope/jwst/issues/8668

Repeated runs of `Detector1` when run with `OLS_C` show a steady increase in memory usage.
Screenshot 2024-07-28 at 11 40 02 AM
The increase goes away if the algorithm is switched to `OLS`:
Screenshot 2024-07-28 at 11 31 57 AM

The above runs are described in more detail in the linked jwst issue.

kmacdonald-stsci:
…eturned too early, so never decremented reference counts for python objects.

Resolves [JP-3698](https://jira.stsci.edu/browse/JP-3698)

Closes [jwst/#8680](https://github.com/spacetelescope/jwst/issues/8680)
Closes #276 

This PR addresses a memory leak found in the ramp fitting C-extension.  Added decrement reference counters to objects referenced that were passed from the python code.  These missing decrements meant that reference counts were not properly garbage collected.

Attached is a screen shot of the memory usage from `memray` demonstrating there is no memory leak.

jp_3698_memory_flat

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stcal/ramp_fitting/src/slope_fitter.c
+++ b/src/stcal/ramp_fitting/src/slope_fitter.c
@@ -3,9 +3,13 @@
 
 #include <stdlib.h>
 #include <math.h>
-#include <stdio.h>
 #include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+#include <sys/resource.h>
+#include <time.h>
+#include <unistd.h>
 
 #include <numpy/arrayobject.h>
 #include <numpy/npy_math.h>
@@ -705,6 +709,30 @@ is_pix_in_list(struct pixel_ramp * pr)
     return 0;
 }
 
+static inline long long
+print_pid_info(long long prev, int line, char * label) {
+    struct rusage res_usage;
+    long long now_time = (long long)time(NULL);
+    long long mem_usage = -1;
+    long long diff = 0;
+    pid_t pid = getpid();
+    // dbg_ols_print("PID:  %d\n", pid);
+
+    getrusage(RUSAGE_SELF, &res_usage);
+    mem_usage =  res_usage.ru_maxrss;
+    if (prev > 0) {
+        diff = mem_usage - prev;
+        dbg_ols_print("[%d] time: %lld, Mem: %lld, diff: %lld, prev: %lld, pid: %d  '%s'\n",
+            line, now_time, mem_usage, diff, prev, pid, label);
+    } else {
+        dbg_ols_print("[%d] time: %lld, Mem: %lld, diff: %lld, pid: %d  '%s'\n",
+            line, now_time, mem_usage, diff, pid, label);
+    }
+
+    return mem_usage;
+}
+
+
 /* ------------------------------------------------------------------------- */
 /*                            Module Functions                                */
 /* ------------------------------------------------------------------------- */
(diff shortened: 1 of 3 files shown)


src/stcal/ramp_fitting/src/slope_fitter.c
```diff
@@ -899,23 +927,29 @@ clean_ramp_data(
...
-        if (current) {
-            next = current->flink;
-            SET_FREE(current);
-            current = next;
+    Py_XDECREF(rd->data);
+    Py_XDECREF(rd->groupdq);
+    Py_XDECREF(rd->err);
+    Py_XDECREF(rd->pixeldq);
+    Py_XDECREF(rd->zframe);
+    Py_XDECREF(rd->dcurrent);
```
braingram, line 935: Should this contain `Py_XDECREF`s for `gain` and `rnoise`?

src/stcal/ramp_fitting/src/slope_fitter.c
```diff
@@ -1647,6 +1690,8 @@ get_ramp_data(
         }
     }
 
+END:
+    // Py_XDECREF(Py_ramp_data); /* XXX Probably should be here */
```
braingram, line 1694: Is this `Py_XDECREF` needed only if `get_ramp_data_parse` succeeds?

drlaw1558 (approved):
I can't speak to the details of the C changes, but empirically it looks like this has fixed the memory leak in my tests as well.  Previously I saw the memory used grow by about 6-7 GB over the course of processing about 25 files from PID 1523, and with this branch that number is more like 1 GB (which I'd consider within uncertainty given how I've been measuring usage).

src/stcal/ramp_fitting/src/slope_fitter.c
kmacdonald-stsci: This was part of the additional testing I wanted to do.  These variables get set at:

https://github.com/kmacdonald-stsci/stcal/blob/2df21c4fb36e1d302717771fbc4b8878781fc31a/src/stcal/ramp_fitting/src/slope_fitter.c#L1772

This doesn't appear to increment the reference count.  If I include `Py_XDECREF` for `gain` and `rnoise`, this causes problems when returning from ramp fitting.  When these variables are used during the read noise recalculation for CHARGELOSS those objects aren't available anymore.

braingram:
Thanks for the responses.

Changes look good to me. Is there a link to jwst regression tests with this PR?

Do you think it's worth adding a unit test to check for the re-introduction of this problem? Running main the extra references fixed in this PR can be seen (with `sys.getrefcount`) for `data`, `err`, `pixeldq` `groupdq` and `averagedarkcurrent` after a call to `ramp_fit_data`. I looked for a close unit test that might be modifiable but I'm not seeing any that use `OLS_C`. Do the unit tests only cover the python version?

kmacdonald-stsci:
> Thanks for the responses.
> 
> Changes look good to me. Is there a link to jwst regression tests with this PR?
> 
> Do you think it's worth adding a unit test to check for the re-introduction of this problem? Running main the extra references fixed in this PR can be seen (with `sys.getrefcount`) for `data`, `err`, `pixeldq` `groupdq` and `averagedarkcurrent` after a call to `ramp_fit_data`. I looked for a close unit test that might be modifiable but I'm not seeing any that use `OLS_C`. Do the unit tests only cover the python version?

I'll think of a unit test that can check the reference counts.  I am working on a PR now to change all unit tests to test `OLS_C`, since it is now the default.

braingram (approved):
Thanks for finding and fixing the bug!

kmacdonald-stsci:
> Thanks for the responses.
> 
> Changes look good to me. Is there a link to jwst regression tests with this PR?
> 
> Do you think it's worth adding a unit test to check for the re-introduction of this problem? Running main the extra references fixed in this PR can be seen (with `sys.getrefcount`) for `data`, `err`, `pixeldq` `groupdq` and `averagedarkcurrent` after a call to `ramp_fit_data`. I looked for a close unit test that might be modifiable but I'm not seeing any that use `OLS_C`. Do the unit tests only cover the python version?

I developed a unit test for this.  Take a look.  It creates a basic `RampData` class, along with some other necessary objects, then runs only the C-extension, examining the reference count before and after the call for objects of interest, then asserts they must be the same.

braingram:
> > Thanks for the responses.
> > Changes look good to me. Is there a link to jwst regression tests with this PR?
> > Do you think it's worth adding a unit test to check for the re-introduction of this problem? Running main the extra references fixed in this PR can be seen (with `sys.getrefcount`) for `data`, `err`, `pixeldq` `groupdq` and `averagedarkcurrent` after a call to `ramp_fit_data`. I looked for a close unit test that might be modifiable but I'm not seeing any that use `OLS_C`. Do the unit tests only cover the python version?
> 
> I developed a unit test for this. Take a look. It creates a basic `RampData` class, along with some other necessary objects, then runs only the C-extension, examining the reference count before and after the call for objects of interest, then asserts they must be the same.

Thanks! Does the unit test use `OLS_C`? I'm not seeing where it overwrites the python default.

tapastro (approved):
LGTM. Ran regression tests just to ensure no unforeseen bugs, which passed (all failures known to be from other issues): https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1669/

@nden do we need to request review from a Roman rep to get this merged?

PR #282, opened 2024-08-21
JP-3664: Numpy 2.0 incompatibility with existing use of newbyteorder

tapastro:
Resolves [JP-3664](https://jira.stsci.edu/browse/JP-3664)

Partially closes [jwst/#8580](https://github.com/spacetelescope/jwst/issues/8580)

This PR addresses a change in behavior for numpy 2.0: 

> arr.newbyteorder('S').byteswap(inplace=True)
> \>\>\> AttributeError: `newbyteorder` was removed from the ndarray class in NumPy 2.0. Use arr.view(arr.dtype.newbyteorder(order))' instead.

This PR implements the dtype method.

**Checklist**

- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/changes/282.bugfix.rst
+++ b/changes/282.bugfix.rst
@@ -1,0 +1,1 @@
+Implement byteorder swap method that is forward-compatible with numpy 2.0 in jwst ramp_fitting.
--- a/src/stcal/ramp_fitting/ols_fit.py
+++ b/src/stcal/ramp_fitting/ols_fit.py
@@ -731,7 +731,7 @@ def handle_array_endianness(arr, sys_order):
     arr_order = arr.dtype.byteorder
     bswap = False
     if (arr_order == ">" and sys_order == "<") or (arr_order == "<" and sys_order == ">"):
-        arr.newbyteorder('S').byteswap(inplace=True)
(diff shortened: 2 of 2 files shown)


kmacdonald-stsci (approved):
We already have many `STCAL` PR's needing approval and merging.  I prefer to wait on those being merged before this one is merged, just in case.  I'll approve this, but would like to wait on the merge.

I have not done any investigation into using Numpy 2.0 with ramp fitting, including the C-extension.  Is this the only change that is required or will a more thorough investigation be required?

CHANGES.rst
```diff
@@ -14,7 +14,8 @@ Changes to API
 Bug Fixes
 ---------
 
-- 
+- Implement byteorder swap method that is forward-compatible with
+  numpy 2.0 in jwst ramp_fitting. [#282]
```
zacharyburnett, line 18:
```suggestion
```
#284 switched change log handling to `towncrier`, so this change log entry should be a file in `changes/`:
```shell
echo "Implement byteorder swap method that is forward-compatible with numpy 2.0 in jwst ramp_fitting." > changes/282.bugfix.rst
```

(new PRs will have these instructions in the checklist)

braingram (approved):

PR #286, opened 2024-09-04
Add optional parameter to alignment methods and handle catalog web service connection issues.

mairanteodoro:
Resolves [RCAL-908](https://jira.stsci.edu/browse/RCAL-908)

This PR add an optional parameter -- `clip_accum` -- to the alignment methods in `stcal.tweakreg`. This is necessary because JWST and Roman do not use the same value when it comes to that parameter. The default value for that parameter in `align_wcs()` is False, which is what JWST uses. However, `clip_accum=True` seems to significantly improve the fitting results for Roman (as compared with `clip_accum=False`).

This PR also adds the ability to better handle catalog web services connectivity issues.

**Checklist**

- [ ] for a public change, added a towncrier news fragment in `changes/`: <details><summary>`echo "changed something" > changes/<PR#>.<changetype>.rst`</summary>

- [ ] updated relevant tests
- [ ] added relevant label(s)

### DIFF
--- a/src/stcal/tweakreg/astrometric_utils.py
+++ b/src/stcal/tweakreg/astrometric_utils.py
@@ -156,7 +156,8 @@ def get_catalog(
     declination,
     epoch=2016.0,
     search_radius=0.1,
-    catalog="GAIADR3"
+    catalog="GAIADR3",
+    timeout=TIMEOUT,
 ):
     """Extract catalog from VO web service.
 
@@ -179,6 +180,9 @@ def get_catalog(
     catalog : str, optional
         Name of catalog to query, as defined by web-service. Default: 'GAIADR3'
 
+    timeout : float, optional
+        Timeout in seconds to wait for the catalog web service to respond. Default: 30.0 s
+
     Returns
     -------
     csv : `~astropy.table.Table`
@@ -199,7 +203,18 @@ def get_catalog(
         catalog
     )
     service_url = f"{SERVICELOCATION}/{service_type}?{spec}"
-    rawcat = requests.get(service_url, headers=headers, timeout=TIMEOUT)
+    try:
+        rawcat = requests.get(service_url, headers=headers, timeout=timeout)
+    except requests.exceptions.ConnectionError:
+        raise requests.exceptions.ConnectionError(
+            "Could not connect to the VO API server. Try again later."
+        )
+    except requests.exceptions.Timeout:
+        raise requests.exceptions.Timeout("The request to the VO API server timed out.")
+    except requests.exceptions.RequestException:
+        raise requests.exceptions.RequestException(
+            "There was an unexpected error with the request."
+        )
     r_contents = rawcat.content.decode()  # convert from bytes to a String
     rstr = r_contents.split("\r\n")
     # remove initial line describing the number of sources returned
--- a/src/stcal/tweakreg/tweakreg.py
+++ b/src/stcal/tweakreg/tweakreg.py
@@ -46,6 +46,7 @@ def relative_align(correctors: list,
                    fitgeometry: str = "rshift",
                    nclip: int = 3,
                    sigma: float = 3.0,
+                   clip_accum: bool = False,
                    ) -> list:
 
     if separation <= _SQRT2 * tolerance:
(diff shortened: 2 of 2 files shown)


emolter (commented):
The changes themselves look good to me pending the jwst downstream tests.  Why did you add the no_changelog_entry_needed tag?  It seems to me it would be useful to know which PR added these changes just in case it does affect something unexpectedly

mairanteodoro:
> Why did you add the no_changelog_entry_needed tag?
I did that temporarily -- it's still a draft PR. I am working on the changelog entry now.

emolter (approved):
LGTM: downstream JWST failures match the most recent test on main

emolter (approved):
thanks for fixing the docs build, everything else is the same as before and approved as far as I'm concerned

changes/286.general.rst
```diff
@@ -0,0 +1,3 @@
+Add clip_accum parameter to alignment methods.
```
kmacdonald-stsci, line 1: I am not quite sure of the correct formatting for this, but I think there should be a `[tweakreg]` at the beginning of each line to ensure the change log update goes under the correct header.  You can verify with @zacharyburnett to make sure the formatting for these fragments.

kmacdonald-stsci (approved):

changes/286.general.rst
zacharyburnett: there does not *need* to be (`towncrier` only looks at the filename when determining the header), however, adding `[step]` makes it a bit clearer what it's doing

zacharyburnett: this is how it looks at the moment:
[image](https://stcal--286.org.readthedocs.build/en/286/stcal/changes.html)

changes/286.general.rst
```diff
@@ -0,0 +1,2 @@
+- [tweakreg] Add clip_accum parameter to alignment methods.
+- [tweakreg] Improve handling of catalog web service connectivity issues.
```
braingram:
```suggestion
Add clip_accum relative_align and timeout to get_catalog.
```

If the goal is to have changelog entries divided by submodule we can update `pyproject.toml` to list the sections. Otherwise I suggest we keep the changelog entries free of any complicated formatting to avoid issues when this is rendered by towncrier and by sphinx.

The `[...]` syntax won't be rendered as a subsection as seen for https://stcal--275.org.readthedocs.build/en/275/stcal/changes.html

src/stcal/tweakreg/astrometric_utils.py
```diff
@@ -199,7 +203,18 @@ def get_catalog(
         catalog
     )
     service_url = f"{SERVICELOCATION}/{service_type}?{spec}"
-    rawcat = requests.get(service_url, headers=headers, timeout=TIMEOUT)
+    try:
+        rawcat = requests.get(service_url, headers=headers, timeout=timeout)
+    except requests.exceptions.ConnectionError:
+        raise requests.exceptions.ConnectionError(
```
braingram, line 209: Why change the error messages?

mairanteodoro: We're raising a specific error for each known scenario (connectivity and timeout), and one for unknown reasons.

braingram: Thanks. Is this necessary?
I ask because in all 3 situations the exception from requests is being discarded losing the information about the original cause of the error. This discards potentially useful information. If it's not required I suggest keeping the exceptions as-is or chaining a new exception if something needs to be caught be romancal.

mairanteodoro: Should we use logging or just a regular print for the request error message?

braingram: On second thought. I think this is fine as-is. I happened to hit one of these exceptions in:
https://github.com/spacetelescope/romancal/actions/runs/10745125293/job/29803544145#step:10:922
and since the exceptions are implicitly chained we can just look back the chain to find the error information (like the url that failed).

braingram (approved):
LGTM! Thanks for fielding my last-minute questions.

PR #289, opened 2024-09-23
JP-3734: Making OLS_C Default for CI Tests

kmacdonald-stsci:
…ith CHARGELOSS.

Resolves [JP-3734](https://jira.stsci.edu/browse/JP-3734)

Closes #

The JWST default algorithm was changed to be OLS_C for ramp fitting.  The STCAL ramp fit testing has been changed to reflect this change in default algorithm.

## Tasks

</details>

### DIFF
--- a/src/stcal/ramp_fitting/ramp_fit_class.py
+++ b/src/stcal/ramp_fitting/ramp_fit_class.py
@@ -49,7 +49,6 @@ class RampData:
         # C code debugging switch.
         self.run_c_code = True
         self.run_chargeloss = False
-        # self.run_chargeloss = False
 
         self.one_groups_locs = None  # One good group locations.
         self.one_groups_time = None  # Time to use for one good group ramps.
--- a/src/stcal/ramp_fitting/ramp_fit_class.py
+++ b/src/stcal/ramp_fitting/ramp_fit_class.py
@@ -47,8 +47,8 @@ class RampData:
         self.suppress_one_group_ramps = False
 
         # C code debugging switch.
-        self.run_c_code = False
-        self.run_chargeloss = True
+        self.run_c_code = True
+        self.run_chargeloss = False
         # self.run_chargeloss = False
 
         self.one_groups_locs = None  # One good group locations.
--- a/changes/289.general.rst
+++ b/changes/289.general.rst
@@ -1,0 +1,1 @@
+[ramp_fitting] Changing the default algorithm for ramp fitting to be ``OLS_C`` for CI tests.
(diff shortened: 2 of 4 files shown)


src/stcal/ramp_fitting/ramp_fit_class.py
```diff
@@ -47,8 +47,8 @@ def __init__(self):
         self.suppress_one_group_ramps = False
 
         # C code debugging switch.
-        self.run_c_code = False
-        self.run_chargeloss = True
+        self.run_c_code = True
+        self.run_chargeloss = False
         # self.run_chargeloss = False
```
melanieclarke, line 52: Remove commented out code

melanieclarke (commented):
Looks fine to me!  One very small clean up request.

Did you run regression tests somewhere?

melanieclarke:
@kmacdonald-stsci - did regression tests get run for JWST for this change?

tapastro:
It looks like this is breaking a unit test:

`FAILED tests/test_ramp_fitting.py::test_neg_med_rates_single_integration_multi_segment_optional - AssertionError: 
  Not equal to tolerance rtol=1e-06, atol=0
  
  -inf location mismatch:
   ACTUAL: array([ 0.     , 60.53071,     -inf], dtype=float32)
   DESIRED: array([0., 0., 0.])
 `
 
 We should probably resolve this failure before merging.

kmacdonald-stsci:
> It looks like this is breaking a unit test:
> 
> ```
> FAILED tests/test_ramp_fitting.py::test_neg_med_rates_single_integration_multi_segment_optional - AssertionError: 
>   Not equal to tolerance rtol=1e-06, atol=0
>   
>   -inf location mismatch:
>    ACTUAL: array([ 0.     , 60.53071,     -inf], dtype=float32)
>    DESIRED: array([0., 0., 0.])
> ```
> 
> We should probably resolve this failure before merging.

It's not clear to me this is caused by this PR.  And I've never seen such a strange problem like this.  If I run `pytest tests/test_ramp_fitting.py` the output to be checked for this test is different that if I run `pytest tests/test_ramp_fitting.py::test_neg_med_rates_single_integration_multi_segment_optional`.  

It doesn't matter if I run with the `OLS` algorithm or the `OLS_C` algorithm.  The screenshots of the test outputs show that the returned array is all zero when running the whole test script, but is non-zero if only the specific test is run, regardless of the choice of algorithm.  Because this difference occurs independent of which algorithm is chosen, I don't think this PR can be the problem.

I'll continue to investigate.

 


Screenshot 2024-10-03 at 3 08 59 PM
Screenshot 2024-10-03 at 3 20 23 PM

PR #283, opened 2024-08-26
JP-3593: 2nd group saturation part 2

drlaw1558:
Ready for review: this addresses JP-3593, flagging cosmic ray saturation in group 2 data in a consistent manner between both IRS2 and non-IRS2 modes using the new use_readpatt keyword.  This PR is tied to JWST PR https://github.com/spacetelescope/jwst/pull/8731

**Checklist**

- [ ] added entry in `CHANGES.rst` (either in `Bug Fixes` or `Changes to API`)
- [ ] updated relevant tests
- [ ] updated relevant documentation
- [ ] updated relevant milestone(s)
- [ ] added relevant label(s)

PR #290, opened 2024-09-23
TEST: Revert "JP-3593: 2nd group saturation part 2"

braingram:
Reverts spacetelescope/stcal#283 to test romancal


braingram:
This confirmed the regtest differences were from this change. Closing.

PR #292, opened 2024-09-27
JP-3768: Move outlier detection median computers to stcal

emolter:
Resolves [JP-3768](https://jira.stsci.edu/browse/JP-3768)

This PR  ports the median calculation machinery added by https://github.com/spacetelescope/jwst/pull/8782 into stcal for use by other missions.  See https://github.com/spacetelescope/jwst/pull/8840 for the related changes in jwst.

## Tasks

</details>


emolter:
[jwst regtests started here](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1758/)

kmacdonald-stsci (approved):
There are some incomplete docstrings missing a brief description of functions/methods.  Other than that, looks fine.

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,355 @@
...
+
+_ONE_MB = 1 << 20
+
+
+def nanmedian3D(cube: np.ndarray, overwrite_input: bool = True) -> np.ndarray:
+    """Compute the median of a cube ignoring warnings and with
+    memory efficiency optimizations. np.nanmedian always uses at least 64-bit
+    precision internally, and this is too memory-intensive. Instead, loop over
+    the median calculation to avoid the memory usage of the internal upcasting
+    and temporary array allocation. The additional runtime of this loop is
```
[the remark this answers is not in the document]
emolter: If I understand the question you are asking about the runtime performance of the nanmedian3D function, right?  This conversation on the JWST PR may be relevant, although no test was done with the zeroth (time/n_groups) axis being as large as that.

When you say that it seriously slowed down runtime in ramp fitting, do you mean that a median calculation was done, or just that the entire step scaled poorly with the number of integrations?

For imaging data, where we've seen the slowest processing, the largest n_groups we have seen is only ~100, since the step takes _cal files as input instead of _calints.  However, I can look into whether there would be any coronagraphic data that had a huge number of integrations in their _calints files, as this is also used by jwst for coronagraphic data

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,355 @@
...
+        return output_arr
+
+
+class MedianComputer:
+    """
+    Top-level class to treat median computation uniformly, whether in
+    memory or on disk.
+    """
+
+    def __init__(self: MedianComputer,
```
braingram, line 48:
```suggestion
    def __init__(self: MedianComputer,
```
Is annotating `self` common practice for type hinting? I imagine it would make class renaming difficult and what would it do for a subclass?

emolter: I didn't think it was necessary either, but then mypy was complaining locally.  I'll check if this can be removed setting mypy up to take its configuration from the stcal pyproject.toml

braingram: Perhaps `from typing import Self`? (only works for python 3.11+ maybe that's enough for our CI checks).

emolter: I was wrong about the failures being from mypy, they were actually caused by our ruff check.  See the small change I made to the pyproject.toml: [ANN101](https://docs.astral.sh/ruff/rules/missing-type-self/) and [ANN102](https://docs.astral.sh/ruff/rules/missing-type-cls/) are specific to `self` and `cls` annotations, and both are deprecated

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,355 @@
...
+                msg = "Index must be provided when using in-memory median"
+                raise ValueError(msg)
+            # populate pre-allocated memory with the drizzled data
+            self._median_computer[idx] = data
+        else:
+            # distribute the drizzled data into the temporary storage
+            self._median_computer.add_image(data)
+
+    def evaluate(self: MedianComputer) -> np.ndarray:
+        """
```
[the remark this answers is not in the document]
emolter: Let me know if the updated docstring looks ok to you

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,355 @@
...
+            computer = OnDiskMedian(full_shape,
+                                    dtype=dtype,
+                                    buffer_size=buffer_size)
+        self._median_computer: Any = computer
+
+    def append(self: MedianComputer,
+               data: np.ndarray,
+               idx: int | None = None
+               ) -> None:
+        """
```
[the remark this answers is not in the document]
emolter: Let me know if the updated docstring looks ok to you

src/stcal/outlier_detection/median.py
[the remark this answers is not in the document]
emolter: I tried this on a relatively large coronagraphy dataset I had lying around from [JWSTDMS-921](https://jira.stsci.edu/browse/JWSTDMS-921).  The  shape of the array going into this function was (1250, 224, 288), so still not 6000 but this is on the large side of what is processed here in practice.  The median calculation does not blow up the runtime in this case: the entire outlier detection step took only 5 seconds to run, as compared with ~1 hour for the coronagraphy-specific `align_refs` and `klip` steps.  So I don't think this is a concern for runtime but let me know if you would like to see additional tests run.

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+        self._slice_shape = slice_shape
+        self._dtype = np.dtype(dtype)
+        self._append_count = 0
+
+    @property
+    def shape(self) -> tuple:
+        return (self._append_count, *self._slice_shape)
+
+    def append(self, data: np.ndarray) -> None:
+        """Add a new slice to the temporary file."""
```
braingram, line 180:
```suggestion
        """
        Add a new slice to the temporary file.
        
        Parameters
        ------------
        data
            Must have a shape and dtype matching the ``slice_shape``
            and ``dtype`` provided during ``__init__``.
        """
```

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+            buffer_size = imrows * imcols * self.itemsize
+        per_model_buffer_size = buffer_size / self._expected_nframes
+        min_buffer_size = imcols * self.itemsize
+        section_nrows = \
+            min(imrows, int(per_model_buffer_size // min_buffer_size))
+
+        if section_nrows <= 0:
+            buffer_size = min_buffer_size * self._expected_nframes
+            msg = "Buffer size is too small to hold a single row. "
+            msg += f"Increasing buffer size to {buffer_size / _ONE_MB} MB"
```
braingram, line 283:
```suggestion
            msg = (
                "Buffer size is too small to hold a single row. "
                f"Increasing buffer size to {buffer_size / _ONE_MB} MB"
            )
```
This is likely a nit pick but for multi-line strings this should pass the ruff rules and avoid the need to add strings (which results in re-allocations and copies). This comment applies to several places in this PR.

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+        msg = f"Computing median over {self._expected_nframes} "
+        msg += f"groups in {nsections} sections "
+        msg += f"with total memory buffer {buffer_size / _ONE_MB} MB"
+        log.info(msg)
+        return nsections, section_nrows
+
+    def _temparray_setup(self,
+                         dtype: str | np.dtype
+                         ) -> list[DiskAppendableArray]:
+        """Set up temp file handlers for each spatial section."""
```
braingram, line 298:
```suggestion
        """Set up temp file handlers, one for each section."""
```

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+                shp = (self.frame_shape[0] - (self.nsections-1) *
+                       self.section_nrows, self.frame_shape[1])
+            arr = DiskAppendableArray(shp, dtype, self._temp_path / f"{i}.bin")
+            temp_arrays.append(arr)
+        return temp_arrays
+
+    def add_image(self, data: np.ndarray) -> None:
+        """
+        Split resampled model data into spatial sections
+        and write to disk.
```
braingram, line 313:
```suggestion
        Split data into ``nsections`` and write/append each to the
        corresponding temporary file.
```

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+            msg = f"Data shape {data.shape} does not match expected shape "
+            msg += f"{self.frame_shape}"
+            raise ValueError(msg)
+        if data.dtype != self.dtype:
+            msg = f"Data dtype {data.dtype} does not match expected dtype "
+            msg += f"{self.dtype}"
+            raise ValueError(msg)
+
+    def cleanup(self) -> None:
+        """Remove the temporary files and directory when finished."""
```
braingram, line 339:
```suggestion
        """
        Remove the temporary files and directory when finished.
        
        This should be called when this instance is no longer needed
        and the instance should not be used after a call to this method.
        """
```

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+        self._temp_dir.cleanup()
+
+    def compute_median(self) -> np.ndarray:
+        """
+        Read spatial sections from disk and compute the median across groups
+        (median over number of exposures on a per-pixel basis).
```
braingram, line 345:
```suggestion
        Compute median of previously added data.
        
        Returns
        -------
        median
            2 dimensional array of shape [``imrows``, ``imcols``] containing
            median values computed across all images previously provided
            to ``add_image``.
```

tests/outlier_detection/test_median.py
```diff
@@ -0,0 +1,205 @@
...
+    with pytest.raises(IsADirectoryError):
+        DiskAppendableArray(slice_shape, "float3", tempdir)
+
+
+# filter warnings such that "ResourceWarning: Implictly cleaning up" is raised
+# unfortunately need to do this indirectly
+# by finding "finalize" string in PytestUnraisableException
+@pytest.mark.filterwarnings(
+        "error:.*finalize.*:pytest.PytestUnraisableExceptionWarning"
+        )
```
braingram, line 88:
```suggestion
```
I don't see this warning if I remove this filter. Is it still needed?

tests/outlier_detection/test_median.py
```diff
@@ -0,0 +1,205 @@
...
+    for i in range(full_shape[0]):
+        frame = np.full((21, 20), i, dtype=np.float32)
+        comp_memory.append(frame, i)
+        comp_disk.append(frame, i)
+    assert np.allclose(comp_memory.evaluate(), comp_disk.evaluate())
+
+
+@pytest.mark.filterwarnings(
+        "error:.*finalize.*:pytest.PytestUnraisableExceptionWarning"
+        )
```
braingram, line 167:
```suggestion
```
I don't see this warning if I remove this filter. Is it still needed?

braingram (commented):
Most comments are docs/docstring related.

Would you add `median.py` to the docs? It will likely be easier if the submodule has an `__all__` (I added a comment about this) which will also help to define the "public" API.

tests/outlier_detection/test_median.py
emolter: The filter ensures that the test fails if cleanup didn't occur properly.  Maybe instead I should make an expected failure related to this.  Will look into doing that

braingram: Thanks! I'd say remove it then. If the test fails I'd be ok seeing a second failure because of the warning. That seems preferable to ignoring the warning (which we do want to produce an error or at least a warning in case a future update to the code in this PR breaks `cleanup`).

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+    """
+    Top-level class to treat median computation uniformly, whether in
+    memory or on disk.
+    """
+
+    def __init__(self,
+                 full_shape: tuple,
+                 in_memory: bool,
+                 buffer_size: int | None = None,
+                 dtype: str | np.dtype = "float32"
```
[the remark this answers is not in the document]
emolter: good catch

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+    the median calculation to avoid the memory usage of the internal upcasting
+    and temporary array allocation. The additional runtime of this loop is
+    indistinguishable from zero, but this loop cuts overall step memory usage
+    roughly in half for at least one test association.
+    """
+    with warnings.catch_warnings():
+        warnings.filterwarnings(action="ignore",
+                                message="All-NaN slice encountered",
+                                category=RuntimeWarning)
+        output_arr = np.empty(cube.shape[1:], dtype=np.float32)
```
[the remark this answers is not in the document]
emolter: changed

tests/outlier_detection/test_median.py
emolter: We aren't ignoring the warning, we are upgrading it to a failure so that the test does fail if a PR breaks cleanup

emolter:
@braingram I think I incorporated all your comments with the most recent push.  Thanks for the docstring updates, I think they improved things a lot, and I agree with your idea to make the OnDiskMedian and DiskAppendableArray private

tests/outlier_detection/test_median.py
braingram: Thanks. Yeah I certainly got confused there. What about adding `PytestUnraisableExceptionWarning` to `pyproject.toml` instead:
https://github.com/spacetelescope/stcal/blob/4ba7a620ad6545067695fe3808b405a7a50e4d53/pyproject.toml#L102-L104
Hopefully it doesn't cause unrelated failures but if it does I'd be in favor of adding it as a separate PR and leaving the filtering out of this one.

emolter: It appears not to cause any unrelated failures so I went ahead and added it

tests/outlier_detection/test_median.py
```diff
@@ -0,0 +1,196 @@
...
+
+from stcal.outlier_detection.median import (
+    MedianComputer,
+    _DiskAppendableArray,
+    _OnDiskMedian,
+    nanmedian3D,
+)
+
+
+def test_disk_appendable_array(tmpdir):
```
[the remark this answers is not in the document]
emolter: done

pyproject.toml
```diff
@@ -101,6 +101,7 @@ norecursedirs = [
 ]
 filterwarnings = [
     "error::ResourceWarning",
+    "error:.*finalize.*:pytest.PytestUnraisableExceptionWarning",
```
braingram, line 104:
```suggestion
    # Files left open by tests can trigger ResourceWarnings during test
    # teardown which otherwise would be converted back into a warning
    # by pytest. Turn these PytestUnraisableExceptionWarnings back
    # into errors
    "error::pytest.PytestUnraisableExceptionWarning",
```
Is the message wildcard needed?

braingram:
> Would you add `median.py` to the docs? It will likely be easier if the submodule has an `__all__` (I added a comment about this) which will also help to define the "public" API.

Was the addition of this new API to the docs not pushed?

emolter:
> > Would you add `median.py` to the docs? It will likely be easier if the submodule has an `__all__` (I added a comment about this) which will also help to define the "public" API.
> 
> Was the addition of this new API to the docs not pushed?

I didn't even see the comment, my fault.  I will look at the docs from this branch and see what changes are needed.  I did already add the `__all__`

pyproject.toml
emolter: I committed the suggestion, but why is this wildcard not a good thing?  I thought it should catch only specifically the exception that happens when cleanup fails. I don't really know the full scope of when an UnraisableExceptionWarning can happen

braingram: Honestly it's more for the prior PR that removed pytest-openfiles than this PR (where the finalize filter would work). Leaving a file open can result in a PytestUnraisableExceptionWarning that doesn't contain 'finalize'.

One reason related to this PR is that I'm not sure if pytest considers it's warning messages as part of it's stable API. I kind of doubt it so any pytest release might change the warning message. I expect that the exception class is part of the stable API.

emolter: good point RE the stability of the API and thanks for explaining your thinking

braingram:
Downstream romancal failures are what I'd consider unrelated. It appears the CI is using the pytest configuration from stcal when running those tests (instead of using the one in romancal). The jwst tests haven't finished but I wouldn't be surprised if it does the same thing (and is now turning some unclosed file warnings into errors).

Thanks for updating the docs, they look good to me:
https://stcal--292.org.readthedocs.build/en/292/stcal/outlier_detection/index.html

I think https://github.com/spacetelescope/stcal/pull/292#discussion_r1782630616 is the only unresolved comment I made.

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        Append data to the median computer.
+
+        Parameters
+        ----------
+        data
+            The data to append to the median computer.
+
+        idx
+            The index at which to append the data. Required if using in-memory
```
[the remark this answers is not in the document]
emolter: sorry, missed this, fixed

src/stcal/outlier_detection/median.py
```diff
@@ -0,0 +1,400 @@
...
+        """
+        Append data to the median computer.
+
+        Parameters
+        ----------
+        data
+            The data to append to the median computer. Must have shape full_shape[1:].
+
+        idx
+            The index at which to append the data. Mus be between 0 and full_shape[0].
```
braingram, line 118:
```suggestion
            The index at which to append the data. Must be between 0 and full_shape[0].
```

braingram (approved):
Thanks! This looks great to me.

PR #293, opened 2024-10-01
Enable style checks

braingram:
Style checks (via pre-commit) have not been checking any files in this repo as part of the CI.
See this recent run on main:
https://github.com/spacetelescope/stcal/actions/runs/11017955113/job/30597119856#step:10:63
With all checks showing:
```
check for added large files..........................(no files to check)Skipped
```
If I run them locally I see a large number of automatic changes:
36 files changed, 1091 insertions(+), 942 deletions(-)
and even after these see 79 ruff errors, many from codespell and 6 from reporeview.

In an effort to check something this PR:
- fixes the `check-style` job to check files
- disables all failing ruff and pre-commit rules (maybe we can re-enable these one-by-one in follow-up PRs)

With this PR the checks run:
https://github.com/spacetelescope/stcal/actions/runs/11130795981/job/30930929199?pr=293#step:10:49

## Tasks

</details>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,8 +12,8 @@ repos:
       - id: check-symlinks
       - id: debug-statements
       - id: detect-private-key
-      - id: end-of-file-fixer
-      - id: trailing-whitespace
+      # - id: end-of-file-fixer
+      # - id: trailing-whitespace
 
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.10.0
@@ -22,26 +22,26 @@ repos:
       - id: rst-inline-touching-normal
       - id: text-unicode-replacement-char
 
-  - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.5
-    hooks:
-      - id: codespell
-        args: ["--write-changes"]
-        additional_dependencies:
-          - tomli
+  # - repo: https://github.com/codespell-project/codespell
+  #   rev: v2.2.5
+  #   hooks:
+  #     - id: codespell
+  #       args: ["--write-changes"]
+  #       additional_dependencies:
+  #         - tomli
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.1.5"
     hooks:
       - id: ruff
         args: ["--fix", "--show-fixes"]
-      - id: ruff-format
+      # - id: ruff-format
 
   - repo: https://github.com/MarcoGorelli/cython-lint
     rev: v0.15.0
     hooks:
       - id: cython-lint
-      - id: double-quote-cython-strings
+      # - id: double-quote-cython-strings
 
   - repo: https://github.com/pycqa/isort
     rev: 5.12.0
@@ -57,7 +57,7 @@ repos:
         additional_dependencies:
           - black==22.12.0
 
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v3.0.1"
-    hooks:
-      - id: prettier
+  # - repo: https://github.com/pre-commit/mirrors-prettier
+  #   rev: "v3.0.1"
+  #   hooks:
+  #     - id: prettier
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -30,7 +30,7 @@ deps =
     pre-commit
 commands =
     pre-commit install-hooks
-    pre-commit run {posargs:--color always --all-files --show-diff-on-failure}
+    pre-commit run --color always --all-files --show-diff-on-failure {posargs}
```
braingram, line 33: Since the CI runs this with a trailing `--` tox was skipping the important `--all-files` and instead trying to run on only non-committed files of which there were none.

zacharyburnett (approved):

tox.ini
braingram: We could leave it there to always be green 😆

.pre-commit-config.yaml
```diff
@@ -12,8 +12,8 @@ repos:
       - id: check-symlinks
       - id: debug-statements
       - id: detect-private-key
-      - id: end-of-file-fixer
-      - id: trailing-whitespace
+      # - id: end-of-file-fixer
```
kmacdonald-stsci, line 15: Why are these removed?  Don't we want these checks?  And why comment them out, instead of deleting them?

.pre-commit-config.yaml
```diff
@@ -57,13 +57,7 @@ repos:
...
-    rev: "v3.0.1"
-    hooks:
-      - id: prettier
-
-  - repo: https://github.com/scientific-python/cookie
-    rev: 2023.10.27
-    hooks:
-      - id: sp-repo-review
-        additional_dependencies: ["repo-review[cli]"]
+  # - repo: https://github.com/pre-commit/mirrors-prettier
```
kmacdonald-stsci, line 60: Why is this commented out and not simply removed?

.pre-commit-config.yaml
```diff
@@ -22,26 +22,26 @@ repos:
...
       - id: text-unicode-replacement-char
 
-  - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.5
-    hooks:
-      - id: codespell
-        args: ["--write-changes"]
-        additional_dependencies:
-          - tomli
+  # - repo: https://github.com/codespell-project/codespell
```
kmacdonald-stsci, line 25: Why is this commented out instead of just removed?

pyproject.toml
```diff
@@ -120,50 +120,50 @@ src = [
 ]
 
 [tool.ruff.lint]
-extend-select = [
-    "F",      # Pyflakes (part of default flake8)
-    "W", "E", # pycodestyle (part of default flake8)
-    "I",      # isort (import sorting)
+select = [
+    #"F",      # Pyflakes (part of default flake8)
```
kmacdonald-stsci, line 124: Why are these commented out instead of just removed?

kmacdonald-stsci (approved):

.pre-commit-config.yaml
braingram: Doh, I had clicked auto-merge so this got merged as soon as you approved it. I can follow this up with a PR to remove them if that's what you'd like.

I commented them out because these are all failing and likely have been for some time. The PR that enabled them had a bug in the workflow where it ran the checks but never supplied any files to check (so it was checking nothing). See https://github.com/spacetelescope/stcal/pull/293#discussion_r1783271648 for a comment on where this bug is fixed. Here's a PR that uncomments all the checks to show the extent of the failures:
https://github.com/spacetelescope/stcal/pull/294
The reported diff for the checks is over 5000 lines long:
https://github.com/spacetelescope/stcal/actions/runs/11142692931/job/30966194873?pr=294
(and doesn't fix all the failures and many require manual updates).

I chose to not remove them because I assumed that the maintainers want these checks to run. If that's not the case I'd say we remove them entirely. If the re-enabling the checks are preferred I was thinking it would be easiest to enable them perhaps one-by-one since each will require file changes.

kmacdonald-stsci: I'm not sure what's the best approach for this.  It could be good to have them commented out, so we have explicit documentation for what is not used, but removing them to clean up the script could be good, too.  I just wanted to make sure you wanted the comments left in the scripts.

PR #295, opened 2024-10-02
Adjust style checks to align with jwst

braingram:
This PR adjusts the style checks to roughly align with jwst (with a few adjustments for things like this package using cython).

For reference the ruff checks and pre-commit hooks for jwst are as follows:
https://github.com/spacetelescope/jwst/blob/e5c73fa23d157f4aa65416b4070c2b838e978494/pyproject.toml#L254
https://github.com/spacetelescope/jwst/blob/dd76184e82c1e863bce22d02de093f9bfe12f16f/.github/workflows/ci.yml

This PR sets the checks in this repo to roughly align:
- use default ruff rules
- extend ruff rules to include isort (which was previously a pre-commit hook in this repo)
- run bandit (like jwst does)
- enable cython-lint (as this package uses cython)

I applied auto-fixes in:
https://github.com/spacetelescope/stcal/pull/295/commits/a85bd00a2ceb5c9f1a26ea5992e04afd28c7741f
These are largely import sorting but I think it's worth a second set of eyes on these changes as I did notice one log message formatting change that likely indicates a bug in the log message (where it's currently not formatting the expected variable):
https://github.com/spacetelescope/stcal/pull/295/commits/a85bd00a2ceb5c9f1a26ea5992e04afd28c7741f#diff-c04284467dcbe86296aa89b83fd434e0dde8a32d790d157a9588ba3c7d169484R264

I then did some manual fixes removing unused variables and a duplicate test in:
https://github.com/spacetelescope/stcal/pull/295/commits/3c59de9a3fecca54438291c716a68e65e89939fa

There are still failures with this PR that I'm not sure how to address.
```

src/stcal/ramp_fitting/likely_algo_classes.py:289:24: F821 Undefined name `fit_ramps`
    |
287 |         n = alpha.shape[0]
288 |         z = np.zeros((len(cvec), len(countrates)))
289 |         result_low_a = fit_ramps(z, self, sig, countrateguess=countrates)
    |                        ^^^^^^^^^ F821
290 | 
291 |         # try to avoid problems with roundoff error
    |

src/stcal/ramp_fitting/likely_algo_classes.py:294:25: F821 Undefined name `fit_ramps`
    |
292 |         da_incr = da * (countrates[np.newaxis, :] + sig**2)
293 | 
294 |         result_high_a = fit_ramps(z, self, sig, countrateguess=countrates + da_incr)
    |                         ^^^^^^^^^ F821
295 |         # finite difference approximation to dw/da
    |

tests/test_ramp_fitting_likely_fit.py:613:34: F821 Undefined name `cube1`
    |
611 | def dbg_print_cube(cube, pix=(0, 0), label=None):
612 |     data, dq, vp, vr, err = cube
613 |     data1, dq1, vp1, vr1, err1 = cube1
    |                                  ^^^^^ F821
614 |     row, col = pix
615 |     nints = data1.shape[0]
```

2 are uses of an undefined `fit_ramps` in `likely_algo_classes.py`:
https://github.com/spacetelescope/stcal/blob/787aa81a6719bb0c72be8972171c7f3f88c8adf0/src/stcal/ramp_fitting/likely_algo_classes.py#L289
These would crash if anything tried to execute that code.

The last one is the use of an undefined `cube` in `dbg_print_cube` (which is unused). Calling this function would crash.

## Tasks

</details>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,5 @@
+exclude: ".*\\.asdf$"
+
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
@@ -15,49 +17,20 @@ repos:
       # - id: end-of-file-fixer
       # - id: trailing-whitespace
 
-  - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: v1.10.0
-    hooks:
-      - id: rst-directive-colons
-      - id: rst-inline-touching-normal
-      - id: text-unicode-replacement-char
-
-  # - repo: https://github.com/codespell-project/codespell
-  #   rev: v2.2.5
-  #   hooks:
-  #     - id: codespell
-  #       args: ["--write-changes"]
-  #       additional_dependencies:
-  #         - tomli
-
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.1.5"
+    rev: 'v0.5.2'
     hooks:
       - id: ruff
-        args: ["--fix", "--show-fixes"]
+        args: ["--fix"]
       # - id: ruff-format
+
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.9
+    hooks:
+      - id: bandit
+        args: ["-r", "-ll", "src", "tests"]
 
   - repo: https://github.com/MarcoGorelli/cython-lint
     rev: v0.15.0
     hooks:
       - id: cython-lint
-      # - id: double-quote-cython-strings
-
-  - repo: https://github.com/pycqa/isort
-    rev: 5.12.0
-    hooks:
-      - id: isort
-        name: isort (cython)
-        types: [cython]
-
-  - repo: https://github.com/adamchainz/blacken-docs
-    rev: 1.16.0
-    hooks:
-      - id: blacken-docs
-        additional_dependencies:
-          - black==22.12.0
-
-  # - repo: https://github.com/pre-commit/mirrors-prettier
-  #   rev: "v3.0.1"
-  #   hooks:
-  #     - id: prettier
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ exclude: ".*\\.asdf$"
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v4.6.0
     hooks:
       - id: check-added-large-files
       - id: check-ast
@@ -18,19 +18,19 @@ repos:
       # - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.5.2'
+    rev: 'v0.6.8'
     hooks:
       - id: ruff
         args: ["--fix"]
       # - id: ruff-format
 
   - repo: https://github.com/PyCQA/bandit
-    rev: 1.7.9
+    rev: 1.7.10
     hooks:
       - id: bandit
         args: ["-r", "-ll", "src", "tests"]
 
   - repo: https://github.com/MarcoGorelli/cython-lint
-    rev: v0.15.0
+    rev: v0.16.2
     hooks:
       - id: cython-lint
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -262,8 +262,8 @@ def ramp_fit_data(
     ngroups = ramp_data.data.shape[1]
     if algorithm.upper() == "LIKELY" and ngroups < likely_fit.LIKELY_MIN_NGROUPS:
         log.info("When selecting the LIKELY ramp fitting algorithm the"
-                  " ngroups needs to be a minimum of {likely_fit.LIKELY_MIN_NGROUPS},"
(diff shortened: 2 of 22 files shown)


braingram:
@kmacdonald-stsci Any advice on how to handle the bugs revealed by these style checks?

Can `dbg_print_cube` be removed?

What is the expected definition of `fit_ramps` in `likely_algo_classes.py`?

kmacdonald-stsci:
> @kmacdonald-stsci Any advice on how to handle the bugs revealed by these style checks?
> 
> Can `dbg_print_cube` be removed?
> 
> What is the expected definition of `fit_ramps` in `likely_algo_classes.py`?

I have a few `dbg_` functions that I use for development and debugging, but aren't used at any other time.  I'd prefer not to remove those functions.  

The `fit_ramps` is defined here:

https://github.com/spacetelescope/stcal/blob/787aa81a6719bb0c72be8972171c7f3f88c8adf0/src/stcal/ramp_fitting/likely_fit.py#L493

braingram:
Thanks!

 > I have a few `dbg_` functions that I use for development and debugging, but aren't used at any other time. I'd prefer not to remove those functions.

I added a noqa for this line. It will however continue to fail if used.

> The `fit_ramps` is defined here:
> 
> https://github.com/spacetelescope/stcal/blob/787aa81a6719bb0c72be8972171c7f3f88c8adf0/src/stcal/ramp_fitting/likely_fit.py#L493

Since likely_fit imports likely_algo_classes:
https://github.com/spacetelescope/stcal/blob/787aa81a6719bb0c72be8972171c7f3f88c8adf0/src/stcal/ramp_fitting/likely_fit.py#L16
I think the import of `fit_ramps` needs to be local to avoid a circular import error. I updated the PR to include the local import.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -261,7 +261,7 @@ def ramp_fit_data(
     # a minimum of a four group ramp is needed.
     ngroups = ramp_data.data.shape[1]
     if algorithm.upper() == "LIKELY" and ngroups < likely_fit.LIKELY_MIN_NGROUPS:
-        log.info(f"When selecting the LIKELY ramp fitting algorithm the"
```
braingram, line 264: Here's one likely bug in this log message. Since only first first item is a f-string all the subsequent items don't get formatted. This means the log message will contain the string "{likely_fit.LIKELY_MIN_NGROUPS}" instead of the formatted string for that variable.

tests/test_jump.py
```diff
@@ -485,42 +484,6 @@ def test_find_faint_extended_sigclip():
     #  Check that the flags are not applied in the 3rd group after the event
     assert np.all(gdq[0, 4, 12:22, 14:23]) == 0
 
-# No shower is found because the event is identical in all ints
-def test_find_faint_extended_sigclip():
```
braingram, line 489: This test is duplicated. See:
https://github.com/spacetelescope/stcal/blob/787aa81a6719bb0c72be8972171c7f3f88c8adf0/tests/test_jump.py#L439
and
https://github.com/spacetelescope/stcal/blob/787aa81a6719bb0c72be8972171c7f3f88c8adf0/tests/test_jump.py#L489

braingram:
@kmacdonald-stsci any suggestions for a second reviewer? Most of the changes are code style but the duplicate jump test might be worth checking with a maintainer of that code to see if the duplicate is supposed to be different.

kmacdonald-stsci:
> @kmacdonald-stsci any suggestions for a second reviewer? Most of the changes are code style but the duplicate jump test might be worth checking with a maintainer of that code to see if the duplicate is supposed to be different.

For anything in STCAL, if you need an additional reviewer besides me, I think you can always tag @melanieclarke and/or @tapastro.

src/stcal/ramp_fitting/ramp_fit.py
braingram: Thanks. Since it changed the log message I originally left it out of the style changes. I just pushed a fix in 84dbac6

melanieclarke: Looks good, thanks for the fix!

melanieclarke (approved):

braingram:
Thanks for the quick review @melanieclarke Let me know how the jwst checks evolve and we can update these as well.

.pre-commit-config.yaml
```diff
@@ -15,49 +17,20 @@ repos:
...
 
-  - repo: https://github.com/MarcoGorelli/cython-lint
-    rev: v0.15.0
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.10
     hooks:
-      - id: cython-lint
-      # - id: double-quote-cython-strings
+      - id: bandit
+        args: ["-r", "-ll", "src", "tests"]
```
WilliamJamieson, line 31: Why is this being added? Ruff can do this check, https://docs.astral.sh/ruff/rules/#flake8-bandit-s.

braingram: This was added to align the checks with jwst which uses bandit:
https://github.com/spacetelescope/jwst/blob/ffa02661593f4c53d66ab17fa837aa72a358def6/.pre-commit-config.yaml#L28

braingram: No objection from me but that's beyond the scope of this PR.

PR #297, opened 2024-10-02
Isolate jwst/romancal pytest configuration

braingram:
Currently downstream testing of jwst and romancal end up using the pyproject.toml in this repository for determining pytest configuration.
https://github.com/spacetelescope/stcal/actions/runs/11142379456/job/30965215790#step:10:107
This ends up with pytest using options in these downstream tests that differ from the options used by the downstream packages in their tests (when run in their own CI). This in turn results in lots of false failures in our downstream testing.

This PR changes how the jwst and romancal downstream tests to use the pytest options defined in the respective repositories.

## Tasks

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,3 +45,4 @@ jobs:
       envs: |
         - linux: py311-jwst-cov-xdist
         - linux: py311-romancal-cov-xdist
+      coverage: codecov
--- a/changes/297.general.rst
+++ b/changes/297.general.rst
@@ -1,0 +1,1 @@
+Update downstream tests for jwst and romancal to fix pytest configurations.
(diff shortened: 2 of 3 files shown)


kmacdonald-stsci (approved):
Needs a changes fragment for the change log, but otherwise fine.

braingram:
> Needs a changes fragment for the change log, but otherwise fine.

Thanks! I pushed on in 24b0a2b I gave it the 'general' category. I originally left one off since this is all internal/testing changes. If it's helpful to have these in the changelog we could add a 'testing' catagory for these types of changes.

braingram:
Thanks for finding that. I removed those from the envlist in 60bd3b86914202b228cf49b17fd079bee30f3f94 The oldestdeps job should check the lower numpy pin, the "normal" jobs the latest release and the devdeps the dev version so I think we have numpy well covered by those.

jhunkeler:
No problem! I don't see anything else that jumps out at me.

jhunkeler (approved):

PR #299, opened 2024-10-07
Infrastructure for memory usage tests

emolter:
This PR adds a context manager to more easily check the peak memory usage of a code block and raise an error if a memory limit is exceeded.  A unit test of the memory usage in the median calculation in outlier detection provides an example use-case.

Relates to [JP-3775](https://jira.stsci.edu/browse/JP-3775) but only fixes one small part of that ticket.

## Tasks

</details>

### DIFF
--- a/src/stcal/testing_helpers.py
+++ b/src/stcal/testing_helpers.py
@@ -1,6 +1,6 @@
 import tracemalloc
 
-MEMORY_UNIT_CONVERSION = {"B": 1, "KB": 1024, "MB": 1024 ** 2, "GB": 1024 ** 3}
+MEMORY_UNIT_CONVERSION = {"B": 1, "KB": 1024, "MB": 1024 ** 2, "GB": 1024 ** 3, "TB": 1024 ** 4}
 
 class MemoryThresholdExceeded(Exception):
     pass
@@ -15,7 +15,7 @@ class MemoryThreshold:
         # code that should not exceed expected
 
     If the code in the with statement uses more than the expected_usage
-    memory (in bytes) a ``MemoryThresholdExceeded`` exception
+    memory a ``MemoryThresholdExceeded`` exception
     will be raised.
     
     Note that this class does not prevent allocations beyond the threshold
@@ -30,7 +30,7 @@ class MemoryThreshold:
         expected_usage : str
             Expected peak memory usage expressed as a whitespace-separated string
             with a number and a memory unit (e.g. "100 KB").
-            Supported units are "B", "KB", "MB", "GB".
+            Supported units are "B", "KB", "MB", "GB", "TB".
         """
         expected, self.units = expected_usage.upper().split()
         self.expected_usage_bytes = float(expected) * MEMORY_UNIT_CONVERSION[self.units]
--- a/src/stcal/testing_helpers.py
+++ b/src/stcal/testing_helpers.py
@@ -13,6 +13,14 @@ class MemoryThreshold:
     example usage:
     with MemoryThreshold(expected_usage):
         # code that should not exceed expected
+
+    If the code in the with statement uses more than the expected_usage
+    memory (in bytes) a ``MemoryThresholdExceeded`` exception
+    will be raised.
+    
+    Note that this class does not prevent allocations beyond the threshold
+    and only checks the actual peak allocations to the threshold at the
+    end of the with statement.
     """
 
     def __init__(self, expected_usage, log_units="KB"):
--- a/src/stcal/testing_helpers.py
+++ b/src/stcal/testing_helpers.py
@@ -1,5 +1,10 @@
 import tracemalloc
 
+
+class MemoryThresholdExceeded(Exception):
+    pass
+
+
 class MemoryThreshold:
     """
     Context manager to check peak memory usage against an expected threshold.
@@ -26,7 +31,7 @@ class MemoryThreshold:
         _, peak = tracemalloc.get_traced_memory()
         tracemalloc.stop()
 
-        assert peak <= self.expected_usage, (
-            "Peak memory usage exceeded expected usage: "
-            f"{peak / 1024:.2f} KB > {self.expected_usage / 1024:.2f} KB"
-        )
+        if peak > self.expected_usage:
+            msg = ("Peak memory usage exceeded expected usage: "
+                  f"{peak / 1024:.2f} KB > {self.expected_usage / 1024:.2f} KB")
+            raise MemoryThresholdExceeded(msg)
--- a/src/stcal/testing_helpers.py
+++ b/src/stcal/testing_helpers.py
@@ -1,5 +1,7 @@
 import tracemalloc
 
+MEMORY_UNIT_CONVERSION = {"B": 1, "KB": 1024, "MB": 1024 ** 2, "GB": 1024 ** 3}
+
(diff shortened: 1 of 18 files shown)


braingram (commented):
Since the parent ticket [JP-3775](https://jira.stsci.edu/browse/JP-3775) is for building out infrastructure for tests like this, what do you think about building out that infrastructure and using it here? Something like:
```python
with memory_threshold(expected_mem):
    computer = MedianComputer...
```
I can understand that the infrastructure is beyond the scope of this PR but I think an argument can be made that building the infrastructure first (before this PR) makes sense since:
- this is a useful test for the infrastructure
- if we merge this PR first a follow-up PR will modify this test again when we have the new infrastructure

emolter:
> I can understand that the infrastructure is beyond the scope of this PR but I think an argument can be made that building the infrastructure first (before this PR) makes sense

I completely agree.  I'll work on making a context manager that is called like you show

emolter:
@braingram I added the context manager and a fixture that calls it.  I wasn't sure exactly how to structure this, because based on what I read it's not great practice to import from conftest.py directly.  Let me know if the proposed helper.py and conftest.py look right to you.

Also, do you think it's necessary/useful to add infrastructure for testing the fixture itself like in [this StackOverflow answer](https://stackoverflow.com/a/56635224)?

braingram:
> @braingram I added the context manager and a fixture that calls it. I wasn't sure exactly how to structure this, because based on what I read it's not great practice to import from conftest.py directly. Let me know if the proposed helper.py and conftest.py look right to you.
> 
> Also, do you think it's necessary/useful to add infrastructure for testing the fixture itself like in [this StackOverflow answer](https://stackoverflow.com/a/56635224)?

Since the fixture isn't needed for this test I'd say we leave it out of this PR. The class you added in `tests.helpers` seems to cover all the needs for the added test. I'll leave a few other comments inline.

tests/helpers.py
```diff
@@ -0,0 +1,32 @@
+import tracemalloc
```
braingram, line 1: Let's put this in the packaged files (under `src/stcal`)? That way it can be used from jwst, etc.

tests/helpers.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __enter__(self):
+        tracemalloc.start()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        _, peak = tracemalloc.get_traced_memory()
+        tracemalloc.stop()
+
+        assert peak <= self.expected_usage, (
```
braingram, line 29: I'm on the fence about this being an `assert` vs a custom exception. Since it's only intended for testing `assert` seems fine but can you think of a time we might want to catch this in a test? If so, I'd say we make and use a custom exception.

tests/helpers.py
kmacdonald-stsci: Is this expected to be used by `STCAL` or `JWST` or just for testing?  If it's just for testing, I think it should be here.

kmacdonald-stsci (approved):

tests/helpers.py
braingram: The expectation is that this will be used in the jwst tests. Here's one example that can be replaced with the class added here:
https://github.com/spacetelescope/jwst/blob/f3df1e805c126ed7593a4bd2447b124935a6dfa7/jwst/regtest/test_miri_image.py#L126
The overall goal is to add memory checks to stcal, jwst, romancal code to make sure future code changes don't unexpectedly increase memory usage.

tests/helpers.py
emolter: would the idea then be that the tests would error out instead of failing?

tests/helpers.py
emolter: I think we're deciding that.  I think it would be good to have it available to jwst and perhaps romancal for testing, in which case it needs to be importable and therefore somewhere within `src/stcal` as Brett said.  If the `assert` statement were changed to an exception, then it could be used elsewhere, but currently I don't think there's an existing use-case for that.

https://github.com/spacetelescope/jwst/issues/8775 comes to mind - I could imagine a scenario where a custom exception would be useful if something tries to allocate more memory than allowed.  But for this particular case I still don't think this would be used, since we would want to figure out the predicted memort allocation _before_ attempting to allocate it

tests/helpers.py
braingram: The test would fail (not error) for a non-assert exception (assuming this class is used in a test not a fixture).

One place where we might want to use a custom exception would be in testing of this class.
```python
def test_memory_limit():
    with pytest.raises(MemoryThresholdExceeded):
        with MemoryThreshold(100):
            buff = b"\x00" * 1000
```
maybe the other place is if we use this as part of a fixture that also runs a pipeline (like the regtests in jwst). Given all the unknowns there it's probably not worth at this point trying to implement the needed features.

I can't think of a downside to making this a custom exception. I'm good either way.

tests/test_infrastructure.py
```diff
@@ -0,0 +1,16 @@
+"""Tests of custom testing infrastructure"""
+
+import pytest
+import numpy as np
+from stcal.testing_helpers import MemoryThreshold, MemoryThresholdExceeded
+
+
+def test_memory_threshold():
+    with MemoryThreshold(1000):
```
braingram, line 9: Looks like some conditions have too much overhead for this test and it's failing with python 3.12.
https://github.com/spacetelescope/stcal/actions/runs/11259184554/job/31307581409?pr=299

emolter: thanks yeah noticed that

src/stcal/testing_helpers.py
```diff
@@ -0,0 +1,36 @@
...
+    pass
+
+
+class MemoryThreshold:
+    """
+    Context manager to check peak memory usage against an expected threshold.
+
+    example usage:
+    with MemoryThreshold(expected_usage):
+        # code that should not exceed expected
```
braingram, line 13:
```suggestion
    Context manager to check peak memory usage against an expected threshold.

    example usage:
    with MemoryThreshold(expected_usage):
        # code that should not exceed expected
    
    If the code in the with statement uses more than the expected_usage
    memory (in bytes) a ``MemoryThresholdExceeded`` exception
    will be raised.
    
    Note that this class does not prevent allocations beyond the threshold
    and only checks the actual peak allocations to the threshold at the
    end of the with statement.
```
How does this look? I was hoping to add:
- the exception that is raised when the threshold is passed
- a note that this class doesn't cap the memory allocations so code can still allocate more than the threshold (it will just fail at the end if the peak was more than the threshold).

emolter: seems fine to me

braingram (approved):
Thanks! LGTM.

emolter:
status update: @penaguerrero is using this to write a test for the jump step, and in so doing is evaluating this PR.  So this is awaiting her review

penaguerrero (commented):
I tried the code in writing a test for jwst, and it works as intended. Other than suggesting to allow to present the result in other units I have no further comments. LGTM

emolter:
@penaguerrero can you test the units option I added and see if it covers your use-case?

penaguerrero (approved):
I just tested again with the changes. It works as expected. LGTM!

PR #302, opened 2024-10-08
Removed redundant calculations, unneeded array allocations from jump step

t-brandt:
Resolves [JP-3697](https://jira.stsci.edu/browse/JP-3697)

Closes #

This PR affects the two-point-difference jump step only.  It removes duplicate and redundant calculations and unneeded array allocations, which make a significant difference for large uncal files.  Masked array median is also changed to nanmedian, and a few variables are renamed to avoid overwriting a quantity that is needed at the end of the routine.  

## Tasks

</details>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -76,8 +76,7 @@ General
   <https://github.com/spacetelescope/stcal/issues/286>`_)
 - Improve handling of catalog web service connectivity issues. (`#286
   <https://github.com/spacetelescope/stcal/issues/286>`_)
-- [jump] Remove redundant calculations and unneeded array allocations.
-  (`JP-3697<https://jira.stsci.edu/browse/JP-3697>`_)
+
 
 1.8.2 (2024-09-10)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,8 @@ General
   <https://github.com/spacetelescope/stcal/issues/286>`_)
 - Improve handling of catalog web service connectivity issues. (`#286
   <https://github.com/spacetelescope/stcal/issues/286>`_)
-
+- [jump] Remove redundant calculations and unneeded array allocations.
+  (`JP-3697<https://jira.stsci.edu/browse/JP-3697>`_)
 
 1.8.2 (2024-09-10)
 ==================
--- a/src/stcal/jump/twopoint_difference.py
+++ b/src/stcal/jump/twopoint_difference.py
@@ -173,8 +173,6 @@ def find_crs(
         return gdq, row_below_gdq, row_above_gdq, 0, dummy
     else:
         # set 'saturated' or 'do not use' pixels to nan in data
-        dat[np.where(np.bitwise_and(gdq, sat_flag))] = np.nan
-        dat[np.where(np.bitwise_and(gdq, dnu_flag))] = np.nan
         dat[np.where(np.bitwise_and(gdq, dnu_flag + sat_flag))] = np.nan
 
         # calculate the differences between adjacent groups (first diffs)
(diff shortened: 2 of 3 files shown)


kmacdonald-stsci (approved):
There are some down stream test failures.  Are those related to this PR?  If not, this looks fine.

ddavis-stsci:
The romancal tests for these changes pass, 
https://github.com/spacetelescope/RegressionTests/actions/runs/11328631054

melanieclarke:
JWST regtests are running here:
https://github.com/spacetelescope/RegressionTests/actions/runs/11672280970

src/stcal/jump/twopoint_difference.py
```diff
@@ -173,8 +173,6 @@ def find_crs(
         return gdq, row_below_gdq, row_above_gdq, 0, dummy
     else:
         # set 'saturated' or 'do not use' pixels to nan in data
-        dat[np.where(np.bitwise_and(gdq, sat_flag))] = np.nan
-        dat[np.where(np.bitwise_and(gdq, dnu_flag))] = np.nan
         dat[np.where(np.bitwise_and(gdq, dnu_flag + sat_flag))] = np.nan
```
melanieclarke, line 176: Is there any chance that either the dnu or sat flags might be set without both being set?

CHANGES.rst
```diff
@@ -23,7 +23,8 @@ General
   <https://github.com/spacetelescope/stcal/issues/286>`_)
 - Improve handling of catalog web service connectivity issues. (`#286
   <https://github.com/spacetelescope/stcal/issues/286>`_)
-
+- [jump] Remove redundant calculations and unneeded array allocations.
+  (`JP-3697<https://jira.stsci.edu/browse/JP-3697>`_)
```
melanieclarke, line 27: This needs to be removed and added to a towncrier change note instead.

melanieclarke (commented):
These changes look pretty straightforward, and seem worth merging as is, before we do any more profiling or refactoring for jump.

One quick question about the change to the flag handling, below.  Also, the change note needs to be moved out of the CHANGES.rst file to a newer style towncrier note -- see the PR template description for the new method.  That should fix the readthedocs error and the changelog check.

src/stcal/jump/twopoint_difference.py
t-brandt: The bottom line subsumes the previous two, right?  As long as dnu_flag and sat_flag are different, bitwise_and will be nonzero any time either of those bits is set, so np.where will pick out all points where either flag is set?  I tested this myself to be absolutely sure.

t-brandt: In general, the syntax 
`np.where(np.bitwise_and(gdq, flag_a + flag_b + flag_c))`
should return all pixels where at least one of flags a, b, and c is set.  If it is clearer, we could write something like
`dat[np.bitwise_and(gdq, flag_a + flag_b + flag_c) != 0] = np.nan`
instead.

melanieclarke: Oh yes, you're right. I always struggle with bit-wise math.  Thanks!

t-brandt: Sure!  The only hole this opens is if someone is foolish enough to make sat_flag and dnu_flag equal.  Hopefully that gets checked somewhere? :-)

melanieclarke: Pretty sure that's not possible, but to be super safe, we could cover that case by using `|` instead of `+`.

t-brandt:
@melanieclarke Sorry for dropping the ball on this for a while.  If you have the time and bandwidth to make the minor changes, please go ahead.  Otherwise I'll try to do so over the weekend or next week.

melanieclarke:
Rerunning jwst regtests here:
https://github.com/spacetelescope/RegressionTests/actions/runs/12323165118

melanieclarke:
Romancal tests look fine.  There are some failures in jwst unit and regression tests that are related to some ongoing gwcs work.  I will let that settle and rerun before we merge this.

t-brandt:
I actually want to make a few more changes--I can squeeze a significant amount of additional speedup while shortening the code.  I will push those changes tonight or tomorrow.  We can work on merging this after 11.2 :-)

melanieclarke:
> We can work on merging this after 11.2 :-)

Sounds good, thanks.  Can you please mark this draft for now, and let us know when you're ready?

t-brandt:
@melanieclarke: I updated the PR with new changes.  The new changes are more significant so I think a bit more testing/more eyeballs are warranted, which presumably will wait until after 11.2 is done.  I checked that the generated gdq arrays were identical for a few test files.  Memory improvements are modest, speed improvements are large, and crashes/virtual memory/very slow performance when overflagging jumps should all be gone.

melanieclarke:
Superseded by #337

PR #304, opened 2024-10-10
JP-3708: Adding the CRMAG Array to the Optional Results Product

kmacdonald-stsci:
Resolves [JP-3708](https://jira.stsci.edu/browse/JP-3708)

This PR addresses a crash that occurs when running multiprocessing for ramp fitting with the C extension while choosing to save the optional results product.  This is because the CRMAG array was not implemented in the C extension, so the C extension returns a `NoneType`, so when the data slices attempt to reassemble an exception is thrown because there is no CRMAG to reassemble.

The CRMAG array is a 4-D array, with dimensions `(nints, max_num_crs, nrows, ncols)`.  For each integration ramp there is some number of cosmic ray jumps, including 0.  The dimension `max_num_crs` is the maximum number of jumps found over all integrations and all pixels.  For ramps that have fewer than the max number of jumps the last elements in that pixel ramp are 0.  For example, if the maximum number of jumps found was 5, but a ramp has only two jumps, the first two elements of the CRMAG array for that integration, row, and column are non-zero, while the last three are 0.  The non-zero elements of the ramp array are the magnitude of the jump, which is the difference between the two groups where a jump was detected.

The CRMAG is now implemented in the C extension and functions properly with multiprocessing.

## Tasks

</details>

### DIFF
--- a/src/stcal/ramp_fitting/src/slope_fitter.c
+++ b/src/stcal/ramp_fitting/src/slope_fitter.c
@@ -1352,7 +1352,7 @@ create_opt_res(
 
     /* cr_mag has different dimensions */
     dims[1] = rd->max_num_crs;
-    opt_res->cr_mag = (PyArrayObject*)PyArray_EMPTY(nd, dims, NPY_FLOAT, fortran);
+    opt_res->cr_mag = (PyArrayObject*)PyArray_ZEROS(nd, dims, NPY_FLOAT, fortran);
     if (!opt_res->cr_mag) {
         goto FAILED_ALLOC;
     }
--- a/src/stcal/ramp_fitting/src/slope_fitter.c
+++ b/src/stcal/ramp_fitting/src/slope_fitter.c
@@ -218,6 +218,8 @@ struct ramp_data {
     int save_opt;                   /* Save optional results value */
     int max_num_segs;               /* Max number of segments over all ramps. */
     struct simple_ll_node ** segs;  /* The segment list for each ramp. */
+    int max_num_crs;                /* Max number of segments over all ramps. */
+    struct cr_node ** crs;          /* The segment list for each ramp. */
     real_t * pedestal;              /* The pedestal computed for each ramp. */
 
     /* Meta data */
@@ -285,6 +287,20 @@ struct segment_list {
     npy_intp max_segment_length;    /* The max group length of a segment */
 }; /* END: struct segment_list */
 
+
+/* Cosmic ray node for linked list */
+struct cr_node {
+    struct cr_node * flink; /* Next cosmic ray. */
+    real_t crmag;           /* The magnitude of the cosmic ray. */
+};
+
+/* The list structure for cosmic rays. */
+struct cr_list {
+    struct cr_node * head;  /* The head node of the list */
+    struct cr_node * tail;  /* The tail node of the list */
+    ssize_t size;           /* The size of the list */
+};
+
 /*
  * For each integration, count how many groups had certain flags set.
  */
@@ -328,6 +344,11 @@ struct pixel_ramp {
     /* This needs to be an array for each integration */
     ssize_t max_num_segs;           /* Max number of segments in an integration */
     struct segment_list * segs;     /* Array of integration segments */
+
+    /* XXX CRs */
+    /* This needs to be an array for each integration */
+    ssize_t max_crs;
+    struct cr_list * crs;
 
     struct integ_gdq_stats * stats; /* Array of integration GDQ stats */
 
@@ -423,6 +444,12 @@ static int
 compute_integration_segments(
     struct ramp_data * rd, struct pixel_ramp * pr, struct segment_list * segs,
     int chargeloss, npy_intp integ);
+
+static int
+cr_list_add(struct cr_list * crs, real_t crmag);
+
+static void
+cr_list_clean(struct cr_list * crs);
 
 static int
 create_opt_res(struct opt_res_product * opt_res, struct ramp_data * rd);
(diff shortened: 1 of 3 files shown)


kmacdonald-stsci:
A regression test for this is here:

src/stcal/ramp_fitting/src/slope_fitter.c
```diff
@@ -1173,6 +1227,56 @@ compute_integration_segments(
...
+
+    if (0==crs->size) {
+        crs->head = new_node;
+        crs->tail = new_node;
+        crs->size = 1;
+    } else {
+        crs->tail->flink = new_node;
+        crs->tail = new_node;
+        crs->size++;
+    }
```
jemorrison, line 1254: @kmacdonald-stsci I have not used linked lists in c before. I need a quick overview. I googled it and these double linked lists ? I could use a little more detail so I can figure out what it going on. 
Is 'head' the pointer for the VERY FIRST CR segment. And 'tail' the location of the very last one. So are the segments beginning and ends stored or just the first and last locations ? 
In the cr_list_add on line 1248 why is the tail = new_node. I did not think head and tail would equal each other. Is that just for the first cr found. 
Also I am unclear that flink is supposed represent.

kmacdonald-stsci: The linked list I am using for this is a singly linked list.  There are a few different ways to implement this.  My implementation uses two structures, the linked list container and the nodes being listed, with only the forward link being defined in each node.  The forward link is the `flink` parameter that points to the next node in the list.  When `flink` is `NULL`, then then that node is the tail of the list:

https://github.com/spacetelescope/stcal/blob/0f3a23aaa946de203dbf0bbebd1972a29bc87f6e/src/stcal/ramp_fitting/src/slope_fitter.c#L283

The linked list data structure contains useful meta data for the list.  In particular, the head, the tail, and the total number of nodes in each list, which will be used to create the second dimension for the 4-D CRMAG array, which has dimensions `(nints, max_num_crs, nrows, ncols)`.  The tail parameter in the linked list is a convenience variable to go directly to the tail and add the next node, rather than have to traverse the list to the tail each time a new node is added to the list.

https://github.com/spacetelescope/stcal/blob/0f3a23aaa946de203dbf0bbebd1972a29bc87f6e/src/stcal/ramp_fitting/src/slope_fitter.c#L289

So the linked list with _n_ CRs detected looks like this:

```
cr_list->head = cr_node->flink = pointer to the next node
                cr_node->flink = pointer to the next node
                cr_node->flink = pointer to the next node
                cr_node->flink = pointer to the next node
                ...
cr_list->tail = cr_node->flink = NULL
cr->size = _n_
```

So, if some ramp looks like this:

```
data =    [1.,   2.,   13.,      14.,  15.,  16.,  37.,      38.,  39.,  40.,  41.   102.,     103., 104., 105.]
groupdq = [GOOD, GOOD, JUMP_DET, GOOD, GOOD, GOOD, JUMP_DET, GOOD, GOOD, GOOD, GOOD, JUMP_DET, GOOD, GOOD, GOOD]
```

Then the list of CR magnitudes for this ramp will look like this:

```
cr_list->head = {crmag = 11., flink = pointer to next node}
                           {crmag = 21., flink = pointer to next node}
cr_list->tail =    {crmag = 61., flink = NULL}
cr_list->size = 3
```

So, yes, cr_list->head is the very first CR magnitude encountered in the ramp and is the difference between that group data and the previous group data.  The cr_tail->tail is the last known CR magnitude computed, so if another one is found, a new node is allocated and added as the tail.  The result is an ordered linked list starting with the first detected CR magnitude and ending with the last detected CR magnitude.

At line 

https://github.com/spacetelescope/stcal/blob/0f3a23aaa946de203dbf0bbebd1972a29bc87f6e/src/stcal/ramp_fitting/src/slope_fitter.c#L1246

This conditional checks to see if a current list is empty.  If it's empty, the size will be 0.  When a CR is detected, the CR magnitude is computed and a new cr_node is allocated to store that value.  The cr_node is then added to the empty list, resulting in a list of size 1, therefore the head and tail are the same and the size is set to 1.  This list will look like this:

```
cr_list->head = cr_node->flink = NULL
cr_list->tail    = cr_node->flink = NULL
cr_list->size = 1
```

The "else" in this conditional means the list is not empty, so put the newly computed CR magnitude at tail of the list and the size is incremented by 1.

jemorrison (approved):
I have gone through the code a few times. To the best of my ability it looks  correct to me and frees the necessary memory when it is done. The comments you gave me were essential for me to understand the flow. If you can somehow add some of those details into the code so in the future when we need to look at the code it is easier to understand what is going on that might be helpful.

kmacdonald-stsci:
> I have gone through the code a few times. To the best of my ability it looks correct to me and frees the necessary memory when it is done. The comments you gave me were essential for me to understand the flow. If you can somehow add some of those details into the code so in the future when we need to look at the code it is easier to understand what is going on that might be helpful.

I added comments to the linked lists used to more thoroughly describe the implementations.  Please take a look at them to let me know if they help clarify how they are implemented.

melanieclarke:
I'm testing these changes on some sample data I have on disk, and this command, running without multiprocessing completes in seconds:
`strun ramp_fit jw01247667001_02101_00001_mirifulong_jump.fits --save_opt=True --opt_name=opt_serial.fits`

This one, with multiprocessing on, I killed after 5 minutes because it didn't look like it was ever going to complete:
`strun ramp_fit jw01247667001_02101_00001_mirifulong_jump.fits --save_opt=True --opt_name=opt_parallel.fits --maximum_cores=half`
After killing it, it reports 20 leaked semaphores.

This one, with multiprocessing on but optional output off completes in seconds:
`strun ramp_fit jw01247667001_02101_00001_mirifulong_jump.fits --maximum_cores=half`
After completion, it reports 10 leaked semaphores.

Can you please double check the interaction between multiprocessing and optional output results?  I'm wondering if there's a deadlock somewhere.

kmacdonald-stsci:
Closing to fix multiprocessing bug in the optional results product.

PR #305, opened 2024-10-10
Fix saturation flag issue for numpy 2+ in romancal

WilliamJamieson:
The saturation step fails in stcal when running using numpy 2+. This appears to be caused by the use of an `IntEnum` by Roman instead of a dictionary. I believe this might be due to a numpy bug reported in numpy/numpy#27540.

In the mean time, the easiest fix is to cast the values directly to python `int` even though the `IntEnum` members should be python `int` values. In theory, this should be a totally transparent change to all code except on numpy 2+ when passing an `IntEnum`, in which case it simply strips the `Enum` related bits of the integer.

## Tasks

</details>

### DIFF
--- a/changes/305.bugfix.rst
+++ b/changes/305.bugfix.rst
@@ -1,0 +1,1 @@
+Fix `IntEnum` saturation flag issue with numpy 2+ for romancal.


schlafly (approved):
LGTM, thanks William.

kmacdonald-stsci (approved):

PR #307, opened 2024-10-16
AL-837: Calculate output WCS for resample from already-known s_region

emolter:
Resolves [AL-837](https://jira.stsci.edu/browse/AL-837)

This PR simplifies the calculation of the output WCS for resample.  Previously, the `wcs_from_footprints` function accepted a list of WCS objects, one for each input model, and computed the footprints from those WCS objects to pass into `_calculate_new_wcs()`.  However, computing the footprints here is unnecessary because that information is already stored in the S_REGION keyword.  This PR makes `wcs_from_footprints` take in an actual footprint instead, in the form of either a numpy array or an S_REGION string.

This PR will require small changes to both jwst and romancal, both of which are using the `wcs_from_footprints` function.  I'm not sure there's a way to avoid that.  One other non-hidden function changed, `compute_fiducial`, but this is not used for JWST or Romancal - both repositories have their own copy of a function with that name.

## Tasks

</details>

### DIFF
--- a/changes/307.apichange.0.rst
+++ b/changes/307.apichange.0.rst
@@ -1,0 +1,1 @@
+Add wcs_from_sregions function to compute a combined WCS from a list of s_regions.
--- a/changes/307.apichange.1.rst
+++ b/changes/307.apichange.1.rst
@@ -1,0 +1,1 @@
+Deprecate wcs_from_footprints. Use wcs_from_sregions instead.
--- a/changes/307.apichange.rst
+++ b/changes/307.apichange.rst
@@ -1,1 +1,0 @@
-Make wcs_from_footprints accept s_regions instead of wcs objects
--- a/changes/307.apichange.rst
+++ b/changes/307.apichange.rst
@@ -1,0 +1,1 @@
+Make wcs_from_footprints accept s_regions instead of wcs objects
--- a/src/stcal/alignment/util.py
+++ b/src/stcal/alignment/util.py
@@ -596,6 +596,8 @@ def wcs_from_footprints(
     warnings.warn(msg, DeprecationWarning, stacklevel=2)
     _validate_wcs_list(wcs_list)
     footprints = [w.footprint() for w in wcs_list]
+    if ref_wcs is None:
+        ref_wcs = wcs_list[0]
     return wcs_from_sregions(
         footprints,
         ref_wcs,
@@ -717,7 +719,7 @@ def wcs_from_sregions(
 
     transform = _generate_tranform(
         ref_wcs,
-        ref_wcsinfo,
+        wcsinfo=ref_wcsinfo,
         pscale_ratio=pscale_ratio,
         pscale=pscale,
         rotation=rotation,
--- a/src/stcal/alignment/util.py
+++ b/src/stcal/alignment/util.py
@@ -512,9 +512,8 @@ def wcs_from_footprints(
         containing (RA, Dec) vertices demarcating the footprint of the input WCSs.
 
     ref_wcs :
-        A valid datamodel whose WCS is used as reference for the creation of the output
+        A WCS used as reference for the creation of the output
         coordinate frame, projection, and scaling and rotation transforms.
-        If not supplied the first model in the list is used as ``refmodel``.
 
     ref_wcsinfo : dict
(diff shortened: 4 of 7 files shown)


emolter:
JWST regression tests [started here](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1794/)

emolter:
tracking down JWST regression test failures now

mairanteodoro:
`romancal`'s [issue 1455](https://github.com/spacetelescope/romancal/issues/1455) needs to be resolved before this PR, since it seems that `TweakRegStep` is only updating the WCS but not `s_region`.

emolter:
@mairanteodoro good to know.  When you do update s_region, it might be good to be aware of [this comment](https://github.com/spacetelescope/stcal/blob/dfe1d6d51f0c2a400dff918a52c779328ae19b7d/src/stcal/alignment/util.py#L674).  The regression test failures for JWST come from the fact that `wcs.footprint()` and S_REGION differ by half a pixel due to JWST's use of `center=True` in `compute_s_region_imaging()`

kmacdonald-stsci (approved):

braingram:
Does this have a corresponding jwst PR? It is showing test failures in both jwst and romancal. Based on the comment from @mairanteodoro this can't go in until after romancal is updated to compute s_region. @emolter would you update this PR to avoid the breaking changes to the API that are causing the downstream failures?

emolter:
corresponding [JWST PR is here](https://github.com/spacetelescope/jwst/pull/8893).  I know this needs to wait for that Romancal change, and then someone (it could be me) will need to make a similar PR for Roman as JWST.

It's a bit pointless to "avoid" the breaking changes according to my understanding of what is going on here.  One of the main reasons we want to do this is so that we can avoid passing in a WCS object for every input model, preferring to pass in s_region strings instead.  And that's the part of the API that is changing.

I suppose it would be possible to support both, i.e., allow `wcs_from_footprints()` to take in either one, and I'll do that if you think it's a good idea, but I personally don't think it's necessarily a good idea.

braingram:
> corresponding [JWST PR is here](https://github.com/spacetelescope/jwst/pull/8893). I know this needs to wait for that Romancal change, and then someone (it could be me) will need to make a similar PR for Roman as JWST.
> 
> It's a bit pointless to "avoid" the breaking changes according to my understanding of what is going on here. One of the main reasons we want to do this is so that we can avoid passing in a WCS object for every input model, preferring to pass in s_region strings instead. And that's the part of the API that is changing.
> 
> I suppose it would be possible to support both, i.e., allow `wcs_from_footprints()` to take in either one, and I'll do that if you think it's a good idea, but I personally don't think it's necessarily a good idea.

I think this would be easier to manage and review if a new function was added `wcs_from_s_regions`. That wouldn't break jwst or romancal right? That would also mean this PR could be merged at any point without having to coordinate it with simultaneous merging of this a jwst and a romancal PR.

I believe stcal is semantically versioned and as `wcs_from_footprints` is a documented part of the public API any breaking change should be preceded by a deprecation warning and the change made in a major version change.

emolter:
> I think this would be easier to manage and review if a new function was added wcs_from_s_regions. That wouldn't break jwst or romancal right? That would also mean this PR could be merged at any point without having to coordinate it with simultaneous merging of this a jwst and a romancal PR.

> I believe stcal is semantically versioned and as wcs_from_footprints is a documented part of the public API any breaking change should be preceded by a deprecation warning and the change made in a major version change.

Sure, can do.  Should we deprecate `wcs_from_footprints`?  It will no longer be used by either jwst or roman, and while stcal is supposed to be generally applicable, I don't think we really have any other users.  So I'd be in favor of removing it just to avoid having to maintain it.  Plus, I dislike the name, as it doesn't take in footprints, it takes in a WCS list.

emolter:
new JWST regtests with this branch [here](https://github.com/spacetelescope/RegressionTests/actions/runs/11486982631)

nden (approved):
I think we need to come up with a consistent naming of fiducial and other user facing functions but this PR touches those only in private functions so this is for a future PR.

mairanteodoro:
Roman's `TweakRegStep` update for review: https://github.com/spacetelescope/romancal/pull/1484

PR #308, opened 2024-10-16
JP-3737: Handle micrometeorite flashes

drlaw1558:
This PR addresses JP-3737 by adding an additional check in the jump step to look for micrometeorite flashes, which manifest as bursts of jumps across a large fraction of the detector.

Kept in draft form for now pending potential future modifications.

## Tasks

</details>

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -60,6 +60,7 @@ def detect_jumps(
     min_diffs_single_pass=10,
     mask_persist_grps_next_int=True,
     persist_grps_flagged=25,
+    mmflashfrac=1.0,
 ):
     """
     This is the high-level controlling routine for the jump detection process.
@@ -221,6 +222,11 @@ def detect_jumps(
     min_diffs_single_pass : int
        The minimum number of groups to switch to flagging all outliers in a single pass.
 
+    mmflashfrac: float
+       Fraction of the array in a given group that has to have a jump detected
+       in order to flag the entire array as a jump.  This is designed to deal with
+       sporadic micrometeorite flashes.
+
     Returns
     -------
     gdq : int, 4D array
@@ -338,6 +344,12 @@ def detect_jumps(
             )
             log.info("Total showers= %i", num_showers)
             number_extended_events = num_showers
+
+        # This is where we look for micrometeorite flashes if the triggering
+        # threshold is less than the entire array
+        if (mmflashfrac < 1):
+            gdq = find_micrometeorite_flashes(gdq, jump_flag, mmflashfrac)
+
     else:
         yinc = int(n_rows // n_slices)
         slices = []
@@ -503,6 +515,12 @@ def detect_jumps(
             )
             log.info("Total showers= %i", num_showers)
             number_extended_events = num_showers
+
+        # This is where we look for micrometeorite flashes if the triggering
+        # threshold is less than the entire array
+        if (mmflashfrac < 1):
+            gdq = find_micrometeorite_flashes(gdq, jump_flag, mmflashfrac)
+
     elapsed = time.time() - start
     log.info("Total elapsed time = %g sec", elapsed)
 
@@ -514,6 +532,21 @@ def detect_jumps(
     # Return the updated data quality arrays
     return gdq, pdq, total_primary_crs, number_extended_events, stddev
 
+# Find micrometeorite flashes that light up the entire array in a small number
+# of groups.  Do this by looking for cases where greater than mmflash_pct percent
+# of the array is flagged as 'JUMP_DET' and flag the entire array.
+def find_micrometeorite_flashes(gdq, jump_flag, mmflashfrac):
+    npix = gdq.shape[2]*gdq.shape[3] # Number of pixels in array
+    # Loop over integrations
+    for ii in range(0,gdq.shape[0]):
+        # Loop over groups
+        for jj in range(0,gdq.shape[1]):
+            indx = np.where(gdq[ii,jj,:,:] & jump_flag != 0)
+            fraction = float(len(indx[0])) / npix
+            if (fraction > mmflashfrac):
+                gdq[ii,jj,:,:] = gdq[ii,jj,:,:] | jump_flag
+                log.info("Detected a micrometeorite flash in integ = %i, group= %i", ii, jj)
+    return gdq
 
 def flag_large_events(
     gdq,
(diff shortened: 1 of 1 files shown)


src/stcal/jump/jump.py
```diff
@@ -338,6 +344,12 @@ def detect_jumps(
             )
             log.info("Total showers= %i", num_showers)
             number_extended_events = num_showers
+
+        # This is where we look for micrometeorite flashes if the triggering
```
kmacdonald-stsci, line 348: Since this is done in both conditionals and done last, it should be moved out side the conditionals completely, instead of invoked twice.

The `expand_large_events` and `find_showers` can be moved outside the conditionals for the same reason.

src/stcal/jump/jump.py
```diff
@@ -338,6 +344,12 @@ def detect_jumps(
...
+        # threshold is less than the entire array
+        if (mmflashfrac < 1):
```
kmacdonald-stsci, line 350: Since this is done in both conditionals and done last, it should be moved out side the conditionals completely, instead of invoked twice.

The `expand_large_events` and `find_showers` can be moved outside the conditionals for the same reason.

src/stcal/jump/jump.py
```diff
@@ -514,6 +532,21 @@ def detect_jumps(
     # Return the updated data quality arrays
     return gdq, pdq, total_primary_crs, number_extended_events, stddev
 
+# Find micrometeorite flashes that light up the entire array in a small number
+# of groups.  Do this by looking for cases where greater than mmflash_pct percent
+# of the array is flagged as 'JUMP_DET' and flag the entire array.
+def find_micrometeorite_flashes(gdq, jump_flag, mmflashfrac):
+    npix = gdq.shape[2]*gdq.shape[3] # Number of pixels in array
```
kmacdonald-stsci, line 539: Add spaces before and after `*`.

src/stcal/jump/jump.py
```diff
@@ -514,6 +532,21 @@ def detect_jumps(
...
+    # Loop over integrations
+    for ii in range(0,gdq.shape[0]):
```
kmacdonald-stsci, line 541: Add space after `,`.

src/stcal/jump/jump.py
```diff
@@ -514,6 +532,21 @@ def detect_jumps(
...
+        # Loop over groups
+        for jj in range(0,gdq.shape[1]):
```
kmacdonald-stsci, line 543: Add space after `,`.

src/stcal/jump/jump.py
```diff
@@ -514,6 +532,21 @@ def detect_jumps(
     # Return the updated data quality arrays
     return gdq, pdq, total_primary_crs, number_extended_events, stddev
 
+# Find micrometeorite flashes that light up the entire array in a small number
+# of groups.  Do this by looking for cases where greater than mmflash_pct percent
+# of the array is flagged as 'JUMP_DET' and flag the entire array.
+def find_micrometeorite_flashes(gdq, jump_flag, mmflashfrac):
```
kmacdonald-stsci, line 538: This is missing a docstring.

Also, start the function with `nints, ngroups, nrows, ncols = gdq.shape`, then use these more descriptive variable names, rather than `shape[k]`.

src/stcal/jump/jump.py
```diff
@@ -514,6 +532,21 @@ def detect_jumps(
...
+    npix = gdq.shape[2]*gdq.shape[3] # Number of pixels in array
+    # Loop over integrations
+    for ii in range(0,gdq.shape[0]):
+        # Loop over groups
+        for jj in range(0,gdq.shape[1]):
+            indx = np.where(gdq[ii,jj,:,:] & jump_flag != 0)
```
kmacdonald-stsci, line 544: Add spaces after all `,`.

src/stcal/jump/jump.py
```diff
@@ -514,6 +532,21 @@ def detect_jumps(
...
+            fraction = float(len(indx[0])) / npix
+            if (fraction > mmflashfrac):
+                gdq[ii,jj,:,:] = gdq[ii,jj,:,:] | jump_flag
```
kmacdonald-stsci, line 547: Add spaces after all `,`.

src/stcal/jump/jump.py
```diff
@@ -514,6 +532,21 @@ def detect_jumps(
...
+                log.info("Detected a micrometeorite flash in integ = %i, group= %i", ii, jj)
```
kmacdonald-stsci, line 548: Why not use an f-string?

kmacdonald-stsci (changes requested):
There is some code called in two separate places due to a conditional, but can be moved outside the conditionals and called once, since they are called in both the `if` and `else` portion of the conditional.

There is also some formatting and style issues.

src/stcal/jump/jump.py
drlaw1558: Done

src/stcal/jump/jump.py
drlaw1558: Done

src/stcal/jump/jump.py
drlaw1558: Done

src/stcal/jump/jump.py
drlaw1558: Done

src/stcal/jump/jump.py
drlaw1558: Good point- done

src/stcal/jump/jump.py
drlaw1558: Done

src/stcal/jump/jump.py
drlaw1558: Done

src/stcal/jump/jump.py
drlaw1558: If that's the convention I certainly can- done.

drlaw1558:
Things that remain to be determined:
- Check that this function successfully detects more example flashes from the FGS list
- Determine whether it needs to be restricted in any way, e.g. to only FULL array or non-TSO data
- Determine whether to set mmflashfrac=1.0 by default (i.e., step disabled unless explicitly enabled by param ref file), or to set mmflashfrac=0.1 (for example) to explicitly enable by default unless disabled by param ref file.

Work ongoing to determine these.

kmacdonald-stsci (approved):

drlaw1558:
Closing in favor of https://github.com/spacetelescope/stcal/pull/349, which ports these changes to the updated format of the jump step.

PR #310, opened 2024-10-17
Pull common parts of skymatch from romancal and jwst into stcal

WilliamJamieson:
This PR pulls the common parts of skymatch from romancal and jwst into stcal to be shared.

Note, no tests are included because it is well tested by the step tests in both romancal and jwst.

## Tasks

</details>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -43,6 +43,8 @@ intersphinx_mapping = {
         "https://spacetelescope-drizzle.readthedocs.io/en/latest/",
         None
     ),
+    "imagestats": ("https://stsciimagestats.readthedocs.io/en/latest/", None),
+    "spherical_geometry": ("https://spherical-geometry.readthedocs.io/en/latest/", None),
 }
 
 extensions = [
@@ -82,4 +84,7 @@ nitpicky = False  # True does not work with enumerated parameter values
 nitpick_ignore = [
     ("py:class", "optional"),
     ("py:class", "np.ndarray"),
+    ("py:class", "stsci.imagestats.ImageStats"),  # intersphinx isn't working here
+    ("py:class", "spherical_geometry.polygon.SphericalPolygon"), # intersphinx isn't working here
 ]
+
--- a/src/stcal/skymatch/skyimage.py
+++ b/src/stcal/skymatch/skyimage.py
@@ -379,6 +379,11 @@ class SkyImage:
             of the image are included automatically. If `stepsize` is `None`,
             bounding polygon will contain only vertices of the image.
 
+        Notes
+        -----
+        The bounding polygon is defined from corners of pixels whereas the pixel
+        coordinates refer to their centers and therefore the lower-left corner
+        is located at (-0.5, -0.5)
         """
         ny, nx = self.image_shape
 
@@ -451,14 +456,6 @@ class SkyImage:
             binwidth=binwidth
         )
 
-    # TODO: due to a bug in the sphere package, see
-    #       https://github.com/spacetelescope/sphere/issues/74
-    #       intersections with polygons formed as union does not work.
-    #       For this reason I re-implement 'calc_sky' below with
-    #       a workaround for the bug.
-    #       The original implementation (now called ``_calc_sky_orig``
-    #       should replace current 'calc_sky' once the bug is fixed.
-    #
     def calc_sky(self, overlap=None, delta=True):
         """
         Compute sky background value.
@@ -494,7 +491,15 @@ None, optional
         polyarea : float
             Area (in srad) of the polygon that bounds data used to compute
             sky statistics.
-
+    
+        Notes
+        -----
+        Due to a bug in the sphere package, see
+        https://github.com/spacetelescope/sphere/issues/74
+        intersections with polygons formed as union does not work.
+        For this reason I re-implement 'calc_sky' below with a workaround for
+        the bug. The original implementation should be used when the bug is
+        fixed.
         """
         if overlap is None:
 
@@ -581,109 +586,6 @@ None, optional
             skyval -= self.sky
 
         return skyval, npix, polyarea
-
-#     def _calc_sky_orig(self, overlap=None, delta=True):
-#         """
-#         Compute sky background value.
-#
-#         Parameters
-#         ----------
-#         overlap : SkyImage, SkyGroup, SphericalPolygon, list of tuples, \
-# None, optional
-#             Another `SkyImage`, `SkyGroup`,
-#             :py:class:`spherical_geometry.polygons.SphericalPolygon`, or
-#             a list of tuples of (RA, DEC) of vertices of a spherical
-#             polygon. This parameter is used to indicate that sky statistics
-#             should computed only in the region of intersection of *this*
-#             image with the polygon indicated by `overlap`. When `overlap` is
-#             `None`, sky statistics will be computed over the entire image.
-#
-#         delta : bool, optional
-#             Should this function return absolute sky value or the difference
-#             between the computed value and the value of the sky stored in the
-#             `sky` property.
-#
-#         Returns
-#         -------
-#         skyval : float, None
-#             Computed sky value (absolute or relative to the `sky` attribute).
-#             If there are no valid data to perform this computations (e.g.,
-#             because this image does not overlap with the image indicated by
-#             `overlap`), `skyval` will be set to `None`.
-#
-#         npix : int
-#             Number of pixels used to compute sky statistics.
-#
-#         polyarea : float
-#             Area (in srad) of the polygon that bounds data used to compute
-#             sky statistics.
-#
-#         """
-#
-#         if overlap is None:
-#
-#             if self._mask is None:
-#                 data = self.image
-#             else:
-#                 data = self.image[self._mask.get_data()]
-#
-#             polyarea = self.poly_area
-#
-#         else:
-#             fill_mask = np.zeros(self.image_shape, dtype=bool)
-#
-#             if isinstance(overlap, (SkyImage, SkyGroup, SphericalPolygon)):
-#                 intersection = self.intersection(overlap)
-#                 polyarea = np.fabs(intersection.area())
-#                 radec = intersection.to_radec()
-#
-#             else:  # assume a list of (ra, dec) tuples:
-#                 radec = []
-#                 polyarea = 0.0
-#                 for r, d in overlap:
-#                     poly = SphericalPolygon.from_radec(r, d)
-#                     polyarea1 = np.fabs(poly.area())
-#                     if polyarea1 == 0.0 or len(r) < 4:
-#                         continue
-#                     polyarea += polyarea1
-#                     radec.append(self.intersection(poly).to_radec())
-#
-#             if polyarea == 0.0:
-#                 return None, 0, 0.0
-#
-#             for ra, dec in radec:
-#                 if len(ra) < 4:
-#                     continue
-#
-#                 # set pixels in 'fill_mask' that are inside a polygon
-#                 # to True:
-#                 x, y = self.wcs_inv(ra, dec)
-#                 poly_vert = list(zip(*[x, y]))
-#
-#                 polygon = region.Polygon(True, poly_vert)
-#                 fill_mask = polygon.scan(fill_mask)
-#
-#             if self._mask is not None:
-#                 fill_mask &= self._mask.get_data()
-#
-#             data = self.image[fill_mask]
-#
-#             if data.size < 1:
-#                 return None, 0, 0.0
-#
-#         # Calculate sky
-#         try:
-#
-#             skyval, npix = self._skystat(data)
-#
-#         except ValueError:
-#
-#             return None, 0, 0.0
-#
-#         if delta:
-#             skyval -= self._sky
-#
-#         return skyval, npix, polyarea
 
     def copy(self):
         """
--- a/src/stcal/skymatch/skyimage.py
+++ b/src/stcal/skymatch/skyimage.py
@@ -1,0 +1,946 @@
+"""
+The ``skyimage`` module contains algorithms that are used by
+``skymatch`` to manage all of the information for footprints (image outlines)
+on the sky as well as perform useful operations on these outlines such as
+computing intersections and statistics in the overlap regions.
+
+:Authors: Mihai Cara (contact: help@stsci.edu)
+
+
+"""
+
+# STDLIB
+import abc
+import tempfile
+
+# THIRD-PARTY
+import numpy as np
+from gwcs import region
+from spherical_geometry.polygon import SphericalPolygon
+
+# LOCAL
+from . skystatistics import SkyStats
+
+
+__all__ = ['SkyImage', 'SkyGroup', 'DataAccessor', 'NDArrayInMemoryAccessor',
+           'NDArrayMappedAccessor']
+
+
+class DataAccessor(abc.ABC):
+    """ Base class for all data accessors. Provides a common interface to
+        access data.
+    """
+    @abc.abstractmethod
+    def get_data(self):  # pragma: no cover
+        pass
+
+    @abc.abstractmethod
+    def set_data(self, data):  # pragma: no cover
+        """ Sets data.
+
+        Parameters
+        ----------
+        data : numpy.ndarray
+            Data array to be set.
+
+        """
+        pass
+
+    @abc.abstractmethod
+    def get_data_shape(self):  # pragma: no cover
+        pass
+
+
+class NDArrayInMemoryAccessor(DataAccessor):
+    """ Accessor for in-memory `numpy.ndarray` data. """
+    def __init__(self, data):
+        super().__init__()
+        self._data = data
+
+    def get_data(self):
+        return self._data
+
+    def set_data(self, data):
+        self._data = data
+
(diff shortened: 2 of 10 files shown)


src/stcal/skymatch/skystatistics.py
```diff
@@ -0,0 +1,136 @@
+"""
+The `skystatistics` module provides statistics computation class used by
+:py:func:`~stcal.skymatch.skymatch.match`
+and :py:class:`~stcal.skymatch.skyimage.SkyImage`.
+"""
+
+from copy import deepcopy
+
+# THIRD PARTY
+from stsci.imagestats import ImageStats  # type: ignore[import-untyped]
```
emolter, line 10: might be worth considering making the relevant third-party type ignores global by putting them here https://github.com/spacetelescope/stcal/blob/d6a87c346b56daafcd6b8b43231fdbb8b9222965/pyproject.toml#L212

src/stcal/skymatch/skystatistics.py
```diff
@@ -0,0 +1,136 @@
...
+
+__all__ = ["SkyStats"]
+
+
+class SkyStats:
+    """
+    This is a superclass build on top of
+    :py:class:`stsci.imagestats.ImageStats`. Compared to
```
emolter, line 18: This comment is confusing because the class doesn't actually inherit from `ImageStats`

WilliamJamieson:
> This looks great, thanks William! I just had a few questions about whether some of the ruff and mypy problems could be fixed instead of ignored.
> 
> I also had a few questions about the code itself, but those also exist on jwst main, so they're technically beyond the scope of this PR. But I'm still curious to know the answers to those and I figure now is as good a time as any to ensure the code is written as cleanly as possible

I agree they should be fixed. However, as you said they are beyond the scope of this PR. I think for making it possible to trace the history of changes it makes sense to move the code in as close a state as possible from JWST to stcal and then do a follow on PR making cleanups and addressing the issues uncovered here.

src/stcal/skymatch/skyimage.py
```diff
@@ -0,0 +1,1020 @@
...
+        npty = ys.size
+
+        npts = 2 * (nptx + npty)
+
+        borderx = np.empty((npts + 1,), dtype=float)
+        bordery = np.empty((npts + 1,), dtype=float)
+
+        # "bottom" points:
+        borderx[:nptx] = xs
+        bordery[:nptx] = -0.5
```
kmacdonald-stsci, line 454: I suggest a comment explaining the -0.5 number or the use of a descriptive constant detailing what this number is.

kmacdonald-stsci (approved):

src/stcal/skymatch/skyimage.py
WilliamJamieson: I agree, and I believe this has something to do with all the pixel offsetting. However this is largely just a copy of the stuff from JWST with minimal changes. If someone knows exactly what this is I would be happy to add a comment here.

Really, we should be maintaining a common "constants" object/module to document these and back reference that as opposed to having a bunch of ad hoc magic numbers some of which are documented with comments. This however, is something to consider later.

emolter:
> I agree they should be fixed. However, as you said they are beyond the scope of this PR.

IMO making the code adhere to the more stringent style rules in stcal remains within the scope.  I don't much like ignoring them.  I think the downstream testing should assuage any concerns that those changes might break something.

I agree that my comment RE the potentially unnecessary use of `@property` can be ignored for now though.

src/stcal/skymatch/skyimage.py
mcara: Centers of pixels have integer image coordinates. Edges of the pixels are located `+/- 0.5` pixels away from centers. I think this convention is explained somewhere in the docs. IMO, it does not make sense to add a comment like this every single time we see a 0.5 in the code.

src/stcal/skymatch/skyimage.py
```diff
@@ -0,0 +1,1020 @@
...
+
+    def __init__(
+        self,
+        image,
+        wcs_fwd,
+        wcs_inv,
+        pix_area=1.0,
+        convf=1.0,
+        mask=None,
+        id=None,  # noqa: A002
```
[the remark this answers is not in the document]
mcara: probably could be changed to `image_id`???

src/stcal/skymatch/skyimage.py
```diff
@@ -0,0 +1,1020 @@
...
+
+        convf : float, optional
+            Conversion factor that when multiplied to `image` data converts
+            the data to "uniform" (across multiple images) surface
+            brightness units.
+
+            .. note::
+
+              The functionality to support this conversion is not yet
+              implemented and at this moment `convf` is ignored.
```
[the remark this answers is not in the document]
mcara: Based on ancient (>9 years old) discussions with relevant people, this feature, _in theory_, may not be needed for JWST. _In practice_, however, we don't know. I suggest leaving this alone as is.

src/stcal/skymatch/skyimage.py
```diff
@@ -0,0 +1,1020 @@
...
+
+    @property
+    def image_shape(self):
+        """Get `SkyImage`'s ``image`` data shape."""
+        if self._image_shape is None and self._image is not None:
+            self._image_shape = self._image.get_data_shape()
+        return self._image_shape
+
+    @property
+    def id(self):  # noqa: A003
```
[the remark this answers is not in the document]
mcara: Since the original code was written ~11-12 years ago (not sure about this but I think so), I no longer remember the intention. Maybe during the development I thought that, for example, `pix_area` will be computed on demand but ended up storing it for efficiency. It just allows more flexibility. Also, since both `SkyImage` and `SkyGroup` can be used as inputs to `match`, they should have  a common set of properties. I do not remember what was the Python version used during the development but it may have been `2.x` (due to PyRAF / Tables/ whatever limitations) so I don't think `ABC` was available at that time and even if it was, I am not sure properties or attributes of classes could be made abstract. Another reason may have been the desire to get these properties documented (in the docstrings).

I do not think anything should be changed in this PR but in the future they could subclass from a `BaseSky`, `ABCSky` or something like that.

I do not see what is the harm in using properties.

src/stcal/skymatch/skyimage.py
mcara: and `-0.5` is shorter than `coordinate_of_left_pixel_border(_relative_to_center)`

src/stcal/skymatch/skyimage.py
```diff
@@ -0,0 +1,1020 @@
...
+    """
+    Holds multiple :py:class:`SkyImage` objects whose sky background values
+    must be adjusted together.
+
+    `SkyGroup` provides methods for obtaining bounding polygon of the group
+    of :py:class:`SkyImage` objects and to compute sky value of the group.
+
+    """
+
+    def __init__(self, images, id=None, sky=0.0):  # noqa: A002
```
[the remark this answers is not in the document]
mcara: I suggest `sky_id` as it would work with both `SkyImage` and `SkyGroup`.

src/stcal/skymatch/skymatch.py
```diff
@@ -0,0 +1,539 @@
...
+
+__all__ = ["match"]
+
+__local_debug__ = True
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+
+def match(images, skymethod="global+match", match_down=True, subtract=False):
```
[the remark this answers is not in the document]
mcara: I would suggest something along the lines `do_match` or `perform_match`, or `match_sky`. The intent for the name of this function was for it to be a verb. `skymatch` does not sound to me (a non-native English speaker) as a verb but I may be wrong.

src/stcal/skymatch/skyimage.py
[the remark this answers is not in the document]
mcara: I suggest `sky_id` as it would work with both `SkyImage` and `SkyGroup`.

src/stcal/skymatch/skystatistics.py
mcara: Docs could be changed to explain that `SkyStats` delegates calls to an `ImageStats` object if that is more clear but the comment about persistence of settings is important as it is what differentiates it from `ImageStats` and I think it helps with performance.

src/stcal/skymatch/skyimage.py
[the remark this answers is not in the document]
emolter: no harm it's just unnecessarily verbose since neither the getter nor the setter does anything.  It's therefore confusing, because when I see those methods, I expect them to do something.  But it doesn't matter- I agree with William's comment that these changes shouldn't go into this PR, if they happen at all.

src/stcal/skymatch/skymatch.py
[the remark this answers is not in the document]
emolter: I'm happy with whatever name

src/stcal/skymatch/skystatistics.py
emolter: I agree that the comment itself explains an important distinction, and yes the simple wording change from "superclass built on top of" to "delegates calls to" would be clearer

src/stcal/skymatch/skyimage.py
emolter: I agree with Mihai that in the context of pixels it's relatively clear what a 0.5 offset is doing. 
 However it looks to me like `calc_bounding_polygon` is the only function where this is really relevant, so would it work to add a note to the docstring saying something like,
"""
Notes
-----
The bounding polygon is defined from corners of pixels
whereas the reference pixel is defined at the center of a pixel
and therefore the lower-left corner is located at (-0.5, -0.5)
"""

or whatever you want, I'm sure I made some small inaccuracies in describing that but you get the idea

mcara: "whereas the reference pixel" (in @emolter suggestion) should be something like "whereas the pixel coordinates refer to their centers" (or something like that). "Reference pixel" has a special meaning in the WCS world.

src/stcal/skymatch/skyimage.py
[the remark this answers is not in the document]
WilliamJamieson: This is actually being used in places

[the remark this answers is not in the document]
emolter: Can you please point me to where?  If that's the case, this comment in the docstring should be removed

src/stcal/skymatch/skyimage.py
emolter: I see that this was resolved but I don't see any docstring update, is there any objection to my previous suggestion?

src/stcal/skymatch/skyimage.py
```diff
@@ -0,0 +1,946 @@
...
+        )
+
+    # TODO: due to a bug in the sphere package, see
+    #       https://github.com/spacetelescope/sphere/issues/74
+    #       intersections with polygons formed as union does not work.
+    #       For this reason I re-implement 'calc_sky' below with
+    #       a workaround for the bug.
+    #       The original implementation (now called ``_calc_sky_orig``
+    #       should replace current 'calc_sky' once the bug is fixed.
+    #
```
emolter, line 461: Can this be moved into a Note for the `calc_sky` function?

src/stcal/skymatch/skyimage.py
```diff
@@ -0,0 +1,946 @@
...
+
+        if not np.isfinite(skyval):
+            return None, 0, 0.0
+
+        if delta:
+            skyval -= self.sky
+
+        return skyval, npix, polyarea
+
+#     def _calc_sky_orig(self, overlap=None, delta=True):
```
emolter, line 585: remove instead of commenting out

src/stcal/skymatch/skyimage.py
WilliamJamieson: Oops I copied over that with something, I'll re-add this

emolter (changes requested):
Thanks for all the updates!  Before merging I think it would be good to update https://github.com/spacetelescope/jwst/pull/8901 such that we can see if the changes are likely to work well with jwst.  Perhaps you were already planning to do so.

src/stcal/skymatch/skyimage.py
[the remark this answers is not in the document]
WilliamJamieson: Thats all logging but I'm not qualified to tell if we want to keep any parts of those logs

WilliamJamieson:
@emolter I believe this is ready now.

docs/conf.py
```diff
@@ -39,9 +39,16 @@
...
+    "imagestats": ("https://stsciimagestats.readthedocs.io/en/latest/", None),
+    "spherical_geometry": ("https://spherical-geometry.readthedocs.io/en/latest/", None),
 }
 
-nitpick_ignore = [("py:class", "optional"), ("py:class", "np.ndarray")]
+nitpick_ignore = [
+    ("py:class", "optional"),
+    ("py:class", "np.ndarray"),
+    ("py:class", "stsci.imagestats.ImageStats"),  # intersphinx isn't working here
+    ("py:class", "spherical_geometry.polygon.SphericalPolygon"), # intersphinx isn't working here
```
WilliamJamieson, line 50: These should be working with the intersphinx links I added. I believe the fix may require some fixes in the documentation for those packages.

It is possible something is off in the `stcal` docs causing the link failure. Please feel free to push changes to attempt to fix this.

src/stcal/skymatch/skyimage.py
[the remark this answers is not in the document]
emolter: I don't think "used in logging" qualifies as being used given that `convf` is hard-coded to equal 1.0 and therefore has no effect on the scaling reported in the log messages.  So I agree with Mihai we should leave this as-is, including the docstring

src/stcal/skymatch/skyimage.py
emolter: thanks!

emolter (commented):
All my previous questions have been answered now.

I don't think the changes have been so large that it would be helpful for me to go through it again, but let me know if you'd like me to.

So this looks good to me pending fixing the bug that is causing the failing jwst regtests

mcara:
It would still be useful to have some unit tests developed for this. We have always relied on regression tests (both for HST and JWST) but PRs to this code in the future should be able to use some quick unit tests. It could be done either as part of this PR or as a follow-up PR.

mcara (approved):

WilliamJamieson:
> It would still be useful to have some unit tests developed for this. We have always relied on regression tests (both for HST and JWST) but PRs to this code in the future should be able to use some quick unit tests. It could be done either as part of this PR or as a follow-up PR.

I think this should be a follow on PR. I looked into moving some of the unit tests from JWST. But at this time all those tests directly involve running the entire skymatch step. I personally don't feel like I can write any useful unit tests for this part of skymatch independent of JWST at this time.

emolter:
Thanks @nden, that's fine with me. As I said, I'll only look at this again if needed, @WilliamJamieson I see you re-requested my review after @nden dismissed the stale one.  Does that mean you'd like me to look again, or would you just like me to hit the Approve button?

WilliamJamieson:
> Thanks @nden, that's fine with me. As I said, I'll only look at this again if needed, @WilliamJamieson I see you re-requested my review after @nden dismissed the stale one. Does that mean you'd like me to look again, or would you just like me to hit the Approve button?

If you want to do another full review your welcome, though there hasn't been any substantive changes since I addressed your last one. In any case, in order for me to merge this PR I do need two approving reviews.

nden:
@schlafly @ddavis-stsci This is ready to be merged from JWSt point of view. Romancal unit tests are passing and a [PR using this code is in the works for Roman.](https://github.com/spacetelescope/romancal/pull/1465). Please review when available.

emolter (approved):

tapastro:
This was waiting for a Roman team member to approve - in general, stcal PRs require approvals not just from any two people, but one from jwst and one from Roman. This still needs eyes from the Roman team. 

We can make this procedure more clear to the branch.

WilliamJamieson:
> This was waiting for a Roman team member to approve - in general, stcal PRs require approvals not just from any two people, but one from jwst and one from Roman. This still needs eyes from the Roman team.
> 
> We can make this procedure more clear to the branch.

I have been tracking this for both JWST and RCAL. JWST was the main sticking point with RCAL waiting for this to go in.

tapastro:
Eddie, the Romancal product owner, listed himself and Dave Davis as approved reviewers for stcal PRs. He is free to add you to that list, but I have not heard of an update there and am trying to follow protocol so that we do not merge changes that Romancal is not aware of.

schlafly:
Thanks.  I'm happy for this to go in.

I'll note that I got confused trying to understand how the "self.sky" feature was supposed to function in the SkyGroups, and where or how the image.backgrounds are used.
https://github.com/spacetelescope/stcal/pull/310/files#diff-1d5ed0c79f81a6487ae90d08de3bf1b263bb11b97aa1aee2175b28918a78f7b3R731
https://github.com/spacetelescope/stcal/pull/310/files#diff-1d5ed0c79f81a6487ae90d08de3bf1b263bb11b97aa1aee2175b28918a78f7b3R813

I'm sure I just missed it, though, and presumably it is duplicating existing code.

PR #311, opened 2024-10-18
JP-3683: fix abs_deriv handling of off-edge and nan values

braingram:
Resolves [JP-3683](https://jira.stsci.edu/browse/JP-3683)

As part of outlier detection an "absolute derivative" is computed. From the [outlier detection docs](https://jwst-pipeline.readthedocs.io/en/latest/jwst/outlier_detection/outlier_detection_imaging.html):
```
The derivative of the blotted image gets created using the blotted median image to compute the absolute value of the difference between each pixel and its four surrounding neighbors with the largest value being the recorded derivative.
So in this case the "absolute derivative" is the maximum absolute difference between a pixel and it's neighbors.
```

This PR addresses 2 issues with `_abs_deriv` (used in outlier detection) where it:
- treated off-edge pixels as 0
- included `nan` values in the `max` used for combining differences between adjacent pixels (effectively dilating nan values)

With the previous code:
```python
>> arr = np.arange(25, dtype='f4').reshape((5, 5))
>> arr[2, 2] = np.nan
>> _abs_deriv(arr)
array([[ 5.,  5.,  5.,  5.,  5.],
       [ 5.,  5., nan,  5.,  9.],
       [10., nan, nan, nan, 14.],
       [15.,  5., nan,  5., 19.],
       [20., 21., 22., 23., 24.]])
```
Note that now instead of one nan at (2, 2) there are 5 nan pixels and that edge values are larger than expected. For example (1, 4) is 9 when the largest difference with all defined pixels is 5.

With this PR the result is:
```python
array([[ 5.,  5.,  5.,  5.,  5.],
       [ 5.,  5.,  5.,  5.,  5.],
       [ 5.,  5., nan,  5.,  5.],
       [ 5.,  5.,  5.,  5.,  5.],
       [ 5.,  5.,  5.,  5.,  5.]], dtype=float32)
``` 

Tests are added for both issues.

It is expected this will change regression tests for both jwst and romancal as `abs_deriv` is used to compute the outlier detection threshold. With this PR more outliers will be detected (since the threshold will be lower due to edge pixels having lower values and less nans in the blot images).

Romancal regtests running: https://github.com/spacetelescope/RegressionTests/actions/runs/11682867663
JWST regtests running: https://github.com/spacetelescope/RegressionTests/actions/runs/11682842450

## Tasks

</details>

### DIFF
--- a/src/stcal/outlier_detection/utils.py
+++ b/src/stcal/outlier_detection/utils.py
@@ -93,21 +93,22 @@ def _abs_deriv(array):
         out[np.isnan(array)] = np.nan
 
     # compute row-wise absolute diffference
-    d = np.abs(np.diff(array, axis=0))
-    np.putmask(out[1:], np.isfinite(d), d)  # no need to do max yet
+    row_diff = np.abs(np.diff(array, axis=0))
+    np.putmask(out[1:], np.isfinite(row_diff), row_diff)  # no need to do max yet
 
     # since these are absolute differences |r0-r1| = |r1-r0|
     # make a view of the target portion of the array
-    v = out[:-1]
+    row_offset_view = out[:-1]
     # compute an in-place maximum
-    np.putmask(v, d > v, d)
+    np.putmask(row_offset_view, row_diff > row_offset_view, row_diff)
+    del row_diff
 
     # compute col-wise absolute difference
-    d = np.abs(np.diff(array, axis=1))
-    v = out[:, 1:]
-    np.putmask(v, d > v, d)
-    v = out[:, :-1]
-    np.putmask(v, d > v, d)
+    col_diff = np.abs(np.diff(array, axis=1))
+    col_offset_view = out[:, 1:]
+    np.putmask(col_offset_view, col_diff > col_offset_view, col_diff)
+    col_offset_view = out[:, :-1]
+    np.putmask(col_offset_view, col_diff > col_offset_view, col_diff)
     return out
 
 
--- a/src/stcal/outlier_detection/utils.py
+++ b/src/stcal/outlier_detection/utils.py
@@ -85,41 +85,30 @@ def compute_weight_threshold(weight, maskpt):
 
 
 def _abs_deriv(array):
-    """
-    Do not use this function.
+    """Take the absolute derivate of a numpy array."""
+    out = np.zeros_like(array)  # use same dtype as input
 
-    Take the absolute derivative of a numpy array.
+    # make output values nan where input is nan (for floating point input)
+    if np.issubdtype(array.dtype, np.floating):
+        out[np.isnan(array)] = np.nan
 
-    This function assumes off-edge pixel values are 0
-    and leads to erroneous derivative values and should
-    likely not be used.
-    """
-    tmp = np.zeros(array.shape, dtype=np.float64)
-    out = np.zeros(array.shape, dtype=np.float64)
+    # compute row-wise absolute diffference
+    d = np.abs(np.diff(array, axis=0))
+    np.putmask(out[1:], np.isfinite(d), d)  # no need to do max yet
 
-    tmp[1:, :] = array[:-1, :]
-    tmp, out = _absolute_subtract(array, tmp, out)
-    tmp[:-1, :] = array[1:, :]
-    tmp, out = _absolute_subtract(array, tmp, out)
+    # since these are absolute differences |r0-r1| = |r1-r0|
+    # make a view of the target portion of the array
+    v = out[:-1]
+    # compute an in-place maximum
+    np.putmask(v, d > v, d)
 
-    tmp[:, 1:] = array[:, :-1]
-    tmp, out = _absolute_subtract(array, tmp, out)
-    tmp[:, :-1] = array[:, 1:]
-    tmp, out = _absolute_subtract(array, tmp, out)
-
+    # compute col-wise absolute difference
+    d = np.abs(np.diff(array, axis=1))
+    v = out[:, 1:]
+    np.putmask(v, d > v, d)
+    v = out[:, :-1]
+    np.putmask(v, d > v, d)
     return out
-
-
-def _absolute_subtract(array, tmp, out):
-    """
-    Do not use this function.
-
-    A helper function for _abs_deriv.
-    """
-    tmp = np.abs(array - tmp)
-    out = np.maximum(tmp, out)
-    tmp = tmp * 0.
-    return tmp, out
 
 
 def flag_crs(
--- a/changes/311.bugfix.rst
+++ b/changes/311.bugfix.rst
@@ -1,0 +1,1 @@
+Fix abs_deriv handling of off-edge and nan values.
--- a/tests/outlier_detection/test_utils.py
+++ b/tests/outlier_detection/test_utils.py
@@ -33,12 +33,21 @@ def test_abs_deriv_single_value(shape, diff):
     np.testing.assert_allclose(result, expected)
 
 
-@pytest.mark.skip(reason="_abs_deriv has edge effects due to treating off-edge pixels as 0: see JP-3683")
 @pytest.mark.parametrize("nrows,ncols", [(5, 5), (7, 11), (17, 13)])
 def test_abs_deriv_range(nrows, ncols):
     arr = np.arange(nrows * ncols).reshape(nrows, ncols)
     result = _abs_deriv(arr)
     np.testing.assert_allclose(result, ncols)
+
+
+def test_abs_deriv_nan():
+    arr = np.arange(25, dtype='f4').reshape(5, 5)
+    arr[2, 2] = np.nan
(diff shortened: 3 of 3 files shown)


src/stcal/outlier_detection/utils.py
```diff
@@ -85,43 +85,32 @@ def compute_weight_threshold(weight, maskpt):
...
-
+    """Take the absolute derivate of a numpy array."""
+    out = np.zeros_like(array)  # use same dtype as input
+
+    # make output values nan where input is nan (for floating point input)
+    if np.issubdtype(array.dtype, np.floating):
+        out[np.isnan(array)] = np.nan
+
+    # compute row-wise absolute diffference
+    d = np.abs(np.diff(array, axis=0))
```
kmacdonald-stsci, line 96: Please avoid using one letter variable names (one reason is that when running `ipdb` the letter 'd' has a meaning within that framework).  Maybe `row_diff` for this variable.

src/stcal/outlier_detection/utils.py
```diff
@@ -85,43 +85,32 @@ def compute_weight_threshold(weight, maskpt):
...
+    np.putmask(out[1:], np.isfinite(d), d)  # no need to do max yet
+
+    # since these are absolute differences |r0-r1| = |r1-r0|
+    # make a view of the target portion of the array
+    v = out[:-1]
```
kmacdonald-stsci, line 101: Please don't use one letter variable names.  Maybe `row_view` for this.

tapastro:
Looks like the RT run didn't recognize the provided branch name (in addition to the gcleary issue)

src/stcal/outlier_detection/utils.py
```diff
@@ -85,43 +85,32 @@ def compute_weight_threshold(weight, maskpt):
...
+    # compute an in-place maximum
+    np.putmask(v, d > v, d)
+
+    # compute col-wise absolute difference
+    d = np.abs(np.diff(array, axis=1))
```
kmacdonald-stsci, line 106: Maybe call this `col_diff`.

src/stcal/outlier_detection/utils.py
```diff
@@ -85,43 +85,32 @@ def compute_weight_threshold(weight, maskpt):
...
+    v = out[:, 1:]
```
kmacdonald-stsci, line 107: Maybe call this `col_view`.

kmacdonald-stsci (approved):
I recommend avoid using single letter variable names and using more descriptive variable names.

braingram:
> Looks like the RT run didn't recognize the provided branch name (in addition to the gcleary issue)

Oof. Yeah I ran into "issues" :) The link should be updated to the correct one now:
https://github.com/spacetelescope/RegressionTests/actions/runs/11408938058
It shows 20 failures. These will be easier to sort through (and okify) once the artifactory issue is sorted (and these are re-run).

src/stcal/outlier_detection/utils.py
braingram: Thanks! How does 70f4045 look for variable renaming?

braingram:
A new set of regression tests now that artifactory is back up:
romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/11516702151
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/11516699749

braingram:
Romancal regtest shows 1 unrelated error and 2 failures:
- test_mos_pipeline
- test_mos_skycell_pipeline

That are expected since those outputs are different since this PR changes the number of detected outliers.

braingram:
JWST regtests show several expected differences due to this PR changing the number of outlier detected. Most differences are small (<1%) However some are several % with test_nirspec_mos_fs_spec3 being one of the larger (>5%). Looking at the first slit the difference is largely due to the "nan dilation" on main preventing a large number of pixels from being flagged as outliers.

On main the first slit has 7447 outliers whereas with this PR there are 8171. I make no claim to understand why the error data for this slit has only 27% finite values (when the science data is 59% finite). Looking at a portion of the slit with some finite pixels for data (streched to be able to see some structure in the finite pixels)
sci_stretched
and for err:
err
the computed "blot_deriv" (the output of "_abs_deriv") in this area is (top this PR, bottom main)
blot_deriv_0
Since any pixel in the "blot_deriv" marked as nan will never be flagged as an outlier the last plot shows the substantial number of pixels that on main are not being considered for outlier detection. With this PR these are considered and many for this example are now being flagged.

melanieclarke:
> On main the first slit has 7447 outliers whereas with this PR there are 8171. For outlier detection the science data contains 59% finite values whereas the error (computed as the blotted median following [spacetelescope/jwst#8828](https://github.com/spacetelescope/jwst/pull/8828)) contains 27% finite values. Looking at a portion of the slit with some finite pixels for data (streched to be able to see some structure in the finite pixels)

@braingram - I'd like to look into this. Which slit was this? And which product is in the screenshots?

mcara:
> `Do not use this function.`

😆

braingram:
> @braingram - I'd like to look into this. Which slit was this? And which product is in the screenshots?

Thanks!

The 7447 vs 8171 outliers are for one of the `crf` files compared for [test_nirspec_mos_fs_spec3](https://github.com/spacetelescope/jwst/blob/bf7cf8915c2241ed978c9aca546e0a06201c3855/jwst/regtest/test_nirspec_mos_fs_spec3.py#L25)
https://bytesalad.stsci.edu/ui/native/jwst-pipeline-results/2024-10-25_GITHUB_CI_Linux-X64-py3.12-850/2994_test_nirspec_mos_fs_spec3/jw02674-o004_b000000048_nirspec_f290lp-g395m_crf.fits
vs
https://bytesalad.stsci.edu/ui/native/jwst-pipeline/dev/truth/test_nirspec_mos_fs_spec3/jw02674-o004_b000000048_nirspec_f290lp-g395m_crf.fits

It looks like I left off the `mos` part for generating the plots. The plots are of the first slit processed for [test_nirspec_fs_spec3](https://github.com/spacetelescope/jwst/blob/bf7cf8915c2241ed978c9aca546e0a06201c3855/jwst/regtest/test_nirspec_fs_spec3.py#L34) (taken from `jw01309022001_04102_00001_nrs2_cal.fits`).
On main 24 pixels are flagged as outliers, with this PR 33 are flagged.
Here's a plot of the 'sci' and 'err' (the blotted median error) for that slit showing the large number of nan pixels (likely due to the masking during the median calculation, similar to the masking done for the 'sci' median used to generate the 'blot' array).
Screenshot 2024-10-28 at 11 00 58 AM

melanieclarke:
Hmm, I can't reproduce those images, working from jwst on main and stcal on your branch.  The below is what I see for the NRS2 blot file from S200A1.  Top is SCI, bottom is ERR.

The large area marked NaN at the right is expected: this is a consequence of the default parameters weight_type = 'ivm' and maskpt = 0.70.  NIRSpec is aware of this issue and considering different default parameter combinations -- see [JP-3347](https://jira.stsci.edu/browse/JP-3347) for discussion.

It is not expected that there would be a significant difference between the number of NaN values in the SCI and ERR arrays.  In quick spot checks for the regression test outputs for MOS, MOS/FS, and FS, I haven't found any blot or median products that show a NaN discrepancy -- they all look like the below image.

jw01309022001_04102_00001_nrs1_s200a1_blot

melanieclarke:
> The 'sci' I was referring to was the 'sci_data' passed into "flag_resampled_crs"

Ah!  Okay, I understand now.  In that case, yes, this is expected.  Blotted median sci and err will match up.  The input data will not - the extra NaNs in the blotted product are due to the extra masking in that product.

Either way, I agree, that's a side issue from the fixes for the abs_deriv function.  The changes here to fix the extra NaN dilation in the blot deriv look good.

melanieclarke (approved):
I looked through the differences for the regression tests for NIRSpec MOS and FS, and it looks like where the changes are significant with this branch, they are improvements.  In particular, discrepant edge values in the test_nirspec_mos_fs_spec3 input are better flagged with these changes.

Thanks for the fix!

braingram:
> I looked through the differences for the regression tests for NIRSpec MOS and FS, and it looks like where the changes are significant with this branch, they are improvements. In particular, discrepant edge values in the test_nirspec_mos_fs_spec3 input are better flagged with these changes.
> 
> Thanks for the fix!

Thanks for the review! Let me check with the romancal folks to see how they would like this PR scheduled with their release (since it impacts a few regression tests there as well).

braingram:
FYI merging of this should be delayed until:
- a review from a roman representative (@schlafly and @mairanteodoro are requested reviewers)
- ongoing work in romancal related to L1/L2 refacotring is finished

schlafly (approved):
LGTM, thanks!

mairanteodoro (approved):
Looks good! Thanks, @braingram!

PR #312, opened 2024-10-18
AL-875: Add memory saving options to compute_weight_threshold sigma_clip call

emolter:
Resolves [AL-875](https://jira.stsci.edu/browse/AL-875)

This PR adds memory-saving options to a call to astropy `sigma clip` within `compute_weight_threshold` in an attempt to decrease the overall memory usage of outlier detection step (see flamegraphs on [JP-3685](https://jira.stsci.edu/browse/JP-3685)), which cause that function to modify the input array in-place instead of copying it and allocating extra arrays for masks.  Since the output sigma-clipped array is immediately collapsed into its mean, there is no need to retain it as a masked array and the behavior is unchanged.

Note that the "masked=False" option to sigma_clip does not affect that function's processing of _input_ masked arrays, it only specifies whether the _output_ is a masked array with clipped pixels masked, or a simple np.array with clipped pixels removed.

## Tasks

</details>

### DIFF
--- a/src/stcal/outlier_detection/utils.py
+++ b/src/stcal/outlier_detection/utils.py
@@ -77,7 +77,11 @@ def compute_weight_threshold(weight, maskpt):
     weight_masked = np.ma.array(weight, mask=np.logical_or(
         mask_zero_weight, mask_nans))
     # Sigma-clip the unmasked data
-    weight_masked = sigma_clip(weight_masked, sigma=3, maxiters=5)
+    weight_masked = sigma_clip(weight_masked,
+                               sigma=3,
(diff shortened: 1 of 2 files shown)


emolter:
I'm pretty confident this will not change anything downstream, as the unit tests for the changed function appear to be relatively robust, but I started some downstream [JWST regression tests](https://github.com/spacetelescope/RegressionTests/actions/runs/11409484900) here anyway

mairanteodoro (approved):
Looks good to me!

kmacdonald-stsci (approved):

PR #315, opened 2024-11-01
Revert "AL-875: Add memory saving options to compute_weight_threshold sigma_clip call"

braingram:
Reverts spacetelescope/stcal#312

The reverted PR is causing regression test failures due in part to np.mean on a masked float32 array (which was the pre-PR behavior) producing a different result compared to a np.mean on a float32 array.

### DIFF
--- a/src/stcal/outlier_detection/utils.py
+++ b/src/stcal/outlier_detection/utils.py
@@ -77,12 +77,7 @@ def compute_weight_threshold(weight, maskpt):
     weight_masked = np.ma.array(weight, mask=np.logical_or(
         mask_zero_weight, mask_nans))
     # Sigma-clip the unmasked data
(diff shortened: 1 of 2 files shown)


emolter (approved):
This PR is causing a failure in a single pixel of a _median product on the JWST side, which is an intermediate product.  My opinion is there is no need to revert this based on the JWST regression tests alone.

I'm approving this anyway because it sounds like the Romancal team is still not sure whether the two failing regression tests are a cause for concern, and they'd like this not to be merged until they can figure that out.

schlafly (approved):
Approving.  FWIW, I don't think this is a major issue on the Roman side either, and I'm not pushing hard to remove this change on the basis of the romancal regression tests.  We could go ahead and okify those.  But I understand that we didn't intend to cause changes due to this PR and so may want to revisit it.

PR #319, opened 2024-11-14
Weight threshold part duex

braingram:
This is "take 2" of https://github.com/spacetelescope/stcal/pull/312

At the moment this PR will create small numerical differences in the output of `compute_weight_threshold` but I'm convinced that the current behavior on main is incorrect and a by-product of related numpy bugs:
- https://github.com/numpy/numpy/issues/14462
- https://github.com/numpy/numpy/issues/25548
- https://github.com/numpy/numpy/issues/12313

A minimal example is:
```python
import numpy as np

one_f = np.float32(1)
s_f = np.spacing(one_f)
arr_f = np.array([one_f, one_f + s_f, one_f + s_f + s_f], dtype=np.float32)

one_d = np.float64(1)
arr_d = np.array([one_d, one_d + s_f, one_d + s_f + s_f], dtype=np.float64)

print(f"expected mean: {np.mean(arr_d)=}")
print(f"{np.mean(np.ma.array(arr_f, mask=False))=}")
print(f"{np.mean(arr_f, dtype='f8')=}")
```
Produces:
```
expected mean: np.mean(arr_d)=1.0000001192092896
np.mean(np.ma.array(arr_f, mask=False))=1.0000000794728596
np.mean(arr_f, dtype='f8')=1.0000001192092896
```
The second line is similar to what's occuring on main where a mean of a masked array of float32 values is producing a float64 that is incorrect (given the precision). The third line is similar to what is produced with this PR.

One added benefit is this PR appears to reduce the memory usage lower than #312 (as seen by the `0.9` vs `1.9` multiplier in the memory usage test). I believe this is due to providing a non-masked array to sigma clip to avoid the allocation here:
https://github.com/astropy/astropy/blob/3bd8b404bf558561da469128d570280f3b65e57f/astropy/stats/sigma_clipping.py#L400-L401

Further optimizations might require changes to sigma clip (which appears to create an array copy for `_compute_bounds`:
Screenshot 2024-11-14 at 1 34 03 PM

Regression tests running:
Romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/11956099949
JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/11956106816
(these are expected to have a few minor differences)

## Tasks

</details>

### DIFF
--- a/changes/319.bugfix.rst
+++ b/changes/319.bugfix.rst
@@ -1,0 +1,1 @@
+Update weight threshold calculation in outlier detection to work around numpy bug that introduces small numerical differences for a mean of a masked array.
--- a/src/stcal/outlier_detection/utils.py
+++ b/src/stcal/outlier_detection/utils.py
@@ -68,25 +68,15 @@ def compute_weight_threshold(weight, maskpt):
     float
         The weight threshold for this integration.
     '''
-    # necessary in order to assure that mask gets applied correctly
-    if hasattr(weight, '_mask'):
-        del weight._mask
-    mask_zero_weight = np.equal(weight, 0.)
-    mask_nans = np.isnan(weight)
-    # Combine the masks
-    weight_masked = np.ma.array(weight, mask=np.logical_or(
-        mask_zero_weight, mask_nans))
-    # Sigma-clip the unmasked data
-    weight_masked = sigma_clip(weight_masked,
-                               sigma=3,
-                               maxiters=5,
-                               masked=False,
-                               copy=False,
-                               )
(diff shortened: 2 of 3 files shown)


kmacdonald-stsci (approved):

schlafly (approved):

mairanteodoro (approved):
Looks good to me!

emolter (approved):
Looks good.  I re-ran the memory usage profiling that I used for the original PR and it shows that the memory savings are indeed working as intended.

PR #320, opened 2024-11-27
Add common resample code to stcal.

mcara:
Closes #279

This PR add the common resample code used by both JWST and Roman pipelines to stcal. Also, for the first time, this PR adopts the new `drizzle` API from https://github.com/spacetelescope/drizzle/pull/134 for the resample code used in the pipelines. This PR is a second attempt to #279 and therefore merging this PR should close #279.

This work is related to https://jira.stsci.edu/browse/AL-835

## Tasks

</details>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -39,6 +39,10 @@ intersphinx_mapping = {
     "gwcs": ("https://gwcs.readthedocs.io/en/latest/", None),
     "astropy": ("https://docs.astropy.org/en/stable/", None),
     "tweakwcs": ("https://tweakwcs.readthedocs.io/en/latest/", None),
+    "drizzle": (
+        "https://spacetelescope-drizzle.readthedocs.io/en/latest/",
+        None
+    ),
 }
 
 nitpick_ignore = [("py:class", "optional"), ("py:class", "np.ndarray")]
@@ -75,4 +79,8 @@ html_sidebars = {"**": ["globaltoc.html", "relations.html", "searchbox.html"]}
 html_use_index = True
 
 # Enable nitpicky mode - which ensures that all references in the docs resolve.
-nitpicky = True
+nitpicky = False
+
+nitpick_ignore = [
+    ('py:obj', 'type'),
+]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,8 +45,6 @@ intersphinx_mapping = {
     ),
 }
 
-nitpick_ignore = [("py:class", "optional"), ("py:class", "np.ndarray")]
-
 extensions = [
     "pytest_doctestplus.sphinx.doctestplus",
     "sphinx.ext.autodoc",
@@ -82,5 +80,6 @@ html_use_index = True
 nitpicky = False
 
 nitpick_ignore = [
-    ('py:obj', 'type'),
+    ("py:class", "optional"),
+    ("py:class", "np.ndarray"),
 ]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -77,7 +77,7 @@ html_sidebars = {"**": ["globaltoc.html", "relations.html", "searchbox.html"]}
 html_use_index = True
 
 # Enable nitpicky mode - which ensures that all references in the docs resolve.
-nitpicky = False
+nitpicky = False  # True does not work with enumerated parameter values
 
 nitpick_ignore = [
     ("py:class", "optional"),
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,6 @@ dependencies = [
     "scipy >=1.14.1",
     "scikit-image>=0.20.0",
     "numpy >=1.25.0",
-    "astropy >=5.0.4",
     "opencv-python-headless >=4.6.0.66",
     "asdf >=3.3.0",
     "gwcs >=0.22.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ classifiers = [
 ]
 dependencies = [
     "astropy >=6.0.0",
-    "drizzle >=2.0.0",
+    "drizzle >=2.0.1",
     "scipy >=1.14.1",
     "scikit-image>=0.20.0",
     "numpy >=1.25.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,10 +14,11 @@ classifiers = [
 ]
 dependencies = [
     "astropy >=6.0.0",
-    "drizzle>=1.15.0",
+    "drizzle >=2.0.0",
     "scipy >=1.14.1",
     "scikit-image>=0.20.0",
     "numpy >=1.25.0",
+    "astropy >=5.0.4",
     "opencv-python-headless >=4.6.0.66",
     "asdf >=3.3.0",
     "gwcs >=0.22.0",
--- a/src/stcal/resample/__init__.py
+++ b/src/stcal/resample/__init__.py
@@ -1,12 +1,10 @@
 from .resample import (
-    OutputTooLargeError,
     Resample,
     compute_wcs_pixel_area,
     UnsupportedWCSError,
 )
 
 __all__ = [
-    "OutputTooLargeError",
     "Resample",
     "compute_wcs_pixel_area",
     "UnsupportedWCSError",
--- a/src/stcal/resample/__init__.py
+++ b/src/stcal/resample/__init__.py
@@ -1,0 +1,13 @@
+from .resample import (
+    OutputTooLargeError,
+    Resample,
+    compute_wcs_pixel_area,
+    UnsupportedWCSError,
+)
+
+__all__ = [
+    "OutputTooLargeError",
+    "Resample",
+    "compute_wcs_pixel_area",
+    "UnsupportedWCSError",
+]
--- a/src/stcal/resample/__init__.py
+++ b/src/stcal/resample/__init__.py
@@ -1,11 +1,11 @@
 from .resample import (
     Resample,
-    compute_wcs_pixel_area,
+    compute_mean_pixel_area,
     UnsupportedWCSError,
 )
 
 __all__ = [
     "Resample",
-    "compute_wcs_pixel_area",
+    "compute_mean_pixel_area",
     "UnsupportedWCSError",
 ]
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -1231,8 +1231,8 @@ class Resample:
                                      ymax=None):
         """Resample one variance image from an input model.
 
-        The error image is passed to drizzle instead of the variance, to
-        better match kernel overlap and user weights to the data, in the
+        The error image is passed to drizzle instead of the variance in order to
+        better match kernel overlap and user weights to the data during the
         pixel averaging process. The drizzled error image is squared before
         returning.
         """
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -115,8 +115,8 @@ class Resample:
             (VAR_RNOISE) array stored in each input image. If the
             ``VAR_RNOISE`` array does not exist, the variance is set to 1 for
             all pixels (i.e., equal weighting). If ``weight_type="exptime"``,
-            the weight will be set equal to the measurement time (``TMEASURE``)
-            when available and to the exposure time (``EFFEXPTM``) otherwise.
+            the weight will be set equal to the measurement time
+            when available and to the exposure time otherwise.
 
         good_bits : int, str, None, optional
             An integer bit mask, `None`, a Python list of bit flags, a comma-,
@@ -227,10 +227,10 @@ class Resample:
 
         """
         # to see if setting up arrays and drizzle is needed
+        self._locked = False
         self._finalized = False
         self._n_res_models = 0
 
-        self._n_predicted_input_models = n_input_models
         self._output_model = output_model
         self._create_new_output_model = output_model is not None
 
@@ -819,16 +819,30 @@ class Resample:
         return iscale
 
     @property
-    def finalized(self):
-        """ Indicates whether the output model was "finalized". """
-        return self._finalized
+    def is_locked(self):
+        """ Indicates whether the output model has been "locked". No further
+        calls to :py:meth:`add_model` if output model is locked.
+        """
+        return self._locked
+
+    def lock(self):
+        """ Sets "locked" flag to `True` indicating that intermediate
+        arrays have been freed and therefore no further addition of models
+        to the output model via :py:meth:`add_model` is possible.
+
+        Subclasses should call this method whenever they perform an operation
+        that would prevent further addition of input models to the output
+        model.
+        """
+        self._locked = True
 
     def reset_arrays(self, reset_output=True, n_input_models=None):
         """ Initialize/reset `Drizzle` objects, output model and arrays,
-        and time counters. Output WCS and shape are not modified from
-        `Resample` object initialization. This method needs to be called
-        before calling :py:meth:`add_model` for the first time if
-        :py:meth:`finalize` was previously called.
+        and time counters and clears the "locked" flag. Output WCS and shape
+        are not modified from `Resample` object initialization. This method
+        needs to be called before calling :py:meth:`add_model` for the first
+        time if :py:meth:`finalize` was previously called and locked the output
+        model.
 
         Parameters
         ----------
@@ -884,6 +898,7 @@ class Resample:
 
         self.init_time_counters()
 
+        self._locked = False
         self._finalized = False
 
     def validate_input_model(self, model):
@@ -953,8 +968,9 @@ class Resample:
         is `True`) , and error data (if ``enable_err`` is `True`), and adds
         them to the corresponding
         arrays of the output model using appropriate weighting.
-        It also updates the weight array and context array (if ``enable_ctx`` is `True`)
-        of the resampled data, as well as relevant metadata such as "n_coadds".
+        It also updates the weight array and context array (if ``enable_ctx``
+        is `True`) of the resampled data, as well as relevant metadata such as
+        "n_coadds".
 
         Whenever ``model`` has a unique group ID that was never processed
         before, the "pointings" value of the output model is incremented and
@@ -969,12 +985,13 @@ class Resample:
             and values of actual models used by pipelines.
 
         """
-        if self._finalized:
+        if self._locked:
             raise RuntimeError(
-                "Resampling has been finalized and intermediate arrays have "
+                "Resampling has been locked and intermediate arrays have "
                 "been freed. Unable to add new models. Call 'reset_arrays' "
                 "to initialize a new output model and associated arrays."
             )
+        self._finalized = False
         self.validate_input_model(model)
         self._n_res_models += 1
 
@@ -1055,8 +1072,16 @@ class Resample:
             # use resampled error
             self.output_model["err"] = self._driz_error.out_img
 
+    def is_finalized(self):
+        """ Indicates whether all attributes of the ``output_model`` have been
+        computed from intermediate (running) values.
+        """
+        return self._finalized
+
     def finalize(self, free_memory=True):
-        """ Finalizes all computations and frees temporary objects.
+        """ Performs final computations from any intermediate values,
+        sets ouput model values, and optionally frees temporary/intermediate
+        objects.
 
         ``finalize`` calls :py:meth:`~Resample.finalize_resample_variance` and
         :py:meth:`~Resample.finalize_time_info`.
@@ -1066,13 +1091,21 @@ class Resample:
           with ``free_memory=True`` then intermediate arrays holding variance
           weights will be lost and so continuing adding new models after
           a call to :py:meth:`~Resample.finalize` will result in incorrect
-          variance.
+          variance. In this case `finalize` will set the locked flag to `True`.
+
+        .. note::
+          If `Resample` is subclassed and :py:meth:`~Resample.finalize`
+          overridden, make sure to call :py:meth:`~Resample.lock` if an
+          irreversible operation was performed (such as freeing intermediate
+          arrays) that would prevent further addition of input models to the
+          output model.
 
         """
         if self._finalized:
             # can't finalize twice
             return
-        self._finalized = free_memory
+
+        self._finalized = True
 
         self._output_model["pointings"] = len(self.group_ids)
 
@@ -1120,10 +1153,7 @@ class Resample:
 
             del var_components
 
-        self._finalized = True
-
         self.finalize_time_info()
-
         return
 
     def init_variance_arrays(self):
@@ -1343,7 +1373,7 @@ class Resample:
             output_model["var_flat"] = output_variance
 
         if free_memory:
-            self._finalized = True
+            self.lock()
             del (
                 self._var_rnoise_wsum,
                 self._var_poisson_wsum,
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -176,8 +176,8 @@ class Resample:
             For more details, see documentation for
             `astropy.nddata.bitmask.extend_bit_flag_map`.
 
-        output_wcs : dict, WCS object, None
-            Specifies output WCS either directly as a WCS or a dictionary
+        output_wcs : dict, None
+            Specifies output WCS as a dictionary
             with keys ``'wcs'`` (WCS object) and ``'pixel_scale'``
             (pixel scale in arcseconds). ``'pixel_scale'``, when provided,
             will be used for computation of drizzle scaling factor. When it is
@@ -269,15 +269,11 @@ class Resample:
                     "'output_wcs' parameter or the 'output_model' parameter. "
                 )
             else:
-                if isinstance(output_wcs, dict):
-                    self._output_pixel_scale = output_wcs.get("pixel_scale")
-                    self._pixel_scale_ratio = output_wcs.get(
-                        "pixel_scale_ratio"
-                    )
-                    self._output_wcs = output_wcs.get("wcs")
-                else:
-                    self._output_wcs = output_wcs
-
+                self._output_pixel_scale = output_wcs.get("pixel_scale")
+                self._pixel_scale_ratio = output_wcs.get(
+                    "pixel_scale_ratio"
+                )
+                self._output_wcs = output_wcs.get("wcs")
                 self.check_output_wcs(self._output_wcs)
 
         else:
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -837,12 +837,12 @@ class Resample:
                 )
 
     def add_model(self, model):
-        """ Resamples model image and either variance data (if ``enable_var``
-        was `True`) or error data (if ``enable_err`` was `True`) and adds
-        them using appropriate weighting to the corresponding
-        arrays of the output model. It also updates resampled data weight,
-        the context array (if ``enable_ctx`` is `True`), relevant output
-        model's values such as "n_coadds".
+        """ Resamples model image, variance data (if ``enable_var``
+        is `True`) , and error data (if ``enable_err`` is `True`), and adds
+        them to the corresponding
+        arrays of the output model using appropriate weighting.
+        It also updates the weight array and context array (if ``enable_ctx`` is `True`)
+        of the resampled data, as well as relevant metadata such as "n_coadds".
 
         Whenever ``model`` has a unique group ID that was never processed
         before, the "pointings" value of the output model is incremented and
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -27,15 +27,10 @@ log.setLevel(logging.DEBUG)
 
 __all__ = [
     "compute_wcs_pixel_area"
-    "OutputTooLargeError",
     "Resample",
     "resampled_wcs_from_models",
     "UnsupportedWCSError",
 ]
-
-
-class OutputTooLargeError(RuntimeError):
-    """Raised when the output is too large for in-memory instantiation"""
 
 
 class UnsupportedWCSError(RuntimeError):
@@ -77,7 +72,7 @@ class Resample:
     dq_flag_name_map = {}  # type: dict[str, int]
 
     def __init__(self, n_input_models=None, pixfrac=1.0, kernel="square",
-                 fillval=0.0, wht_type="ivm", good_bits=0,
+                 fillval=0.0, weight_type="ivm", good_bits=0,
                  output_wcs=None, output_model=None,
                  accumulate=False, enable_ctx=True, enable_var=True,
                  compute_err=None):
@@ -116,15 +111,15 @@ class Resample:
             pixels with no contributions from input images will be set to this
             ``fillval`` value.
 
-        wht_type : {"exptime", "ivm"}, optional
-            The weighting type for adding models' data. For ``wht_type="ivm"``
-            (the default), the weighting will be determined per-pixel using
-            the inverse of the read noise (VAR_RNOISE) array stored in each
-            input image. If the ``VAR_RNOISE`` array does not exist,
-            the variance is set to 1 for all pixels (i.e., equal weighting).
-            If ``weight_type="exptime"``, the weight will be set equal
-            to the measurement time (``TMEASURE``) when available and to
-            the exposure time (``EFFEXPTM``) otherwise.
+        weight_type : {"exptime", "ivm"}, optional
+            The weighting type for adding models' data. For
+            ``weight_type="ivm"`` (the default), the weighting will be
+            determined per-pixel using the inverse of the read noise
+            (VAR_RNOISE) array stored in each input image. If the
+            ``VAR_RNOISE`` array does not exist, the variance is set to 1 for
+            all pixels (i.e., equal weighting). If ``weight_type="exptime"``,
+            the weight will be set equal to the measurement time (``TMEASURE``)
+            when available and to the exposure time (``EFFEXPTM``) otherwise.
 
         good_bits : int, str, None, optional
             An integer bit mask, `None`, a Python list of bit flags, a comma-,
@@ -247,7 +242,7 @@ class Resample:
         self._compute_err = compute_err
         self._accumulate = accumulate
 
-        # these are attributes that are used only for information purpose
+        # these attributes are used only for informational purposes
         # and are added to created the output_model only if they are
         # not already present there:
         self._pixel_scale_ratio = None
@@ -259,8 +254,8 @@ class Resample:
         self.fillval = fillval
         self.good_bits = good_bits
 
-        if wht_type in ["ivm", "exptime"]:
-            self.weight_type = wht_type
+        if weight_type in ["ivm", "exptime"]:
+            self.weight_type = weight_type
         else:
             raise ValueError("Unexpected weight type: '{self.weight_type}'")
 
@@ -269,12 +264,12 @@ class Resample:
         self.input_file_names = []
         self._group_ids = []
 
-        # determine output WCS and set-up output model if needed:
+        # determine output WCS and set up output model if needed:
         if output_model is None:
             if output_wcs is None:
                 raise ValueError(
                     "Output WCS must be provided either through the "
-                    "'output_wcs' parameter or the 'ouput_model' parameter. "
+                    "'output_wcs' parameter or the 'output_model' parameter. "
                 )
             else:
                 if isinstance(output_wcs, dict):
@@ -319,7 +314,7 @@ class Resample:
 
         self._output_array_shape = self._output_wcs.array_shape
 
-        # Check that the output data shape has no zero length dimensions
+        # Check that the output data shape has no zero-length dimensions
         npix = np.prod(self._output_array_shape)
         if not npix:
             raise ValueError(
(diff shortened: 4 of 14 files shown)


src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1584 @@
...
+import psutil
+
+from astropy import units as u
+from astropy.nddata.bitmask import (
+    bitfield_to_boolean_mask,
+    interpret_bit_flags,
+)
+from drizzle.utils import calc_pixmap
+from drizzle.resample import Drizzle
+from stdatamodels.jwst.library.basic_utils import bytes2human
```
braingram, line 21:
```suggestion
```
Looks to be unused and is causing the CI to fail.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1584 @@
...
+import os
+import warnings
+import json
+import abc
+from copy import deepcopy
+import sys
+
+import numpy as np
+from scipy.ndimage import median_filter
+import psutil
```
braingram, line 12: This isn't currently a runtime dependency (but I expect we can remove it). More on that below.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1584 @@
...
+
+        n_input_models : int, None, optional
+            Number of input models expected to be resampled. When provided,
+            this is used to estimate memory requirements and optimize memory
+            allocation for the context array.
+
+
+        """
+        if ((allowed_memory is None and
+                "DMODEL_ALLOWED_MEMORY" not in os.environ) or
```
braingram, line 470: This is no longer used in jwst or in roman. See:
https://github.com/spacetelescope/jwst/pull/8975
I suggest that we don't re-introduce this and I think removing it (and the allowed memory check) would allow dropping `psutil` as a runtime dependency.

braingram (commented):
I left a few comments to try and address some of the CI failures.

There are also check-type failures.

I also don't see any tests. Are there tests that can be moved here from jwst/romancal?

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1487 @@
...
+        running sums and weights arrays.
+        """
+        # Divide by the total weights, squared, and set in the output model.
+        # Zero weight and missing values are NaN in the output.
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", "invalid value*", RuntimeWarning)
+            warnings.filterwarnings("ignore", "divide by zero*", RuntimeWarning)
+
+            output_variance = (
+                self._var_rnoise_wsum / self._var_rnoise_weight /
```
kmacdonald-stsci, line 1161: A double division `x/y/z` is equivalent to `x / (y * z)`.  Since divisions are computationally expensive compared to multiplications, I recommend using the latter format.  And, with the parentheses, it's easier to read.

kmacdonald-stsci (approved):
Overall it looks fine, but it's being reported that a lot of the code isn't tested by any test.  Maybe add tests to get more code coverage.

src/stcal/resample/__init__.py
```diff
@@ -0,0 +1,13 @@
+from .resample import (
+    OutputTooLargeError,
+    Resample,
+    compute_wcs_pixel_area,
+    UnsupportedWCSError,
+)
+
+__all__ = [
+    "OutputTooLargeError",
```
emolter, line 9: This custom error and all associated code is being removed from JWST in build 11.2, so it should be removed here as well; see https://github.com/spacetelescope/jwst/pull/8975

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1487 @@
...
+            #     output_wcs.array_shape = wcs_pars["output_shape"]
+            # else:
+            if output_wcs.bounding_box:
+                halfpix = 0.5 + sys.float_info.epsilon
+                output_wcs.array_shape = (
+                    int(output_wcs.bounding_box[1][1] + halfpix),
+                    int(output_wcs.bounding_box[0][1] + halfpix),
+                )
+            else:
+                # TODO: In principle, we could compute footprints of all
```
emolter, line 461: Can this TODO (or any other ones in this code) be removed and if desired, turned into a GitHub issue or JIRA ticket so we can track work still needed?

emolter (commented):
I agree with Brett, the major high-level thing I see is that this needs unit tests

src/stcal/resample/utils.py
emolter, line 1: A few of the public functions here still lack detailed docstrings, with Parameters and Returns.  It would be good to add those.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+    resample_range,
+)
+
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+__all__ = [
+    "compute_wcs_pixel_area"
+    "OutputTooLargeError",
```
emolter, line 30:
```suggestion
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+    "Resample",
+    "resampled_wcs_from_models",
+    "UnsupportedWCSError",
+]
+
+
+class OutputTooLargeError(RuntimeError):
```
emolter, line 37:
```suggestion
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+        self._n_predicted_input_models = n_input_models
+        self._output_model = output_model
+        self._create_new_output_model = output_model is not None
+
+        self._enable_ctx = enable_ctx
+        self._enable_var = enable_var
+        self._compute_err = compute_err
+        self._accumulate = accumulate
+
+        # these are attributes that are used only for information purpose
```
emolter, line 250:
```suggestion
        # these attributes are used only for informational purposes
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+            self.weight_type = wht_type
+        else:
+            raise ValueError("Unexpected weight type: '{self.weight_type}'")
+
+        self._output_wcs = output_wcs
+
+        self.input_file_names = []
+        self._group_ids = []
+
+        # determine output WCS and set-up output model if needed:
```
emolter, line 272:
```suggestion
        # determine output WCS and set up output model if needed:
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+        if output_model is None:
+            if output_wcs is None:
+                raise ValueError(
+                    "Output WCS must be provided either through the "
+                    "'output_wcs' parameter or the 'ouput_model' parameter. "
```
emolter, line 277:
```suggestion
                    "'output_wcs' parameter or the 'output_model' parameter. "
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+
+        log.info(f"Driz parameter kernel: {self.kernel}")
+        log.info(f"Driz parameter pixfrac: {self.pixfrac}")
+        log.info(f"Driz parameter fillval: {self.fillval}")
+        log.info(f"Driz parameter weight_type: {self.weight_type}")
+        log.debug(
+            f"Output mosaic size (nx, ny): {self._output_wcs.pixel_shape}"
+        )
+
+        # set up an empty (don't allocate arrays at this time) output model:
```
emolter, line 338:
```suggestion
        # set up an empty output model (don't allocate arrays at this time)
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+                f"{self._output_pixel_scale} arcsec."
+            )
+        else:
+            log.info(
+                f"Output pixel scale: {self._output_pixel_scale} arcsec."
+            )
+
+        self._output_array_shape = self._output_wcs.array_shape
+
+        # Check that the output data shape has no zero length dimensions
```
emolter, line 322:
```suggestion
        # Check that the output data shape has no zero-length dimensions
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+        if enable_var:
+            attributes.update(
+                ["var_rnoise", "var_poisson", "var_flat"]
+            )
+            # TODO: if we want to support adding more data to
+            # existing output models, we need to also store weights
+            # for variance arrays:
+            # var_rnoise_weight
+            # var_flat_weight
+            # var_poisson_weight
```
emolter, line 403: is there any plan to do this in the near future?  If not, I suggest removing the comment

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+        ``array_shape`` property is defined.
+
+        Parameters
+        ----------
+        output_wcs : WCS object
+            A WCS object corresponding to the output (resampled) image.
+
+        estimate_output_shape : bool, optional
+            Indicates whether to *estimate* pixel scale of the ``output_wcs``
+            from
```
emolter, line 439: This appears to end before the end of the sentence

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+            raise UnsupportedWCSError(
+                "Output WCS needs 2 coordinate axes but the "
+                f"supplied WCS has {naxes} axes."
+            )
+
+        # make sure array_shape and pixel_shape are set:
+        if output_wcs.array_shape is None and estimate_output_shape:
+            # if wcs_pars and "output_shape" in wcs_pars:
+            #     output_wcs.array_shape = wcs_pars["output_shape"]
+            # else:
```
emolter, line 461:
```suggestion
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+
+        n_input_models : int, None, optional
+            Number of input models expected to be resampled. When provided,
+            this is used to estimate memory requirements and optimize memory
+            allocation for the context array.
+
+        """
+        self._n_predicted_input_models = n_input_models
+
+        # set up an empty (don't allocate arrays at this time) output model:
```
emolter, line 740: same as above

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+            and values of actual models used by pipelines.
+
+        Raises
+        ------
+        KeyError
+            A `KeyError` is raised when ``model`` does not have a required
+            keyword.
+
+        """
+        # TODO: do we need this to just raise a custom
```
emolter, line 800: Has this been resolved?

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+                    "in a data model."
+                )
+
+    def add_model(self, model):
+        """ Resamples model image and either variance data (if ``enable_var``
+        was `True`) or error data (if ``enable_err`` was `True`) and adds
+        them using appropriate weighting to the corresponding
+        arrays of the output model. It also updates resampled data weight,
+        the context array (if ``enable_ctx`` is `True`), relevant output
+        model's values such as "n_coadds".
```
emolter, line 850:
```suggestion
        """ Resamples model image, variance data (if ``enable_var``
        is `True`) , and error data (if ``enable_err`` is `True`), and adds
        them to the corresponding
        arrays of the output model using appropriate weighting.
        It also updates the weight array and context array (if ``enable_ctx`` is `True`)
        of the resampled data, as well as relevant metadata such as "n_coadds".
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+            model,
+            weight_type=self.weight_type,
+            good_bits=self.good_bits,
+        )
+
+        # apply sky subtraction
+        blevel = model["level"]
+        if not model["subtracted"] and blevel is not None:
+            data = data - blevel
+            # self._output_model["subtracted"] = True
```
emolter, line 911:
```suggestion
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+            data.shape,
+            wcs.bounding_box
+        )
+
+        add_image_kwargs = {
+            'exptime': model["exposure_time"],
+            'pixmap': pixmap,
+            'scale': iscale,
+            'weight_map': weight,
+            'wht_scale': 1.0,  # hard-coded for JWST count-rate data
```
emolter, line 923: Is this okay for Romancal?  If not, perhaps now is the time to allow it as a variable

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+            'pixfrac': self.pixfrac,
+            'in_units': 'cps',  # TODO: get units from data model
```
emolter, line 925: is this TODO still relevant?

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1530 @@
...
+    def _resample_one_variance_array(self, name, model, iscale,
+                                     weight_map, pixmap,
+                                     xmin=None, xmax=None, ymin=None,
+                                     ymax=None):
+        """Resample one variance image from an input model.
+
+        The error image is passed to drizzle instead of the variance, to
+        better match kernel overlap and user weights to the data, in the
+        pixel averaging process. The drizzled error image is squared before
+        returning.
```
emolter, line 1243:
```suggestion
        The error image is passed to drizzle instead of the variance in order to
        better match kernel overlap and user weights to the data during the
        pixel averaging process. The drizzled error image is squared before
        returning.
```

emolter (commented):
looks good overall, most of my comments are surface-level

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1487 @@
...
+            .. note::
+                At this time, output error array is not equivalent to
+                error propagation results.
+
+        """
+        # always required:
+        attributes = {
+            "data",
+            "wcs",
+            "wht",
```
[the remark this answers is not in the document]
mcara: These have direct analogue in the JWST `DataModel`. That is why I preserved the name.

src/stcal/resample/resample.py
mcara: It could, but does it make sense to open a ticket to modify code that does not yet exist in the repository.

src/stcal/resample/resample.py
mcara: seems to be OK: https://github.com/spacetelescope/romancal/blob/e250097fb7113be808836a796dc73ba7f687c7b8/romancal/resample/gwcs_drizzle.py#L206

src/stcal/resample/resample.py
mcara: Yes

src/stcal/resample/resample.py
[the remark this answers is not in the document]
mcara: Here are a couple of examples:

https://github.com/spacetelescope/romancal/blob/e250097fb7113be808836a796dc73ba7f687c7b8/romancal/resample/gwcs_drizzle.py#L78

https://github.com/spacetelescope/jwst/blob/40e43c8dbe5c3cec269c4d014dfbf53ddb23a3bb/jwst/resample/resample.py#L413

https://github.com/spacetelescope/jwst/blob/363a9ad90eb66111bcecd65edab36b2a9e631feb/jwst/resample/resample.py#L499

src/stcal/resample/resample.py
emolter: It's not necessarily my decision but I would say yes, because this is definitely going to get added at some point and it's easier to track issues than code comments

src/stcal/resample/resample.py
emolter: great, thanks for checking. maybe update the comment to say that it was also tested for Roman, otherwise it looks like it may have been accidentally ported over from jwst without additional checking

mcara: > maybe update the comment to say that it was also tested for Roman, otherwise it looks like it may have been accidentally ported over from jwst without additional checking

I did not test it and I am not aware of this being tested but _I assume_ the resample code was tested as a whole by Roman. I actually do not see under what circumstances ``wht_scale`` would need to be `!= 1`. I think the comment should be removed as I think units of the data are irrelevant here.

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,298 @@
...
+def is_imaging_wcs(wcs):
+    """ Returns `True` if ``wcs`` is an imaging WCS and `False` otherwise. """
+    imaging = all(
+        ax == 'SPATIAL' for ax in wcs.output_frame.axes_type
+    )
+    return imaging
+
+
+def compute_wcs_pixel_area(wcs, shape=None):
+    """ Computes pixel area in steradians.
```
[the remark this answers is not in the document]
mcara: The purpose of this PR is to move code that may be shared by multiple pipelines (JWST, Roman, ...) to `stcal`. This PR attempts to preserve how current pipelines work and I do not think this should be modified here.

 > I'm not happy with this and the following routine. It is I think overly complex and doesn't even explain the methodology of how the pixel area is computed.

I will add a little more documentation.

> But why not simply compute the area of one pixel in the center?

We would not need to compute anything if we knew where/how `PIXAR_*` were computed and them maybe we could used them directly without computing anything.

> I gather that it is an average area of pixels over the whole valid image area. But why not simply compute the area of one pixel in the center?

It is too complicated to describe it here but this came from the ambiguity of how `PIXAR_SR` was defined for JWST  images. I attempted to provide a longer explanation here: https://github.com/spacetelescope/jwst/pull/7894#issuecomment-1720338980

Further reading: https://github.com/spacetelescope/jwst/issues/7668#issuecomment-1715808893 and other comments in that issue.

> Perhaps it can be handled in a subsequent PR if this is urgent.

Again, this is code already used in JWST. Please open an issue if you feel this is incorrect.

perrygreenfield (approved):
LGTM

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,390 @@
...
+        xmax = data_shape[1] - 1
+        ymax = data_shape[0] - 1
+    else:
+        ((x1, x2), (y1, y2)) = bbox
+        xmin = max(0, int(x1 + 0.5))
+        ymin = max(0, int(y1 + 0.5))
+        xmax = min(data_shape[1] - 1, int(x2 + 0.5))
+        ymax = min(data_shape[0] - 1, int(y2 + 0.5))
+
+    return xmin, xmax, ymin, ymax
```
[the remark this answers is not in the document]
mcara: Somehow, github won't allow me to reply separately to each comment.

> Do you just need one of these (i.e., for squared kernel terms), or do you need all of them?

I think so. I do not think it needs to be done now, especially that I've heard that Roman is no longer interested in this functionality, but this is something to keep in mind for later *only if* we will keep current method of estimating variances. If we propagate errors like it was proposed in https://github.com/spacetelescope/jwst/pull/8997 and https://github.com/spacetelescope/drizzle/pull/163, then these arrays are not needed.

> the maxes at int(x + 0.5) look a bit weird. I might have guessed floor(x + 0.5), ceil(x - 0.5).

`int(x + 0.5)` is the same as `int(floor(x + 0.5))`.  `xmin, xmax, ymin, ymax` must all be integer numbers. If we make a rounding error here, it preferably should be to expand the box, not to shrink it: minor efficiency loss is probably preferable to missing data.

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,390 @@
...
+
+    if xmin > xmax:
+        (xmin, xmax) = (xmax, xmin)
+    if ymin > ymax:
+        (ymin, ymax) = (ymax, ymin)
+
+    xmin = max(0, int(xmin + 0.5))
+    xmax = min(nx - 1, int(xmax - 0.5))
+    ymin = max(0, int(ymin + 0.5))
+    ymax = min(ny - 1, int(ymax - 0.5))
```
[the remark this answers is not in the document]
mcara: > ... though the half pixel offsets aren't quite the same.
Exactly.

pyproject.toml
```diff
@@ -14,7 +14,8 @@ classifiers = [
 ]
 dependencies = [
     "astropy >=5.0.4",
-    "drizzle>=1.15.0",
+    # "drizzle>=2.0.0",
+    "drizzle @ git+https://github.com/mcara/drizzle.git@disable-bbox",
```
braingram, line 18: Is there a PR open for this?

braingram:
```suggestion
    "drizzle @ git+https://github.com/mcara/drizzle.git@main",
```
It looks like it was https://github.com/spacetelescope/drizzle/pull/164. My vote is to release this and add a lower pin for the released version.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1660 @@
...
+    (data, variance array) to an output image defined by an output WCS.
+
+    In particular, this class performs the following operations:
+
+    1. Sets up output arrays based on arguments used at initialization.
+    2. Based on information about the input images and user arguments, computes
+       scale factors needed to obtain correctly convert resampled counts to
+       fluxes.
+    3. For each input image computes coordinate transformations (``pixmap``)
+       from coordinate system of the input image to the coordinate system of
```
emolter, line 53:
```suggestion
       from the coordinate system of the input image to the coordinate system of
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1660 @@
...
+       the output image.
+    4. For each input image computes weight image.
```
emolter, line 55:
```suggestion
    4. Computes the weight image for each input image.
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1660 @@
...
+        # set up an empty output model (don't allocate arrays at this time):
+        if self._output_model is None:
+            self._output_model = self.create_output_model()
+
+        self.reset_arrays(reset_output=False, n_input_models=n_input_models)
+
+    def get_input_model_pixel_area(self, model):
+        """
+        Computes or retrieves pixel area of an input model. Currently,
+        this is the average pixel area of input model's pixels within either
```
emolter, line 347:
```suggestion
        this is the average pixel area of the input model's pixels within either
```

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1660 @@
...
+        the bounding box (if available) or the entire data array.
+
+        This value is used to compute a scale factor that will be applied
+        to input image data. This scale factor takes into account the
+        difference in the definition of the pixel area reported in model's
+        ``meta`` and the pixel area at the location used to construct
```
emolter, line 353:
```suggestion
        difference in the definition of the pixel area reported in
        ``model.meta`` and the pixel area at the location used to construct
```
...although, now that this is written in such a way that everything is passed in and out with dictionaries, is `meta` still relevant?

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1660 @@
...
+        output WCS from the WCS of input models using ``pixel_scale_ratio``.
+
+        Intensity scale factor is computed elsewhere as the ratio of the value
```
emolter, line 356:
```suggestion
        The intensity scale factor is computed elsewhere as the ratio of the value
```

emolter (commented):
Mostly grammar, but a few other things I'm wondering about too.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1660 @@
...
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+__all__ = [
+    "Resample",
+    "UnsupportedWCSError",
+]
+
+
+class UnsupportedWCSError(RuntimeError):
```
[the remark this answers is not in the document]
mcara: I think the original exception was a `RuntimeError` and then, I think, @braingram suggested/modified it to be a custom exception and so it inherits from `RuntimeError` for backward compatibility purpose.

src/stcal/resample/resample.py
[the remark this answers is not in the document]
mcara: > It looks like this is being changed in jwst, so can it also be changed here?

I am sorry, what is being changed? can you post a link? Thanks!

docs/conf.py
```diff
@@ -75,4 +79,8 @@
 html_use_index = True
 
 # Enable nitpicky mode - which ensures that all references in the docs resolve.
-nitpicky = True
+nitpicky = False
+
+nitpick_ignore = [
+    ('py:obj', 'type'),
+]
```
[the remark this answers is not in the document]
mcara: I don't remember. some style check, I think.

src/stcal/resample/resample.py
mcara: "model: A dictionary containing data arrays and other meta attributes". So I think it is ok to refer to "meta" but not ``model.meta``.

mcara: I will restore previous text.

docs/conf.py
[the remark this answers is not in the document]
mcara: Ah, some warnings about undefined references, I don't recall exactly.

mcara:
> oldest

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1660 @@
...
+            in_units="cps",
+            xmin=xmin,
+            xmax=xmax,
+            ymin=ymin,
+            ymax=ymax,
+        )
+
+        return driz.out_img ** 2
+
+    def build_driz_weight(self, model, weight_type=None, good_bits=None):
```
braingram, line 1409: `build_driz_weight` is currently a function provided by [resample_utils](https://github.com/spacetelescope/jwst/blob/f1cb0a61e9ee8989bc770157f297f64b1b7dbc68/jwst/resample/resample_utils.py#L194) (in jwst) and used by outlier detection.
It is used even if [no resampling is performed](https://github.com/spacetelescope/jwst/blob/f1cb0a61e9ee8989bc770157f297f64b1b7dbc68/jwst/outlier_detection/utils.py#L124).

By making it a method on the `ResampleData` class we'd have to either keep the duplicated code or construct a `ResampleData` instance (which would allocate an unused output model etc).

Does this need to be a method on `ResampleData`?

src/stcal/resample/utils.py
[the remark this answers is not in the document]
mcara: @schlafly can I resolve this conversation?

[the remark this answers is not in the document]
schlafly: Yes, thanks!

docs/conf.py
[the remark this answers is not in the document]
emolter: Could you provide more details?  I worry that this is hiding problems with the documentation build.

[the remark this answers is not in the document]
mcara:
```
building [mo]: targets for 0 po files that are out of date
writing output...
building [html]: targets for 19 source files that are out of date
updating environment: [new config] 53 added, 0 changed, 0 removed
reading sources... [100%] stcal/tweakreg/utils
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
copying assets...
copying static files...
Writing evaluated template result to /Users/mcara/repo/stcal/docs/_build/html/_static/basic.css
Writing evaluated template result to /Users/mcara/repo/stcal/docs/_build/html/_static/language_data.js
Writing evaluated template result to /Users/mcara/repo/stcal/docs/_build/html/_static/documentation_options.js
Writing evaluated template result to /Users/mcara/repo/stcal/docs/_build/html/_static/js/versions.js
copying static files: done
copying extra files...
copying extra files: done
copying assets: done
writing output... [100%] stcal/tweakreg/utils
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: {"square" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "gaussian" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "point" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "turbo" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "lanczos2" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "lanczos3"} [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: {"ivm" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "exptime"} [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: WCS object [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: {"from_var" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "driz_err"} [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: {"exptime" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "ivm"} [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: WCS object [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: {"from_var" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "driz_err"} [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: 3D array [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: 2D array [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: {"from_var" [ref.class]
/Users/mcara/repo/stcal/docs/api/stcal.resample.Resample.rst:2: WARNING: py:class reference target not found: "driz_err"} [ref.class]
WARNING: dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting
generating indices... genindex py-modindex done
highlighting module code... [100%] stcal.tweakreg.utils
writing additional pages... search done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 21 warnings.
```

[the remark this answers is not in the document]
emolter: ah ok, it is complaining about any unrecognized types.  I guess this isn't a big deal and fine to ignore, although it should be fixable by changing a bunch of the parameter types from things like `{"from_var", "driz_err"}` to `str` and then putting the options in the parameter description.  And for the arrays, I think the proper type is `np.ndarray` instead of e.g. `2D array`

src/stcal/resample/resample.py
mcara: moved it `build_driz_weight` to `stcal.resample.utils`. `ResampleData` does not exist, I think...

src/stcal/resample/resample.py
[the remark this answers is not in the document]
emolter: the variable name from `wht` to `weight`, which I think you have already done

emolter (approved):
Looks good to me now, thanks for all the changes

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1512 @@
...
+       from the coordinate system of the input image to the coordinate system
+       of the output image.
+    4. Computes the weight image for each input image.
+    5. Calls :py:class:`~drizzle.resample.Drizzle` methods to resample and
+       combine input images and their variance/error arrays.
+    6. Keeps track of total exposure time and other time-related quantities.
+
+    """
+    resample_suffix = 'i2d'
+    resample_file_ext = '.fits'
```
braingram, line 56:
```suggestion
```

Are these used anywhere?

braingram (changes requested):

src/stcal/resample/resample.py
mcara: it was a leftover from an older commit.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1512 @@
...
+            pixels will be assigned zero weight and thus these pixels
+            will not contribute to the output resampled data array.
+
+            Set `good_bits` to `None` to turn off the use of model's DQ
+            array.
+
+            For more details, see documentation for
+            `astropy.nddata.bitmask.extend_bit_flag_map`.
+
+        output_wcs : dict, WCS object, None
```
braingram, line 179: Why are there so many options for this (and the other input arguments)? Is there ever a place where `ouput_wcs` is not a dict?

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,513 @@
...
+
+def build_driz_weight(model, weight_type=None, good_bits=None,
+                      flag_name_map=None):
+    """ Create a weight map that is used for weighting input images when
+    they are co-added to the ouput model.
+
+    Parameters
+    ----------
+    model : dict
+        Input model: a dictionar of relevant keywords and values.
```
braingram, line 71: What are the relevant keywords and values? To check I'm following things, "model" here is a dictionary that needs the following keys:
- "data"
- "dq"
- "var_rnoise" (if "ivm" is used)
if "exptime" is used the dictionary needs either "measurement_time" or "exposure_time" but if "measurement_time" has a value of None then "exposure_time" must also be present.

What did I get wrong there?

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,513 @@
...
+
+    weight_type : {"exptime", "ivm"}, optional
+        The weighting type for adding models' data. For
+        ``weight_type="ivm"`` (the default), the weighting will be
+        determined per-pixel using the inverse of the read noise
+        (VAR_RNOISE) array stored in each input image. If the
+        ``VAR_RNOISE`` array does not exist,
+        the variance is set to 1 for all pixels (i.e., equal weighting).
+        If ``weight_type="exptime"``, the weight will be set equal
+        to the measurement time (``TMEASURE``) when available and to
```
braingram, line 81:
```suggestion
        to the measurement time (``TMEASURE``) when available and to
```

Would you update this docstring to remove this and the other FITS keywords?

src/stcal/resample/utils.py
```diff
@@ -0,0 +1,513 @@
...
+        return model["exposure_time"], False
+    if tmeasure is None:
+        return model["exposure_time"], False
+    else:
+        return tmeasure, True
+
+
+# FIXME: temporarily copied here to avoid this import:
+# from stdatamodels.jwst.library.basic_utils import bytes2human
+def bytes2human(n):
```
braingram, line 223: This is unused right?

mcara: yeah, a leftover since when we were checking for the available memory. Removed.

src/stcal/resample/resample.py
```diff
@@ -0,0 +1,1512 @@
...
+            .. note::
+                At this time, output error array is not equivalent to
+                error propagation results.
+
+        """
+        # to see if setting up arrays and drizzle is needed
+        self._finalized = False
+        self._n_res_models = 0
+
+        self._n_predicted_input_models = n_input_models
```
braingram, line 233: Is this used anywhere?

mcara: removed

src/stcal/resample/utils.py
mcara: removed

src/stcal/resample/resample.py
mcara: For example here: https://github.com/spacetelescope/jwst/blob/144c4dcf88df16d2f143a82c936d07fc77c7c13c/jwst/resample/resample_step.py#L179-L182

braingram: If that's the only non-dict usage why not update that to produce a dict and make the code here simpler?

mcara: I do not believe the code to handle this is overly (or even slightly) complicated. In addition to allowing `load_custom_wcs()` to return a simple WCS object, it also allows users who instantiate `ResampleImage` (or resample step) directly to pass directly a WCS object without having to put it into a dictionary.

braingram: I don't feel the same way. Since the output of `load_custom_wcs` passes through the `ResampleImage` constructor in jwst why can't it wrap it in a dict and pass it to this class (which could support 1 less option for this input type which would certainly be simpler).

braingram: For the `Step` interface in jwst I'm in favor of the second. The point I'm trying to make here is that the `Step` can take whatever the user provides (a `gwcs.WCS` instance or a filename pointing to an asdf file) and convert it to a single (or minimal set) of formats that the stcal interface implemented here can support. I don't see why stcal needs to support such a wide variety of formats when the step can convert to a smaller set of supported ones.

src/stcal/resample/resample.py
mcara: See https://jira.stsci.edu/browse/JP-3874

braingram (commented):
Thanks for all the updates. A few (hopefully final) comments about unused code and simplifications. Also one comment about disabling the context array allocation when `enable_ctx=False`.

braingram (approved):
LGTM (pending jwst regtests and jwst PR approval). Thanks for all your work on this!

PR #314, opened 2024-10-30
Inverse wcs bbox

nden:
This is part of the work for AL-852: using the bounding_box in the inverse transforms.
It requires spacetelescope/gwcs#498 to be merged

Follow up PRs are spacetelescope/jwst#8554  and spacetelescope/romancal#1273

Closes #

This PR addresses ...

## Tasks
- [ ] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [ ] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
  - [ ] run regression tests with this branch installed (`"git+https://github.com/<fork>/stcal@<branch>"`)
    - [ ] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) 
    - [ ] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml)

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.apichange.rst``: change to public API
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.general.rst``: infrastructure or miscellaneous change
</details>

PR #323, opened 2024-12-13
revert #314 and #322

braingram:
#314 and #322 are causing jwst test failures. This PR reverts #314 and #322 to allow ongoing jwst and stcal work to be tested.

## Tasks

</details>

### DIFF
--- a/changes/314.bugfix.rst
+++ b/changes/314.bugfix.rst
@@ -1,1 +1,0 @@
-Do not evaluate the inverse WCS transform within the bounding box in cases where a resampled WCS is computed. Do not pass table columns to the WCS Shared API.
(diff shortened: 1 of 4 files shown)


melanieclarke (approved):

braingram:
Downstream tests are failing since the jwst/romancal downstream had/has stcal@main as a dependency which has gwcs@master as a dependency. This caused gwcs@master to be installed (even though it's not a dependency in this PR) which is the root of the changes leading to test failures. I'm confident (but not certain) that if this PR is merged into stcal@main the tests would no longer pick up gwcs@master and not fail. This does highlight an issue with our test setup here.

PR #324, opened 2024-12-16
Fix issues in outlier det. due to GWCS.inverse enforcing bbox

mcara:
Fixes failures in JWST pipeline due to GWCS inverse transformation enforcing bounding box.

## Tasks

</details>

### DIFF
--- a/changes/324.bugfix.rst
+++ b/changes/324.bugfix.rst
@@ -1,0 +1,1 @@
+Updated ``outlier_detection`` to work with the upcoming GWCS release that by default enables filtering of world coordinates by footprint of the bounding box on the sky for inverse transformations (world->detector).
--- a/src/stcal/outlier_detection/utils.py
+++ b/src/stcal/outlier_detection/utils.py
@@ -284,7 +284,18 @@ def calc_gwcs_pixmap(in_wcs, out_wcs, in_shape):
     log.debug("Bounding box from data shape: {}".format(bb))
 
     grid = gwcs.wcstools.grid_from_bounding_box(bb)
-    return np.dstack(reproject(in_wcs, out_wcs)(grid[0], grid[1]))
+
+    # temporarily disable the bounding box:
+    orig_bbox = out_wcs.bounding_box
+    out_wcs.bounding_box = None
+    try:
+        pixmap = np.dstack(reproject(in_wcs, out_wcs)(grid[0], grid[1]))
+    finally:
+        # restore bounding box:
+        if orig_bbox is not None:
+            out_wcs.bounding_box = orig_bbox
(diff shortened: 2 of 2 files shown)


emolter (commented):
@mcara @nden Did this failure show up anywhere besides the outlier detection unit tests?  e.g., in the regression tests or with any real data?  The outlier detection tests were not calling AssignWcsStep properly and fixing that seemed to resolve the problem without any code changes outside the test itself: https://github.com/nden/jwst/pull/1/files 

 Is there a good reason to explicitly ignore the bounding box when calling reproject()?  It seems to me that if the bounding box is reasonable, then the proposed change should not be necessary.  The bounding box was not reasonable for the outlier tests, and calling AssignWcs made it reasonable and made the tests pass.  I wonder if something similar could be happening for the resample unit test?

mcara:
> Did this failure show up anywhere besides the outlier detection unit tests? e.g., in the regression tests or with any real data?

I did not run regression tests (or real data) if a unit test is failing.

> Is there a good reason to explicitly ignore the bounding box when calling reproject()? It seems to me that if the bounding box is reasonable, then the proposed change should not be necessary. The bounding box was not reasonable for the outlier tests, and calling AssignWcs made it reasonable and made the tests pass. I wonder if something similar could be happening for the resample unit test?

 I am not sure about the root-cause, but _it appears_ (to me) that the issue is only with some instruments. In any case, `drizzle` can filter out pixels that are outside of the output frame by itself (without much overhead) and it can also handle `NaN` in `pixmap`. However, it cannot handle the case when ALL values in `pixmap` are `NaN` which is what the failures were in many unit tests.

emolter:
> I am not sure about the root-cause, but _it appears_ (to me) that the issue is only with some instruments. In any case, `drizzle` can filter out pixels that are outside of the output frame by itself (without much overhead) and it can also handle `NaN` in `pixmap`. However, it cannot handle the case when ALL values in `pixmap` are `NaN` which is what the failures were in many unit tests.

Why does it appear to you that the failures are instrument-specific?

The all-NaN pixmaps were an effect of a bad input WCS, which led `reproject()` to return NaN no matter what was input.  As I said, calling AssignWcsStep prior to running the test (instead of `assign_wcs.pointing.create_fitswcs`) solves the problem for the outlier unit tests.  I don't think an all-NaN pixmap should happen unless the input WCS is not correct for the input data, which is not something that (IMO) should be hot-fixed in outlier detection.  It seems like at best this just hides a pre-existing problem.

kmacdonald-stsci (approved):

braingram:
@mcara is this still relevant?

PR #326, opened 2024-12-18
JP-3824: Fix a bug in wcs_from_sregions when crpix is provided

mcara:
Resolves [JP-3824](https://jira.stsci.edu/browse/JP-3824)

`wcs_from_sregions` is used by `make_output_wcs` function in the JWST pipeline to compute output WCS of the resampled image for the `ResampleStep` . Currently `wcs_from_sregions` assigns a bounding box to the output WCS that is computed in the assumption that `crpix` is `(0, 0)` and it is not adjusted when `crpix` is provided by the user and when it is different from `(0, 0)`. This causes unit test failures in JWST pipeline when run with master branch of `gwcs` since now inverse WCS transforms check for bounding box.

A separate issue observed while working on this (which may or may not be an issue) is that currently `crpix` is considered 0-indexed while in FITS WCS standard it is 1-indexed.

## Tasks

</details>

### DIFF
--- a/changes/326.bugfix.rst
+++ b/changes/326.bugfix.rst
@@ -1,0 +1,4 @@
+Fix a bug in ``alignment.util.wcs_from_sregions()`` and
+``alignment.util.wcs_from_footprints()`` functions due to which, when the
+``crpix`` argument was provided, the bounding box of the created WCS was
+not adjusted to account for this ``crpix`` value.
--- a/changes/326.bugfix.rst
+++ b/changes/326.bugfix.rst
@@ -1,4 +1,3 @@
 Fix a bug in ``alignment.util.wcs_from_sregions()`` and
-``alignment.util.wcs_from_footprints()`` functions due to which, when the
-``crpix`` argument was provided, the bounding box of the created WCS was
-not adjusted to account for this ``crpix`` value.
+``alignment.util.wcs_from_footprints()`` functions that was causing incorrect WCS bounding boxes
+when the ``crpix`` argument was provided.
--- a/src/stcal/alignment/util.py
+++ b/src/stcal/alignment/util.py
@@ -332,7 +332,12 @@ def _calculate_new_wcs(wcs: gwcs.wcs.WCS,
     wcs_new.bounding_box = output_bounding_box
 
     if shape is None:
-        shape = [int(axs[1] - axs[0] + 0.5) for axs in output_bounding_box[::-1]]
+        if crpix is None:
+            shape = [
+                int(axs[1] - axs[0] + 0.5) for axs in output_bounding_box[::-1]
+            ]
+        else:
+            shape = [int(axs[1] + 1.5) for axs in output_bounding_box[::-1]]
 
     wcs_new.pixel_shape = shape[::-1]
     wcs_new.array_shape = shape
--- a/src/stcal/alignment/util.py
+++ b/src/stcal/alignment/util.py
@@ -4,7 +4,8 @@ from __future__ import annotations
 import functools
 import logging
 import re
-from typing import TYPE_CHECKING
+import typing
+from typing import TYPE_CHECKING, Tuple
 import warnings
 
 if TYPE_CHECKING:
@@ -143,10 +144,16 @@ def _generate_tranform(
     return transform
 
 
-def _get_axis_min_and_bounding_box(footprints: list[np.ndarray],
-                                   ref_wcs: gwcs.wcs.WCS) -> tuple:
+def _get_bounding_box_with_offsets(
+        footprints: list[np.ndarray],
+        ref_wcs: gwcs.wcs.WCS,
+        fiducial: Sequence,
+        crpix: Sequence | None,
+        shape: Sequence | None
+        ) -> Tuple[tuple, tuple, astmodels.Model]:
     """
     Calculates axis minimum values and bounding box.
+    Calculates the offsets to the transform.
 
     Parameters
     ----------
@@ -157,28 +164,78 @@ def _get_axis_min_and_bounding_box(footprints: list[np.ndarray],
     ref_wcs : ~gwcs.wcs.WCS
         The reference WCS object.
 
+    fiducial : tuple
+        A tuple containing the world coordinates of the fiducial point.
+
+    crpix : list or tuple, optional
+        0-indexed pixel coordinates of the reference pixel.
+
+    shape : tuple, optional
+        Shape (using `numpy.ndarray` convention) of the image array associated
+        with the ``ref_wcs``.
+
+
     Returns
     -------
+
     tuple
-        A tuple containing two elements:
-            1 - a :py:class:`np.ndarray` with the minimum value in each axis;
-            2 - a tuple containing the bounding box region in the format
-            ((x0_lower, x0_upper), (x1_lower, x1_upper)).
+        A tuple containing the bounding box region in the format
+        ((x0_lower, x0_upper), (x1_lower, x1_upper), ...).
+
+    tuple
+        Shape of the image. When ``shape`` argument is `None`, shape is
+        determined from the upper limit of the computed bounding box, otherwise
+        input value of ``shape`` is returned.
+
+    ~astropy.modeling.Model
+        A model with the offsets to be added to the WCS's transform.
+
     """
     domain_bounds = np.hstack([ref_wcs.backward_transform(*f.T) for f in footprints])
-    axis_min_values = np.min(domain_bounds, axis=1)
-    domain_bounds = (domain_bounds.T - axis_min_values).T
+    domain_min = np.min(domain_bounds, axis=1)
+    domain_max = np.max(domain_bounds, axis=1)
 
-    output_bounding_box = []
-    for axis in ref_wcs.output_frame.axes_order:
-        axis_min, axis_max = (
-            0.0,
-            domain_bounds[axis].max(),
+    native_crpix = ref_wcs.backward_transform(*fiducial)
+
+    if crpix is None:
+        # shift the coordinates by domain_min so that all input footprints
+        # will project to positive coordinates in the offsetted ref_wcs
+        offsets = tuple(ncrp - dmin for ncrp, dmin in zip(native_crpix, domain_min))
+
+    else:
+        # assume 0-based CRPIX and require that fiducial would map to the user
+        # defined crpix value:
+        offsets = tuple(
+            crp - ncrp for ncrp, crp in zip(native_crpix, crpix)
         )
-        # populate output_bounding_box
-        output_bounding_box.append((axis_min, axis_max))
 
-    return (axis_min_values, output_bounding_box)
+    # Also offset domain limits:
+    domain_min += offsets
+    domain_max += offsets
+
+    if shape is None:
+        shape = tuple(int(dmax + 0.5) for dmax in domain_max[::-1])
+        bounding_box = tuple(
+            (-0.5, s - 0.5) for s in shape[::-1]
+        )
+        # code to reproduce old results (for old unit test test_wcs_from_footprints)
+        # bounding_box = tuple(
+        #     (0.0, dmax) for dmax in domain_max
+        # )
+        # shape = tuple(int(dmax + 0.5) for dmax in domain_max[::-1])
+
+    else:
+        # trim upper bounding box limits
+        bounding_box = tuple(
+            (max(0, int(dmin + 0.5)) - 0.5, min(int(dmax + 0.5), sz) - 0.5)
+            for dmin, dmax, sz in zip(domain_min, domain_max, shape[::-1])
+        )
+
+    model = astmodels.Shift(-offsets[0], name="crpix1")
+    for k, shift in enumerate(offsets[1:]):
+        model = model.__and__(astmodels.Shift(-shift, name=f"crpix{k + 2:d}"))
+
+    return bounding_box, shape, model
 
 
 def _calculate_fiducial(footprints: list[np.ndarray],
@@ -206,53 +263,6 @@ def _calculate_fiducial(footprints: list[np.ndarray],
     if crval is not None:
         return tuple(crval)
     return _compute_fiducial_from_footprints(footprints)
-
-
-def _calculate_offsets(fiducial: tuple,
-                       wcs: gwcs.wcs.WCS | None,
-                       axis_min_values: np.ndarray | None,
-                       crpix: Sequence | None) -> astmodels.Model:
-    """
-    Calculates the offsets to the transform.
-
-    Parameters
-    ----------
-    fiducial : tuple
-        A tuple containing the world coordinates of the fiducial point.
-
-    wcs : ~gwcs.wcs.WCS
-        A WCS object. It will be used to determine the
-
-    axis_min_values : np.ndarray
-        A two-elements array containing the minimum pixel value for each axis.
-
-    crpix : list or tuple
-        0-indexed pixel coordinates of the reference pixel.
-
-    Returns
-    -------
-    ~astropy.modeling.Model
-        A model with the offsets to be added to the WCS's transform.
-
-    Notes
-    -----
-    If ``crpix=None``, then ``fiducial``, ``wcs``, and ``axis_min_values`` must be
-    provided, in which case, the offsets will be calculated using the WCS object to
-    find the pixel coordinates of the fiducial point and then correct it by the minimum
-    pixel value for each axis.
-    """
-    if crpix is None and fiducial is not None and wcs is not None and axis_min_values is not None:
-        offset1, offset2 = wcs.backward_transform(*fiducial)
-        offset1 -= axis_min_values[0]
-        offset2 -= axis_min_values[1]
-    elif crpix is None:
-        msg = "If crpix is not provided, fiducial, wcs, and axis_min_values must be provided."
-        raise ValueError(msg)
-    else:
-        # assume 0-based CRPIX
-        offset1, offset2 = crpix
-
-    return astmodels.Shift(-offset1, name="crpix1") & astmodels.Shift(-offset2, name="crpix2")
 
 
 def _calculate_new_wcs(wcs: gwcs.wcs.WCS,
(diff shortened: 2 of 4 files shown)


src/stcal/alignment/util.py
```diff
@@ -302,19 +308,36 @@ def _calculate_new_wcs(wcs: gwcs.wcs.WCS,
...
     )
 
+    if crpix is None:
+        output_bounding_box = bbox
+    else:
+        output_bounding_box = []
+        for axis_range, minval, shift in zip(bbox, axis_min_values, shifts):
+            output_bounding_box.append(
+                (
+                    axis_range[0] + shift + minval,
```
kmacdonald-stsci, line 326: In the docstring for `_calculate_offsets` it says `shifts` are to be subtracted, but they are added here.  Is the doctrstring wrong or is the calculation wrong?

kmacdonald-stsci (approved):
The new return tuple from `_calculate_offsets` needs to be verified the docstring and computation are correct.  Other than that, it looks good.

src/stcal/alignment/util.py
mcara: Neither is wrong: basically `shifts` are equivalent to 0-indexed CRPIX and should be subtracted from input coordinates to the transformation. However, here we are adjusting the bounding box coordinates (not input coordinates to the WCS transform) so that it "moves" with the CRPIX value. Currently bounding box stays glued to the (0, 0) corner regardless of the output image shape or CRPIX location.

changes/326.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Fix a bug in ``alignment.util.wcs_from_sregions()`` and
+``alignment.util.wcs_from_footprints()`` functions due to which, when the
+``crpix`` argument was provided, the bounding box of the created WCS was
+not adjusted to account for this ``crpix`` value.
```
emolter, line 4:
```suggestion
Fix a bug in ``alignment.util.wcs_from_sregions()`` and
``alignment.util.wcs_from_footprints()`` functions that was causing incorrect WCS bounding boxes
when the ``crpix`` argument was provided.
```

src/stcal/alignment/util.py
```diff
@@ -227,13 +227,18 @@ def _calculate_offsets(fiducial: tuple,
...
 
     Returns
     -------
     ~astropy.modeling.Model
         A model with the offsets to be added to the WCS's transform.
 
+    tuple
+        A tuple of offset *values* that are to be *subtracted* from input
+        coordinates (negative values of offsets in the returned transform).
+        Note: these are equivalent to an effective 0-indexed "crpix".
```
emolter, line 240: Does this mean that these offsets are always equal to crpix?  If so, what is the point of returning this, if crpix is an input to the function?  If not, can you clarify this note a little bit?

tests/test_alignment.py
```diff
@@ -30,7 +30,7 @@ def _create_wcs_object_without_distortion(
     shape,
 ):
     # subtract 1 to account for pixel indexing starting at 0
-    shift = models.Shift() & models.Shift()
+    shift = models.Shift(0) & models.Shift(0)
```
emolter, line 33: Why was this change necessary?  Does this shift actually subtract 1 as the comment indicates?

emolter (commented):
Looks good but just a few small questions

tests/test_alignment.py
mcara: Oh, this was a leftover from testing

mcara: The comment is incorrect either way. The default value for the shift is 0 anyway. So my change does not have any effect. I'll undo it.

mcara:
The bigger issue is: should the bounding box of the resampled image follow (i.e., be centered) at CRPIX or should it always be fixed to the bottom-left corner of the resampled image.

The difference between the two approaches:

1. bbox follows crpix: If user sets a large output shape for the resampled image and places rcpix at the middle then most (valid) data will be centered at the center of the output. User can control where data will be by modifying crpix.
2. bbox is fixed: Then no matter what output shape user chooses, the input data will appear only in the bbox which is situated in the lower-left corner of the resampled image (when `with_bounding_box` for inverse WCS is enabled). By changing CRPIX, user shifts where data fall in the output but the bbox is not shifted and so it acts as a mask. So, as an example, if we increase CRPIX, we will move data to the right and there will be an empty region (no data) near the origin (BLC) but the data will not be "full" as bbox will cut off data on the right/top side. That is, we will be moving data out of the "view".

I think currently 2nd approach is implemented (albeit with a bug). 1st approach seems to be more intuitive as it allows control of where data will be in the output image and it is intuitive to estimate this based on the output shape. In the 2nd approach output shape larger than bbox makes no sense at all as it will be filled with 0s or NaN.

I know this is kind of a philosophical question but it is important to make a decision and based on this I will know how to fix https://github.com/spacetelescope/jwst/pull/9017.

CC: @nden @emolter @melanieclarke

mcara:
I have addressed most (all) comments, I believe. Along the way, fixed a typo in ramp test.

mcara:
Just to be clear, this PR takes approach `#1` in my previous (-2) [comment](https://github.com/spacetelescope/stcal/pull/326#issuecomment-2551935710). I think it is more intuitive (to the user) and kind of makes sense: bbox is moving with CRPIX thus allowing user to control where in the image the data is resampled.

With the other approach, both `output_shape` and `crpix` are pretty much useless since resampled pixels are contained only within the bounding box _that is forced to stay in the bottom-left corner_ of the image. Modifying `crpix` in this case only can move valid pixels out of the bounding box. One problem is that we do not know (in the resample step) what this bounding box is and thus we don't know where to choose CRPIX. With a poor initial choice, `pixmap` is all `nan` and resample crashes before we can see what was the computed bounding box. It's counterintuitive.

mcara:
Oh, there is a third idea:

3. Make the bounding box using the entire array shape. First, drizzle figures out overlaps as it resamples. Bounding boxes on the output images do not help much, unless that is what the user wants. Second, output WCS usually is a distortion-free WCS that is easily invertible so guarding against out-of-domain values is not necessary here. So, users can provide a bounding box with their custom WCS if that is what they want but in an automated pipeline, why impose one at all?

This actually makes me think that approach `#1` is probably the better one if we are going to compute a restrictive bounding box in the pipeline. Or option `#3`.

WilliamJamieson (approved):

src/stcal/alignment/util.py
```diff
@@ -302,19 +303,49 @@ def _calculate_new_wcs(wcs: gwcs.wcs.WCS,
...
+                if upper < 1:
+                    log.warning(
+                        "Input images do not overlap with created WCS. "
+                        "Consider adjusting crval and/or crpix values."
+                    )
+                    log.warning(
+                        "Setting minimum array dimension for axis %d to 10."
+                        % (len(output_bounding_box) - k)
+                    )
+                    upper = 10
```
emolter, line 347: why is 10 the default here?

In what cases would upper be less than 1?  Should this ever happen for valid values of crval and crpix?  I thought this PR attempted to fix this lack of overlap for any reasonable input WCS.

emolter (commented):
looks good to me Mihai, I just raised one more question.   But before we merge it I'd like to check whether it indeed fixes the issue that we have now bypassed via Nadia's PR for AL-852 https://github.com/spacetelescope/jwst/commit/9f1bbbd67db92ea588453911be3c0b116de56cf9.  I'll get back to you hopefully later today about that

src/stcal/alignment/util.py
mcara: This PR initially attempted to fix lack of overlap _due to a bug_ in the code: applying `crpix` to coordinate transforms in WCS but not to the bounding box. However, some tests in `romancal` failed when the following two conditions occurred:
1. the combination of `crval` is such that all input images create an outline that is way outside of the bounding box (towards the negative coordinates); AND
2. shape is not specified.

When `crpix` is provided (but `size` is not), it could happen that only a small part (or none) is in the quadrant with all positive coordinates. Also, the bounding box could be shifted to the right and up (from the origin of the image coordinate system). In this case the size should be determined from the upper bounding box coordinates and not from the bounding box width and height. However, if the bounding box is, for example, in the quadrant with all negative coordinates, then what should be the size of the output image. I think `drizzle` needs a minimum of 2 or 3 pixels. 10 is a small enough number that it will not use a lot of resources but we can lower it to let's say 3.

Or would you prefer to raise an exception and quit?

mcara: Hm, I think that is how `size` should be determined at all times (from the largest coordinate values of the bounding box): it just happens that the lower limit of the bounding box is always 0 when `crval` is not specified.

emolter: Ok here is my understanding so far.  I still don't understand WCS transforms very well so please pardon me if anything here is obviously wrong:

- This PR is making it so that the `bounding_box` is updated to respect `crpix != 0`, which means that the forward and reverse transforms should be self-consistent and will respect `bounding_box`.
- there's no reason why `crpix` and `crval` cannot be outside the image itself, but `bounding_box` should always be around the image
- if `bounding_box` is nowhere near the image, then there is something wrong with either `bounding_box` or `crpix` and `crval`

Are these three statements true?
If so, isn't (1) an instance of a bad WCS, since it's making a `bounding_box` that isn't in the image?

mcara: > * This PR is making it so that the `bounding_box` is updated to respect `crpix != 0`, which means that the forward and reverse transforms should be self-consistent and will respect `bounding_box`.

_"forward and reverse transforms should be self-consistent"_ - that is something that _always_ should be happening. _"will respect `bounding_box`"_ this depends on the settings such as `with_bounding_box` (for the inverse) and I think always for the forward transform. This PR does not deal with this.

> * there's no reason why `crpix` and `crval` cannot be outside the image itself, but `bounding_box` should always be around the image

First part of the statement is correct. Second part is correct too _if_ one wants to get finite numbers (not `nan`) for the world coordinates _for pixels in an actual image described by said WCS object_ or if one wants to get any data in the resampled image. If the bounding box is not in the image then pixel coordinates of that image's pixels will not map to finite world coordinates.

> * if `bounding_box` is nowhere near the image, then there is something wrong with either `bounding_box` or `crpix` and `crval`

"something wrong" - it depends. If this is the result of our computations for default parameters - then yes. If it is the result of the user's input - then it's user's choice (maybe by mistake or maybe not ... who knows?

> Are these three statements true? If so, isn't (1) an instance of a bad WCS, since it's making a `bounding_box` that isn't in the image?

What is (1) referring to? A WCS cannot make a `bounding_box`. We create a bounding box to enclose the data. That is what the current code does _before_ adding a shift to the detector coordinates to account for `crpix`.

At least for the case of resampled image, the bounding box should enclose the region of the image/detector coordinate system where all input images will be resampled (projected) but it also should be trimmed to the actual size of the images (a new thing to add to the code).

emolter: > "forward and reverse transforms should be self-consistent" - that is something that always should be happening. "will respect bounding_box" this depends on the settings such as with_bounding_box (for the inverse) and I think always for the forward transform. This PR does not deal with this.

On and around [this line](https://github.com/spacetelescope/jwst/blob/2968d3f144f945a483f8c92fc67e4c52d11a91df/jwst/resample/resample_utils.py#L172) it appears that currently the `forward_transform` and `backward_transform` do _not_ round-trip if the `bounding_box` is not None.  I thought this PR was supposed to fix that.

> if one wants to get finite numbers (not nan) for the world coordinates for pixels in an actual image described by said WCS object or if one wants to get any data in the resampled image. If the bounding box is not in the image then pixel coordinates of that image's pixels will not map to finite world coordinates.

Can you give me an example of when getting NaN for the world coordinates on purpose, or getting no data in the resampled image, would be the desired behavior?

> If it is the result of the user's input - then it's user's choice (maybe by mistake or maybe not ... who knows?

Along the same lines, under what circumstances would this not be a mistake?

> What is (1) referring to? A WCS cannot make a bounding_box. We create a bounding box to enclose the data. That is what the current code does before adding a shift to the detector coordinates to account for crpix.

It was referring to your previous message, specifically:

> the combination of crval is such that all input images create an outline (bounding box) that is way outside of the output image (towards the negative coordinates);

I know that the WCS doesn't make the bounding box.  I meant, "the WCS is defined such that the bounding box has no overlap with the image"

mcara: > On and around [this line](https://github.com/spacetelescope/jwst/blob/2968d3f144f945a483f8c92fc67e4c52d11a91df/jwst/resample/resample_utils.py#L172) it appears that currently the `forward_transform` and `backward_transform` do _not_ round-trip if the `bounding_box` is not None. I thought this PR was supposed to fix that.

You say "currently the `forward_transform` and `backward_transform` do _not_ round-trip". To me "round-tripping is that the following is true:

```python
world = forward_transform(x, y)
assert np.allclose(backward_transform(*world), (x, y))
```

Is this condition not satisfied (modulo `nan`)? I think it still should be. The issue is that the bounding box is off when `crpix` is specified by the user and hence `nan`s in the output. I believe this PR fixes the issue and so those lines in the `reproject` will not be needed.

> Can you give me an example of when getting NaN for the world coordinates on purpose, or getting no data in the resampled image, would be the desired behavior?

No. And that is the issue! Current code creates a bounding box (when `crpix` is defined) that may not be in the resampled => no data.

> Along the same lines, under what circumstances would this not be a mistake?

I don't know.

emolter: ok, thanks for all the explanations, I think I understand now.  The bounding box does not modify the transform in any way, but it does set the results of the transform to NaN outside the box.

Going all the way back to the original question, if we can't think of any real use-cases for a bounding box that is entirely outside the data, then IMO it would be better to raise an exception than log a warning.

I will admit also that some of my confusion came because I made a typo when removing those lines in `resample_utils.reproject()` and it led to a bunch of test failures.  The branch I made (typo fixed) is [here](https://github.com/spacetelescope/jwst/compare/main...emolter:jwst:mihai-bbox), and I'm running our full regtest suite on those changes [here](https://github.com/spacetelescope/RegressionTests/actions/runs/12434172392), pointing to this PR branch as the stcal dependency.  I'm ready to approve this as long as those tests come back clean

emolter: If you did change this `log.warning()` to an error, would that cause the Romancal tests to fail again?

emolter:
starting regtests for JWST here: https://github.com/spacetelescope/RegressionTests/actions/runs/12438347536

src/stcal/alignment/util.py
emolter: Can you help me understand why there are regression test failures in the run I linked above?

mcara:
I made somewhat major changes to the code in ed425888d57d0fa67c2611a9be1a66f8ffc0803c and combined two functions into one. I think it is much clearer this way. I also made the bounding box be on pixel edge (int coord +- 0.5) because even if footprints of input images when projected to the "resampled" WCS cover half of a resampled pixel, in order to get those input data into that "corner" or "edge" resampled pixel, the bounding box should be at the outermost edge of the pixel so as to guarantee that `pixmap` will not be `NaN` at the footprint vertices.

Also, something is not right with `test_wcs_from_footprints()`: how can output image be 4x4 and WCS not be `NaN` outside of it at coordinate (4, 4) - see, i.e., https://github.com/spacetelescope/stcal/blob/fc7183a367466ed711fb1cc916a1b1fa54943569/tests/test_alignment.py#L212? I fixed the test in ed425888d57d0fa67c2611a9be1a66f8ffc0803c.

Regression tests after restoring the old behavior in 77a8176ddb07df03448b2d5c08e43313559d3d45: https://github.com/spacetelescope/RegressionTests/actions/runs/12454766437 (just one or two failures that do not look to me are caused by the code changes that preserve old bounding box behavior).

After removing the code that was reproducing the old behavior in d0a4a7617cb819f619675dd99ea677f06ec3c9ee, I started regression tests here: https://github.com/spacetelescope/RegressionTests/actions/runs/12457411084 I expect 0-4% pixel to have different values in the resampled images depending on output bounding box correction and image size.

src/stcal/alignment/util.py
mcara: > Can you help me understand why there are regression test failures in the run I linked above?

probably my prs were not rebased. 

I removed log warnings and now raise an exception.

mcara:
From the last regression test the only failure that I see that can be ascribed to this PR is:
```
Keyword S_REGION has different values
```
That is to be expected.

Another failure in _both_ "old" boundary box and "new" boundary box code is in `test_nircam_image_sirs`:
```
         Headers have different number of cards:
          a: 197
          b: 195
         Extra keyword 'R_SIRSKL' in a: 'crds://jwst_nircam_sirskernel_0001.asdf'
         Inconsistent duplicates of keyword ''      :
          Occurs 20 time(s) in a, 19 times in (b)
         Keyword         [16] has different values:
            a> NIR Optimized Convolution Kernel reference file information
            b> Mask reference file information
         ....................................
```
I do not see how this failure could be caused by this PR. It is probably due to some reference file change.

emolter:
Just getting back to this after vacation.  Can you please fix the check-types failure?  Also, why are the downstream Romancal tests failing and is there an associated fix for that?

mcara:
`romancal` tests will be fixed via https://github.com/spacetelescope/romancal/pull/1584

emolter:
Sounds good, so we should hold off on merging this one until the Romancal testing one is merged, right?  The Romancal one appears not to break anything even though it's still pointed to stcal/main.

What is going on with the saturation and dark current unit tests for Romancal?  I see lots of failures there too, are those unrelated?  I don't see them on Romancal main though, at least not in the spot I checked.

mcara:
> What is going on with the saturation and dark current unit tests for Romancal? I see lots of failures there too, are those unrelated? I don't see them on Romancal main though, at least not in the spot I checked.

I don't see how they could be related to this PR.

schlafly:
FYI, romancal has a lot of unit test issues at the moment due to trying to merge some PRs that remove units from the Roman reference files; not related to this PR, obviously!

emolter (approved):
LGTM.  This will _not_ require any JWST regtest ok-ifying until we make the change to JWST that removes Nadia's hot-fix bypass of the reverse transform, is that correct?

mcara:
Starting another set of JWST regression tests here: https://github.com/spacetelescope/RegressionTests/actions/runs/12897275821

mcara:
Since this PR changes how the bounding box is computed, I would expect failures in the regression tests that specify custom `crpix` due differences in the computed footprints. So this will need OKifying, I think.

mcara:
And another regression test: https://github.com/spacetelescope/RegressionTests/actions/runs/13299327946

There are 12 failures all related to changes in `S_REGION` as expected.

mcara:
Here is an example of resampling 6 dithered NIRCAM NRCA1 images with

```
'output_shape': (5000, 5000),
'crpix': (2500, 2500),
```
The red bounding box is with the current code on `main` and the green bounding box is computed with the code from this PR:
bbox_with_and_without_stcal_pr

src/stcal/alignment/util.py
```diff
@@ -157,28 +164,74 @@ def _get_axis_min_and_bounding_box(footprints: list[np.ndarray],
     ref_wcs : ~gwcs.wcs.WCS
         The reference WCS object.
 
+    fiducial : tuple
+        A tuple containing the world coordinates of the fiducial point.
+
+    crpix : list or tuple, optional
```
nden, line 170: What would be a good name for this keyword that will distinguish it from FITS? 
pixel_fiducial, pixel_reference?
Perhaps `pixel_fiducial` and `sky_fiducial`?

mcara: I would have preferred that `crpix` be `None` by default and take the middle of the footprint instead of one of the corners.

mcara: > What would be a good name for this keyword that will distinguish it from FITS? pixel_fiducial, pixel_reference? Perhaps `pixel_fiducial` and `sky_fiducial`?

`crpix0`

mcara:
Roman Regression tests started here: https://github.com/spacetelescope/RegressionTests/actions/runs/13317570932

nden (approved):

braingram:
Were the romancal regression test differences okified? I see many similar failures on romancal main with stcal main:
https://github.com/spacetelescope/RegressionTests/actions/runs/13378092074
some from tests added in https://github.com/spacetelescope/romancal/pull/1596 by @stscieisenhamer (which was merged around the same time as these changes).

mcara:
probably not, if they still persist on main

PR #331, opened 2025-01-03
Saturation speedup

t-brandt:
Resolves [JP-3820](https://jira.stsci.edu/browse/JP-3820)

Closes #

This PR improves the performance of the saturation step, reducing its runtime by as much as a factor of 5-10 while leaving results unchanged.  It does this primarily with three changes. First, when a pixel saturates, that saturation is propagated into all subsequent reads using a running boolean tally of pixels that have saturated.  Second, ndimage.binary_dilation is very slow, an order of magnitude slower than writing the operations out explicitly.  Finally, if there is no averaging in groups, there is no point in the diluted saturation check; this check is skipped if the dilution factor is 1.

## Tasks

</details>

### DIFF
--- a/src/stcal/saturation/saturation.py
+++ b/src/stcal/saturation/saturation.py
@@ -83,31 +83,45 @@ def flag_saturated_pixels(
 
     for ints in range(nints):
         # Work forward through the groups for initial pass at saturation
+
+        # We want to flag saturation in all subsequent groups after
+        # the one in which it was found.  Use this boolean array to
+        # keep a running tally of pixels that have saturated.
+        previously_saturated = np.zeros(data[0, 0].shape, dtype='bool')
+
         for group in range(ngroups):
             plane = data[ints, group, :, :]
 
-            flagarray, flaglowarray = plane_saturation(plane, sat_thresh, dqflags)
-
             # for saturation, the flag is set in the current plane
             # and all following planes.
-            np.bitwise_or(gdq[ints, group:, :, :], flagarray, gdq[ints, group:, :, :])
+
+            # Update the running tally of all pixels that have ever
+            # experienced saturation to account for this.
+
+            previously_saturated |= (plane >= sat_thresh)
+            flagarray = (previously_saturated * saturated).astype(np.uint32)
+
+            gdq[ints, group, :, :] |= flagarray
 
             # for A/D floor, the flag is only set of the current plane
-            np.bitwise_or(gdq[ints, group, :, :], flaglowarray, gdq[ints, group, :, :])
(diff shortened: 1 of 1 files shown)


src/stcal/saturation/saturation.py
```diff
@@ -83,64 +83,85 @@ def flag_saturated_pixels(
 
     for ints in range(nints):
         # Work forward through the groups for initial pass at saturation
+
+        # We want to flag saturation in all subsequent groups after
+        # the one in which it was found.  Use this boolean array to
+        # keep a running tally of pixels that have saturated.
+        previously_saturated = np.zeros(data[0, 0].shape, dtype='bool')
```
kmacdonald-stsci, line 90: I think it's clearer to use `shape=(row,col)`, rather than `shape=data[0,0].shape`.

kmacdonald-stsci (approved):

drlaw1558:
Just tested this empirically as well.  Can confirm that the groupdq array is unaffected both for an ungrouped case (jw01864005001_03102_00001_mirimage) and a grouped IRS2 case (jw01908004001_02101_00005_nrs1, used for previous testing of https://github.com/spacetelescope/jwst/pull/8593), with appreciable runtime gains in the long-integration limit (17 seconds to 3 seconds in the 292 group MIRI case here).

Also looks like the slow binary_dilation behavior has been a known issue for the last few years: https://github.com/scipy/scipy/issues/13991

src/stcal/saturation/saturation.py
t-brandt: Sounds good; I changed it.

changes/331.general.rst
```diff
@@ -0,0 +1,2 @@
+- Performance improvements for saturation step. (`#331
+  <https://github.com/spacetelescope/stcal/issues/331>`_)
```
braingram, line 2:
```suggestion
Performance improvements for saturation step
```
towncrier will add the link and bullet when this changelog fragment is consumed (the next build). Accepting this suggestion should fix the failing docs.

melanieclarke:
It looks like we lost track of this one, sorry!  Is there anything more to do from your side, @t-brandt?  If not, I can start regression tests.

t-brandt:
I'm all set with this one, thanks @melanieclarke!

melanieclarke:
Romancal tests here:
https://github.com/spacetelescope/RegressionTests/actions/runs/13270137990

Tagging @schlafly for review: it looks like this code is used in romancal.

melanieclarke:
Both regression test runs are clean.

schlafly (approved):
Thanks!

melanieclarke (approved):
Tests are passing and we have approval from jwst and romancal - I will go ahead and merge now.  Thanks all!

PR #333, opened 2025-01-21
JP-3686: Allow tweakreg to parse skycoord objects in absolute refcat

emolter:
Resolves [JP-3686](https://jira.stsci.edu/browse/JP-3686)

Closes spacetelescope/jwst#8639

This PR allows the tweakreg step to parse absolute reference catalogs that resemble the output from the JWST `source_catalog` step.  That step reports absolute RA, DEC in columns called `sky_centroid.ra` and `sky_centroid.dec` which are parsed by into an astropy Table as a single column containing a SkyCoord.  This PR adds the functionality to re-cast this into RA/DEC columns as expected by `tweakwcs.align_wcs`.

## Tasks

</details>

### DIFF
--- a/src/stcal/tweakreg/tweakreg.py
+++ b/src/stcal/tweakreg/tweakreg.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import numpy as np
+from collections import Counter
 import math
 import warnings
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


emolter:
@mcara this isn't ready for formal review yet, but I'm wondering whether you'd like this change (or a version of it) in tweakwcs instead, or if here is the best place for it in your opinion.

src/stcal/tweakreg/tweakreg.py
```diff
@@ -253,14 +254,43 @@ def _parse_refcat(abs_refcat: str | Path,
...
 
 
+def _parse_sky_centroid(catalog: Table) -> Table:
+    """Turn SkyCoord object into simple RA/DEC columns.
+
+    The inclusion of SkyCoord objects via sky_centroid.ra and sky_centroid.dec
+    permits the use of catalogs directly from the jwst source_catalog step.
+    No action is taken if the catalog already contains RA and DEC columns.
+    """
+    cols = [name.lower() for name in catalog.colnames]
```
mcara, line 272: Can a table contain `RA`, `rA`, `Ra`, and `ra` columns at the same time? If "Yes", then I think this code should be enhanced and throw an exception if there are multiple instances of `ra` in `cols` list. While it may never happen, we must catch something like this.

mcara: Or, maybe implement a logic like: if `RA` and `DEC` are present - use those, if not try `ra` and `dec`, if not use `lower()` and try again...

src/stcal/tweakreg/tweakreg.py
```diff
@@ -253,14 +254,43 @@ def _parse_refcat(abs_refcat: str | Path,
...
+    if ("ra" in cols) and ("dec" in cols):
+        if "sky_centroid" in cols:
+            msg = ("Catalog contains both (RA, DEC) and sky_centroid. "
```
mcara, line 275: is it `contains both (RA, DEC)` or maybe `contains both (ra, dEc)`?

mcara:
> @mcara this isn't ready for formal review yet, but I'm wondering whether you'd like this change (or a version of it) in tweakwcs instead, or if here is the best place for it in your opinion.

I prefer to keep things simple. It is documented what `tweakwcs` expects and supporting various column names only complicates things. Obviously JWST `source_catalog` step was not designed with reusability of those catalogs (by `tweakreg`). I wonder if that could be changed instead of the tweakreg step. In any case, I think any changes should go to `stcal` and/or individual pipelines.

src/stcal/tweakreg/tweakreg.py
emolter: good suggestion, this is changed in the most recent version.  I think it's best for the step to fail if it sees e.g. `RA` and `ra`, so that the user doesn't accidentally get unexpected results from the wrong column.

src/stcal/tweakreg/tweakreg.py
emolter: I updated the error message a bit, let me know if you are ok with the change

emolter:
> I prefer to keep things simple. It is documented what tweakwcs expects and supporting various column names only complicates things. Obviously JWST source_catalog step was not designed with reusability of those catalogs (by tweakreg). I wonder if that could be changed instead of the tweakreg step. In any case, I think any changes should go to stcal and/or individual pipelines.

In that case, I think this is the best place for it.  I think changing the output format of the source catalog step would be more disruptive, e.g., if users have already written workflows that ingest this information.  Changing the columns in the output to source catalog would also make it less Astropy-integrated, which isn't necessarily desired either.  I don't expect this to "blow up" in the sense that we will need to support many different formats of input catalog - this is a special case and should be the only one.

mcara:
Another suggestion for rare cases such as this and maybe setting sky background, etc. is to have `jwst/tools` or something like that with scripts or functions that can be called and rename table columns or set sky background on a list of files, etc.

mcara (approved):

kmacdonald-stsci (approved):

tapastro (approved):
Looks good to me - happy to see our steps work together more coherently.

PR #337, opened 2025-02-04
Jump performance

t-brandt:
Resolves [JP-3697](https://jira.stsci.edu/browse/JP-3697)

Closes #337

A number of changes to `twopoint_difference.py` and `jump.py` reduce memory usage by a factor of about 2 and improve runtime by a factor of anywhere from 3 to 20.  Tested output files (noted below) are unchanged, except for MIRI shower detection in `jw01701012001_02105_00001_mirifulong`.  This is due to a difference in replacing infilled saturated pixels after performing a convolution; the new version identifies fewer showers.  An example comment on this is here: 
https://jira.stsci.edu/browse/JP-3697?focusedId=760740&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-760740
Overflagging of jumps should no longer cause the pipeline to grind to a halt due to `np.where` combined with `for` loops over pixels; it should now have virtually no impact on performance.

Files tested, along with memory usage and runtimes in the jump step (all files are public except for program 5263):

| File                                                                   |  Old Runtime | Old Memory  |   New Runtime | New Memory |
| -------------------------------------------| --------| ---------| ----| -- |
| jw01668002001_04101_00001_mirimage   | 68 s | 27 GB | 15 s | 11 GB |
| jw01283001001_03101_00001_mirimage    | 420 s | 70 GB | 120s  | 30 GB |
| jw05263001001_05101_00001_nrs1             | 500 s | 40 GB | 30 s | 23 GB |
| jw02079004001_03201_00001_nrca1          | 35 s | 2 GB | 6 s | 2 GB |
| jw01701012001_02105_00001_mirifulong     | 13 s | 0.4 GB | 4 s | 0.4 GB |
| jw01366001001_04101_00001-seg001_nis | 155 s | 40 GB | 35 s | 23 GB |

A significant fraction of the line changes are due to fixing a mixture of 8-space indents and 4-space indents in `twopoint_difference.py`.  Most of these lines were not actually changed.

## Tasks

</details>

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -619,32 +619,6 @@ def extend_ellipses(
 
     return gdq_cube, num_ellipses
 
-def find_last_grp(grp, ngrps, num_grps_masked):
-    """
-    Find the last group based on current group and number of groups masked.
-
-    Parameters
-    ----------
-    grp : int
-        The location of the shower
-
-    ngrps : int
-        The number of groups in the integration
-
-    num_grps_masked : int
-        The requested number of groups to be flagged after the shower
-
-    Returns
-    -------
-    last_grp : int
-        The index of the last group to flag for the shower
-
-    """
-    num_grps_masked += 1
-    last_grp = min(grp + num_grps_masked, ngrps)
-    return last_grp
-
-
 def find_ellipses(dqplane, bitmask, min_area):
     """
     Find ellipses based on DQ masks in bitmask.
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -620,6 +620,31 @@ def extend_ellipses(
 
     return gdq_cube, num_ellipses
 
+def find_last_grp(grp, ngrps, num_grps_masked):
+    """
+    Find the last group based on current group and number of groups masked.
+
+    Parameters
+    ----------
+    grp : int
+        The location of the shower
+
+    ngrps : int
+        The number of groups in the integration
+
+    num_grps_masked : int
+        The requested number of groups to be flagged after the shower
+
+    Returns
+    -------
+    last_grp : int
+        The index of the last group to flag for the shower
+
+    """
+    num_grps_masked += 1
+    last_grp = min(grp + num_grps_masked, ngrps)
+    return last_grp
+
 
 def find_ellipses(dqplane, bitmask, min_area):
     """
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -461,8 +461,8 @@ def extend_saturation(cube, grp, sat_ellipses, jump_data, persist_jumps):
 
             alpha = ellipse[2]
 
-            indx, sat_ellipse = ellipse_subim(ceny, cenx, axis1, axis2,
-                                              alpha, satcolor, (nrows, ncols))
+            indx, sat_ellipse = ellipse_subim(
+                ceny, cenx, axis1, axis2, alpha, satcolor, (nrows, ncols))
             (iy1, iy2, ix1, ix2) = indx
 
             # Create another non-extended ellipse that is used to
@@ -474,8 +474,8 @@ def extend_saturation(cube, grp, sat_ellipses, jump_data, persist_jumps):
                 cube[i][iy1:iy2, ix1:ix2][is_sat] = jump_data.fl_sat
 
             ax1, ax2 = (ellipse[1][0], ellipse[1][1])
-            indx, persist_ellipse = ellipse_subim(ceny, cenx, ax1, ax2,
-                                                  alpha, satcolor, (nrows, ncols))
+            indx, persist_ellipse = ellipse_subim(
+                ceny, cenx, ax1, ax2, alpha, satcolor, (nrows, ncols))
             (iy1, iy2, ix1, ix2) = indx
 
             persist_mask = persist_ellipse == satcolor
@@ -591,7 +591,6 @@ def extend_ellipses(
     #  expanded ellipses of pixels with
     # the jump flag set.
     _, ngroups, nrows, ncols = gdq_cube.shape
-    image = np.zeros(shape=(nrows, ncols, 3), dtype=np.uint8)
     num_ellipses = len(ellipses)
     for ellipse in ellipses:
         ceny = ellipse[0][0]
@@ -604,8 +603,8 @@ def extend_ellipses(
         # indices that place this subimage within the full array.
         axis1 = axes[0]*2
         axis2 = axes[1]*2
-        indx, jump_ellipse = ellipse_subim(ceny, cenx, axis1, axis2,
-                                           alpha, jump_data.fl_jump, (nrows, ncols))
+        indx, jump_ellipse = ellipse_subim(
+            ceny, cenx, axis1, axis2, alpha, jump_data.fl_jump, (nrows, ncols))
         (iy1, iy2, ix1, ix2) = indx
         
         # Propagate forward by num_grps_masked groups.
@@ -1155,7 +1154,8 @@ def get_bigcontours(ratio, intg, grp, gdq, pdq, jump_data, ring_2D_kernel):
 
 
 def convolve_fast(inarray, kernel, copy=False):
-    """Faster version of astropy.convolution.convolve(preserve_nan=True)
+    """Convolve an array with a kernel, interpolating over NaNs.
+    Faster version of astropy.convolution.convolve(preserve_nan=True)
     Parameters
     ----------
     inarray : 2D array of floats
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -30,12 +30,12 @@ def detect_jumps_data(jump_data):
     turn.
 
     Note that the detection methods are currently set up on the assumption
-    that the input science and error data arrays will be in units of
+    that the input science data array will be in units of
     electrons, hence this routine scales those input arrays by the detector
     gain. The methods assume that the read noise values will be in units
     of DN.
 
-    The gain is applied to the science data and error arrays using the
+    The gain is applied to the science data array using the
     appropriate instrument- and detector-dependent values for each pixel of an
     image.  Also, a 2-dimensional read noise array with appropriate values for
     each pixel is passed to the detection methods.
@@ -67,7 +67,7 @@ def detect_jumps_data(jump_data):
 
     pdq = setup_pdq(jump_data)
 
-    # Apply gain to the SCI, ERR, and readnoise arrays so they're in units
+    # Apply gain to the SCI and readnoise arrays so they're in units
     # of electrons
     data = jump_data.data * jump_data.gain_2d
     gdq = jump_data.gdq
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -71,7 +71,6 @@ def detect_jumps_data(jump_data):
     # of electrons
     data = jump_data.data * jump_data.gain_2d
     gdq = jump_data.gdq
-    # err = jump_data.err * jump_data.gain_2d
     readnoise_2d = jump_data.rnoise_2d * jump_data.gain_2d
 
     # also apply to the after_jump thresholds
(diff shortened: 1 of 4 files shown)


src/stcal/jump/jump.py
```diff
@@ -66,11 +67,10 @@ def detect_jumps_data(jump_data):
 
     pdq = setup_pdq(jump_data)
 
-    # Apply gain to the SCI, ERR, and readnoise arrays so they're in units
+    # Apply gain to the SCI and readnoise arrays so they're in units
     # of electrons
     data = jump_data.data * jump_data.gain_2d
     gdq = jump_data.gdq
-    # err = jump_data.err * jump_data.gain_2d
     readnoise_2d = jump_data.rnoise_2d * jump_data.gain_2d
```
melanieclarke, line 74: Can we do these multiplications in place so we don't make copies of the data and readnoise arrays?

src/stcal/jump/twopoint_difference.py
```diff
@@ -275,241 +270,200 @@ def find_crs_old(
...
-                        jumpy, jumpx = np.where(gdq[integ, grp, :, :] == jump_flag)
-                        gdq[integ, grp, jumpy, jumpx] = 0
+                    if np.all(gdq[integ, grp] & (jump_flag | dnu_flag) != 0):
+                        gdq[integ, grp] |= dnu_flag
+                        
+                    #if np.all(np.bitwise_or(np.bitwise_and(gdq[integ, grp, :, :], jump_flag),
+                    #                        np.bitwise_and(gdq[integ, grp, :, :], dnu_flag))):
+                    #    jumpy, jumpx = np.where(gdq[integ, grp, :, :] == jump_flag)
+                    #    gdq[integ, grp, jumpy, jumpx] = 0
+
```
melanieclarke, line 302: We discussed this offline, but to add here for other readers...  

I think the logic here is equivalent to:
- Original: if all flags in the group are either jump or DNU, set the ones that are only jump to good.
- New: if all flags in the group are either jump or DNU, make sure they are all DNU.

It's not clear to me if the intent was what the comment says, to set fully flagged groups to DNU, or what the code was actually doing. It seems possible the intent was to preserve a value for the group from the jump-only pixels if the sigma clipping was overzealous.

melanieclarke (commented):
I'll review in more detail soon, but I have a couple initial comments below, then I can kick off regression tests.

src/stcal/jump/jump.py
t-brandt: Yes, though I need to make sure that there are no unintended consequences since the data are modified within the jump step.

t-brandt: Memory use goes down a bit (by 10-15%), but the rateint files differ if I replace
`data = jump_data.data * jump_data.gain_2d`
with
`jump_data.data *= jump_data.gain_2d`
`data = jump_data.data`
I think I'll leave it as-is right now rather than try to track down exactly why this happened: where a copy was not actually a deep copy, etc.

melanieclarke: At the bottom of the function, the gain is divided back out of data.  Did you leave it as is or divide it out of jump_data.data?

t-brandt: If I remove that line without making a copy, the rateint file differs.  That suggests that a deep copy is not actually being made earlier in the jump step (which I think is intended here https://github.com/spacetelescope/jwst/blob/4b6867022496c67e94aab1e2b5c7c535dae0d666/jwst/jump/jump_step.py#L96).  Looking at that line, I am sure that it isn't actually a deep copy and that the arrays are views.

If I leave the line dividing the gain out of the data in, the rateint files still differ, but at the level expected from (a*b)/b != a due to truncation error.

t-brandt: In my opinion, the best thing to do would be to not modify data at all but fold the gain into the photon noise estimate within the jump code.  I tried that a month or two earlier but was having trouble getting identical results to the current approach.  I can try again.

melanieclarke: That line in the jump step is a datamodel copy, which is a deepcopy of the arrays, but the intent there is to not modify the input model.  The `result` model is the one that gets modified in the rest of the jump step and will eventually get used to make the rateints file.

It does seem like it would be nicer not to touch the data if we don't have to, but we can defer that for future improvements if there isn't an easy way to do it here.

t-brandt: Ok, thanks for clarifying that; I think I got it.  It shouldn't be hard to avoid touching the data (famous last words).  I'll give it another brief go.

t-brandt: Aah--there are so many places in there where units are electrons, in terms of the input parameters.  I retract my previous comment on trying to avoid touching the data.  I think we either need to multiply and divide by the gain in-place, accepting very small differences in the rateints files and modifications of the data array inside the jump routine (currently just setting some values to NaN, I think), or leave it all as-is.  The read noise array is small enough that I don't think we need to worry about avoiding that copy.  @melanieclarke @kmacdonald-stsci @drlaw1558 -- do you have an opinion on this?

drlaw1558: Given all of the queued-up changes to the jump step at the moment I'd be inclined to leave it as-is for now and consider further changes after this and other extant PRs are taken care of.

drlaw1558:
Adding a quick note that I've done some initial testing on this too, with good results.  In the case of MIRI TSO observation jw03730013001_03101_00001-seg001_mirimage_uncal.fits the results are identical and jump step runtime goes down from 358 to 156 seconds.

src/stcal/jump/twopoint_difference.py
```diff
@@ -275,241 +270,199 @@ def find_crs_old(
...
+                # compute 'ratio' for each group. this is the value that will be
+                # compared to 'threshold' to classify jumps. subtract the median of
+                # first_diffs from first_diffs, take the abs. value and divide by sigma.
+                # The jump mask is the ratio greater than the threshold and the
+                # difference is usable.  Loop over integrations to minimize the memory
+                # footprint.
+                jump_mask = np.zeros(first_diffs.shape, dtype=bool)
+                for i in range(nints):
+                    jump_candidates = np.abs(first_diffs[i] - median_diffs[np.newaxis, :]) / sigma[np.newaxis, :] > normal_rej_thresh
```
kmacdonald-stsci, line 316: Long line should be broken down into a couple smaller lines.

kmacdonald-stsci (approved):

src/stcal/jump/jump.py
```diff
@@ -524,17 +581,16 @@ def extend_ellipses(
...
     num_ellipses : int
         The number of ellipses passed in as a parameter.
     """
     # For a given DQ plane it will use the list of ellipses to create
     #  expanded ellipses of pixels with
     # the jump flag set.
-    out_gdq_cube = gdq_cube.copy()
-    _, _, nrows, ncols = gdq_cube.shape
+    _, ngroups, nrows, ncols = gdq_cube.shape
     image = np.zeros(shape=(nrows, ncols, 3), dtype=np.uint8)
```
melanieclarke, line 594: It looks like this image cube is unused.

src/stcal/jump/jump.py
```diff
@@ -543,22 +599,26 @@ def extend_ellipses(
...
+            # Only propagate the snowball forward to unsaturated pixels.
 
-    return out_gdq_cube, num_ellipses
+            sat_pix = gdq_cube[intg, flg_grp, iy1:iy2, ix1:ix2] & jump_data.fl_sat
+            jump_ellipse[sat_pix == jump_data.fl_sat] = 0
+            gdq_cube[intg, flg_grp, iy1:iy2, ix1:ix2] |= jump_ellipse
 
+    return gdq_cube, num_ellipses
 
 def find_last_grp(grp, ngrps, num_grps_masked):
```
melanieclarke, line 623: It looks like this function is now unused.  We should remove it, here and in the tests.

src/stcal/jump/jump.py
```diff
@@ -1102,6 +1153,64 @@ def get_bigcontours(ratio, intg, grp, gdq, pdq, jump_data, ring_2D_kernel):
     return bigcontours 
 
 
+
+def convolve_fast(inarray, kernel, copy=False):
+    """Faster version of astropy.convolution.convolve(preserve_nan=True)
```
melanieclarke, line 1158: Can you please rephrase this docstring?  The first line should be a one line summary, in imperative mood, ending in a period.  For example: "Convolve an array with a kernel, preserving NaN values."

src/stcal/jump/jump.py
```diff
@@ -1102,6 +1153,64 @@ def get_bigcontours(ratio, intg, grp, gdq, pdq, jump_data, ring_2D_kernel):
...
+
+    if copy:
+        array = inarray.copy()
+    else:
+        array = inarray
+
+    good = np.isfinite(array)
+    array[~good] = 0
+
+    convolved_array = signal.oaconvolve(array, kernel, mode='same')
```
melanieclarke, line 1191: Doesn't this convolve the zeros with neighboring non-zero data? That seems like it could be problematic. Is that what the astropy method does?

src/stcal/jump/twopoint_difference.py
```diff
@@ -521,6 +474,38 @@ def find_crs_old(
     return gdq, row_below_gdq, row_above_gdq, num_primary_crs, dummy
 
 
+def propagate_flags(boolean_flag, n_groups_flag):
+    """Propagate a boolean flag array npix groups along the first axis.
+    If the number of groups to propagate is not too large, or if a
+    high percentage of pixels are flagged, use boolean or on the
+    array.  Otherwise use np.where.  In both cases operate on the
+    array in-place.
```
melanieclarke, line 482: Just a question - is np.where faster if the number of groups is small?

melanieclarke (commented):
This looks great!  I'd like an answer from on the one flagging logic question, though, and I have a few more questions below.

src/stcal/jump/jump.py
t-brandt: Yes.  I had removed it, but put it back in to avoid failures in the downstream jwst tests.  I guess I should make a corresponding PR in jwst to remove it from the tests?

src/stcal/jump/jump.py
t-brandt: It's identical to the astropy method apart from small truncation differences in the convolution computation.  Both methods interpolate over bad or missing data.

src/stcal/jump/twopoint_difference.py
t-brandt: np.where is faster if there are very few cosmic rays and they are propagated far along the group axis.  If there are not many groups, then this is just a couple of logic operations on the gdq array.  
np.where requires one logic pass through the whole array, the creation of some int64 index arrays, and explicit nested for loops.
The alternative requires jmax logic passes through the whole array.
For np.where to win, we need some combination of a large jmax and a low frequency of jumps so that the nested for loop has few entries.
I set very rough numbers for this in the code.  In practice, I was just making sure that one way or another this step took a negligible fraction of the total time.

src/stcal/jump/jump.py
t-brandt: Whoops, it is in the stcal tests!  Sorry, I will remove it.

src/stcal/jump/jump.py
melanieclarke: Okay, thanks for clarifying.

src/stcal/jump/twopoint_difference.py
melanieclarke: Interesting, thanks.

melanieclarke:
jwst regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/13165362881
romancal regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/13165381062

schlafly (approved):

src/stcal/jump/twopoint_difference.py
t-brandt: Reverted the logic, commented out the line with new logic.

melanieclarke: Looks good, thanks!

melanieclarke:
I think this should be good to go now.  I will re-run the jwst tests one more time for the last small change, then merge.

melanieclarke (approved):

PR #317, opened 2024-11-12
Jump masked groups counting

mwregan2:
Resolves [JP-3793](https://jira.stsci.edu/browse/JP-3793)

This PR addresses a problem in the two point difference code where it determines if there are enough groups for the code to run. The logic was not correct and has now been corrected.

## Tasks
- [ X ] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [ ] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
  - [ ] run regression tests with this branch installed (`"git+https://github.com/<fork>/stcal@<branch>"`)
    - [ ] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) 
    - [ ] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml)

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.apichange.rst``: change to public API
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.general.rst``: infrastructure or miscellaneous change
</details>

PR #338, opened 2025-02-07
JP-3793: Fix numbers of flagged groups

drlaw1558:
Resolves [JP-3793](https://jira.stsci.edu/browse/JP-3793) regarding inaccurate calculation of the number of groups in the jump step.  This PR updates prior https://github.com/spacetelescope/stcal/pull/317 to account for recent unrelated changes.

EDIT: See also [JP-3877](https://jira.stsci.edu/browse/JP-3877).  This PR ensures that long TSO exposures don't fall into the wrong processing type, reducing runtime from over an hour to about 1 minute in the example given there.

### DIFF
--- a/src/stcal/jump/twopoint_difference.py
+++ b/src/stcal/jump/twopoint_difference.py
@@ -225,24 +225,39 @@ def find_crs_old(
     num_flagged_grps = 0
     # determine the number of groups with all pixels set to DO_NOT_USE
     ngrps = dat.shape[1]
+    max_flagged_grps = 0
+    total_flagged_grps = 0
     for integ in range(nints):
+        num_flagged_grps = 0
         for grp in range(dat.shape[1]):
             if np.all(np.bitwise_and(gdq[integ, grp, :, :], dnu_flag)):
                 num_flagged_grps += 1
+        if num_flagged_grps > max_flagged_grps:
+            max_flagged_grps = num_flagged_grps
+        total_flagged_grps += num_flagged_grps
     if only_use_ints:
         total_sigclip_groups = nints
     else:
-        total_sigclip_groups = nints * (ngrps - num_flagged_grps)
-    total_groups = nints * (ngrps - num_flagged_grps)
-    total_diffs = nints * (ngrps - 1 - num_flagged_grps)
-    total_usable_diffs = total_diffs - num_flagged_grps
-    if ((ngrps < minimum_groups and only_use_ints and nints < minimum_sigclip_groups) or
-            (not only_use_ints and nints * ngrps < minimum_sigclip_groups and
-             total_groups < minimum_groups)):
+        total_sigclip_groups = nints * ngrps - num_flagged_grps
+
+    min_usable_groups = ngrps - max_flagged_grps
+    total_groups = nints * ngrps - total_flagged_grps
+    min_usable_diffs = min_usable_groups - 1
+    sig_clip_grps_fails = False
+    total_noise_min_grps_fails = False
+
+    # Determine whether there are enough usable groups the two sigma clip options
+    if (only_use_ints and nints < minimum_sigclip_groups) \
+            or \
+            (not only_use_ints and total_sigclip_groups < minimum_sigclip_groups):
+        sig_clip_grps_fails = True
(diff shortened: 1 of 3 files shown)


tests/test_jump_twopoint_difference.py
```diff
@@ -22,6 +22,89 @@ def _cube(ngroups, nints=1, nrows=204, ncols=204, readnoise=10):
 
     return _cube
 
+def test_sigclip_not_enough_groups(setup_cube):
```
melanieclarke, line 25: This test has three different definitions with the same name, so only the last one will be run.  Testing locally, it looks like the first one fails and the second two pass, as written.

tests/test_jump_twopoint_difference.py
```diff
@@ -22,6 +22,89 @@ def _cube(ngroups, nints=1, nrows=204, ncols=204, readnoise=10):
...
+    data, gdq, nframes, read_noise, rej_threshold = setup_cube(ngroups, nints=nints, nrows=2, ncols=2, readnoise=8)
+    data[0, 0:2, 0:, 0] = 1
+    out_gdq, row_below_gdq, rows_above_gdq, total_crs, stddev = find_crs(
+        data, gdq, read_noise, rej_threshold, rej_threshold, rej_threshold,
+        nframes, False, 200, 10, DQFLAGS,
+        minimum_sigclip_groups=11, minimum_groups=16
+    )
+    assert total_crs != -99
+
+def test_nosigclip_with_enough_groups(setup_cube):
```
melanieclarke, line 70: This test is also defined again later with the same name, so only the second one runs.  Testing locally, it looks like both pass as written.

melanieclarke (commented):
Comparing to the previous PR, it looks like this is a clean copy of the original changes, also incorporating comments from Ken's review.

There are a couple errors in the unit tests that need cleaning up.  When that's done, I can run regression tests.

tests/test_jump_twopoint_difference.py
drlaw1558: Ah, I thought that was odd but wasn't sure whether same-named tests were ok or not.  I've given them unique names now.  Looks like that also fixes why I wasn't able to get any pytests against stcal to run locally.
@mwregan2 Any thoughts on why some of these new tests are failing?

drlaw1558: Sounds good to me; looking at it I agree it doesn't seem to be adding anything.  I've removed the test.

src/stcal/jump/twopoint_difference.py
```diff
@@ -225,24 +225,39 @@ def find_crs_old(
...
+        total_sigclip_groups = nints * ngrps - num_flagged_grps
+
+    min_usable_groups = ngrps - max_flagged_grps
+    total_groups = nints * ngrps - total_flagged_grps
+    min_usable_diffs = min_usable_groups - 1
+    sig_clip_grps_fails = False
+    total_noise_min_grps_fails = False
+
+    # Determine whether there are enough usable groups the two sigma clip options
+    if (only_use_ints and nints < minimum_sigclip_groups) \
```
kmacdonald-stsci, line 250: I recommend using parenthesis, instead of `\` for multiple lines.

kmacdonald-stsci (approved):

melanieclarke:
Regtests here:
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/13294000215
romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/13294013308

drlaw1558:
@melanieclarke It's not surprising, but I'll take a look at the test cases just to make sure.

src/stcal/jump/twopoint_difference.py
drlaw1558: Swapped line formatting.

melanieclarke:
Thanks for checking the diffs @drlaw1558.

This all looks good to me.  Romancal folks should not be impacted by jump step changes, but I'm tagging them for review for their information.

schlafly (approved):
Regression tests pass & we don't use the stcal jump step; approving.

PR #339, opened 2025-02-11
JP-3765: Keep jump flags going into likelihood-based fitting algorithm

drlaw1558:
Partially resolves [JP-3765](https://jira.stsci.edu/browse/JP-3765) by retaining prior jump flags in the likelihood ramp-fitting algorithm.


drlaw1558:
Tagging @t-brandt  to take a look.

t-brandt:
I think the right approach is probably to go with this for now. In the future I think we should explore more selective use of the twopointdifference code in tandem with likelihood-based jump detection.

drlaw1558:
Not sure why I can't tag specific reviewers, so tagging @kmacdonald-stsci here instead.

kmacdonald-stsci (approved):
This looks fine, but I remember having this conversation in the past and deciding the best action is to remove the flagging when LIKELY is selected.  That has been determined to be wrong, now?

I approved this PR, but would like to make sure the answer above is "yes".

drlaw1558:
@kmacdonald-stsci Yes; I think we've learned that there are enough edge cases that the regular jump step has been tuned to handle that it's useful to default to keeping those flags.  They can always be turned off by skipping the jump step, and future work can prioritize more complex interaction between the two if necessary to do so.  This should allow INS testing of the LIKELY method to continue moving ahead.

schlafly:
FWIW, this doesn't affect romancal and is fine on our end.

melanieclarke:
Regression tests here:
jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/13678019808
romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/13678025222

melanieclarke (approved):
Regression tests show no changes for jwst or romancal.  It looks like we do not yet have regression tests that explicitly use the likelihood algorithm.

LGTM for INS testing purposes.  I will add a note to JP-3765 that we will need to add some regression tests for this algorithm.

melanieclarke:
I will take Eddie's comment as approval from the romancal side and go ahead and get this merged.

PR #342, opened 2025-02-20
Add a hook method for add_model in Resample to allow customizations

### LINKED ISSUE #341: Support reusing pixmap for Resample subclasses
Romancal resamples exposure time using the same pixmap as the science arrays.

I don't see any easy way to reuse the pixmap computed in `add_model`:
https://github.com/spacetelescope/stcal/blob/95c1d8fc0dbade35ef38b9ec35822c1dd4b350cd/src/stcal/resample/resample.py#L746
without copying much of the code in the `Resample` class into the subclass in romancal.

Find a way to reorganize `add_model` (or some other way) to allow reuse of the pixmap for romancal exposure time resampling.

mcara:
This PR adds an hook method that is called by `Resample.add_method` to allow subclasses to perform additional processing while reusing computed `pixmap` array and other quantities.

Closes #348
Closes #341

CC: @braingram 

## Tasks

</details>


mcara:
JWST regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/13446821486

Roman regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/13446824076

braingram:
Thanks for putting this together.

xref: https://github.com/spacetelescope/stcal/issues/341

I'd like to wait to make changes to resample here until:
- the initial pass at using this code in romancal is merged: https://github.com/spacetelescope/romancal/pull/1634
- the use of `var_sky` in romancal is sorted: https://github.com/spacetelescope/romancal/pull/1556 (since I expect that will also require stcal resample changes)

mcara:
> Thanks for putting this together.
> 
> xref: #341
> 
> I'd like to wait to make changes to resample here until:
> 
> * the initial pass at using this code in romancal is merged: [RCAL-1014: Update resample to use stcal romancal#1634](https://github.com/spacetelescope/romancal/pull/1634)
> * the use of `var_sky` in romancal is sorted: [Hackday mteodoro sky variance readnoise romancal#1556](https://github.com/spacetelescope/romancal/pull/1556) (since I expect that will also require stcal resample changes)

I disagree:

1. https://github.com/spacetelescope/romancal/pull/1634 should be updated to use the code from this PR. It will make code simpler and cleaner and more efficient.
2. It's unlikely that `var_sky` would require changes to `update_time`;
3. The change in this PR should have been in my original PR but I missed it. It is the right thing to do regardless of whether you want to use it in your PR or not.

braingram:
>     1. [RCAL-1014: Update resample to use stcal romancal#1634](https://github.com/spacetelescope/romancal/pull/1634) should be updated to use the code from this PR. It will make code simpler and cleaner and more efficient.

This PR won't work for https://github.com/spacetelescope/romancal/pull/1634. The exposure time resampling for romancal occurs for each model whereas `update_time` is called only once for each group.

>     2. It's unlikely that `var_sky` would require changes to `update_time`;

I agree that `update_time` on stcal main likely won't require updates to handle `var_sky`. I was referring to the other parts of the resample code in stcal which I expect will require changes. I don't have a fully formed idea yet and I think this can all be handled in follow up PRs. Since this seems to be a sticking point I'll try to explain what I have in mind but:

1. It's not a fully formed idea
2. It's not required for https://github.com/spacetelescope/romancal/pull/1634

Since handling `var_sky` will require:
- resampling `var_sky` arrays computed from input models
- creating a new `var_sky` array on the output model
- other changes to weight calculations that I'll ignore here

One structure that might work would be to break the individual resampling operations (input arrays to output array) into a separate class (like is done in https://github.com/spacetelescope/romancal/pull/1634 for exposure time). This would allow `Resample` to act more like a router that:
- takes an input model dictionary
- computes the pixmap
- passes it to all relevant `Resamplers`

This structure would:
- be easier for subclasses to extend (they can just register new `Resamplers` without having to re-implement `add_model`)
- encapsulate state of array resampling (for example `Resample` would no longer have `_var_rnoise_wsum`, `_var_flat_wsum` etc attributes and instead each `Resampler` could have a `_wsum`)

If we went with the above structure `var_sky` could be implemented in romancal by:
- adding a `VarSkyResampler`
- registering the resampler with `Resample`
- adjusting the error computation to use `var_sky`

All of that is relevant here because exposure time could be handled in a similar way.

>     3. The change in this PR should have been in my original PR but I missed it. It is the right thing to do regardless of whether you want to use it in your PR or not.

Are these changes needed by jwst? If not I'd say we close this PR.

mcara:
I think this PR (in its updated form) allows you to perform additional resamples while re-using most time-consuming computed values using minimal effort.

Even if you decide to do things differently in your `romancal` PR, this PR will allow any other users to easily customize their workflow.

I disagree with your suggestion to break-up `Resample`. This PR solves the issue in a simple way.

changes/342.apichange.rst
```diff
@@ -0,0 +1 @@
+Added ``driz_args`` argument to ``resample.Resample.update_time()`` method.
```
braingram, line 1: Would you update this to reflect the changes in this PR?

src/stcal/resample/resample.py
```diff
@@ -1157,12 +1221,62 @@ def init_time_counters(self):
 
         self._measurement_time_success = []
 
-    def update_time(self, model):
-        """
-        A method called by the :py:meth:`~Resample.add_model` method to
+    def update_time(self, model, pixmap, iscale, weight_map,
```
braingram, line 1224:
```suggestion
    def update_time(self, model, pixmap, iscale, weight_map,
```
Is the change to this function necessary anymore (since it's not used by roman or jwst)?

braingram (commented):
Would you add unit tests for the new `add_model_hook` feature and other changes in this PR?

src/stcal/resample/resample.py
mcara: I removed this. Can be restored later if needed. I think the proposed hook function can do what initial modification to `update_time` could do + more.

changes/342.apichange.rst
mcara: done

mcara:
Added unit test too

braingram (commented):
Code looks good to me. A few comments about docs and test changes. Would you run jwst and romancal regtests.

I put together a romancal branch that uses this new API and it's looking great! I'll open a PR later today.

tests/resample/test_resample.py
```diff
@@ -87,75 +124,42 @@ def test_resample_mostly_defaults(weight_type):
...
-        np.sum(odata * oweight, dtype=float),
-        influx * np.prod(shape),
-        atol=0,
-        rtol=1e-6,
-    )
-
-    assert np.nansum(resample.output_model["var_flat"]) > 0.0
-    assert np.nansum(resample.output_model["var_poisson"]) > 0.0
-    assert np.nansum(resample.output_model["var_rnoise"]) > 0.0
+    assert str(err_info.value) == "raised by subclass' add_model_hook"
```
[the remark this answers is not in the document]
mcara: Why?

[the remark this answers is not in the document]
mcara: Oh, I see

tests/resample/test_resample.py
```diff
@@ -87,75 +124,42 @@ def test_resample_mostly_defaults(weight_type):
...
 
 
-@pytest.mark.parametrize(
-    "compute_err,weight_type",
-    [
-        ("from_var", "ivm-smed"),
-        ("driz_err", "ivm-med5")
-    ]
-)
-def test_resample_compute_error_mode(compute_err, weight_type):
```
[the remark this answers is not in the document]
mcara: restored this test

src/stcal/resample/utils.py
```diff
@@ -87,16 +74,16 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
...
-        ``weight_type="ivm"`` (the default), the weighting will be
+        ``weight_type="ivm"``, the weighting will be
         determined per-pixel using the inverse of the read noise
         (VAR_RNOISE) array stored in each input image. If the
         ``VAR_RNOISE`` array does not exist,
         the variance is set to 1 for all pixels (i.e., equal weighting).
         If ``weight_type="exptime"``, the weight will be set equal
         to the measurement time when available and to
-        the exposure time otherwise.
+        the exposure time otherwise. For the default value of `
```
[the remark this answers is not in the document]
mcara: done

src/stcal/resample/utils.py
```diff
@@ -131,21 +122,28 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
...
+            result = inv_variance * dqmask
         else:
             warnings.warn(
                 "'var_rnoise' array not available. "
                 "Setting drizzle weight map to 1",
                 RuntimeWarning
             )
-            inv_variance = 1.0
-        result = inv_variance * dqmask
+            inv_variance = 1
```
braingram, line 132:
```suggestion
```

braingram (commented):
Thanks for making the updates. One minor comment about what looks like an unused variable. Would you re-run jwst and romancal regtests with this PR?

mcara:
JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/14053707823
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/14053737624

braingram (approved):
LGTM.

jwst failures look to be unrelated and match those on main:
https://github.com/spacetelescope/RegressionTests/actions/runs/14048447450

changes/342.apichange.rst
```diff
@@ -0,0 +1,9 @@
+Added ``add_model_hook()`` method to ``resample.Resample`` that is called by
+``add_model()`` and allows subclasses to perform additional processing while
+reusing quantities computed by ``add_model()``.
```
emolter, line 3:
```suggestion
Added ``add_model_hook()`` method to ``resample.Resample`` that
allows subclasses to perform additional processing while
reusing quantities computed by ``add_model()``.
```

changes/342.apichange.rst
```diff
@@ -0,0 +1,9 @@
...
+
+
+Modified the behavior of ``resample.utils.build_driz_weight()``: any
+value of ``weight_type`` argument other than "ivm", "exptime", or ``None``
```
emolter, line 7:
```suggestion
value of the ``weight_type`` argument other than "ivm", "exptime", or ``None``
```

changes/342.apichange.rst
emolter: suggest to remove this implementation detail

src/stcal/resample/resample.py
```diff
@@ -799,7 +801,68 @@ def add_model(self, model):
 
         if self._compute_err == "driz_err":
             # use resampled error
-            self.output_model["err"] = self._driz_error.out_img
+            self._output_model["err"] = self._driz_error.out_img
+
+    def add_model_hook(self, model, pixmap, iscale, weight_map,
+                       xmin, xmax, ymin, ymax):
+        """ A hook method called by the :py:meth:`~Resample.add_model` method
```
emolter, line 808: I would suggest a one-liner description followed by an extended description

emolter (commented):
I was just reading this to understand it and noticed a few small doc-related things to consider changing

ddavis-stsci (approved):

tapastro (approved):

braingram:
@mcara is this waiting on something or can it be updated and merged?

mcara:
@braingram  merge at will :)

PR #349, opened 2025-03-07
JP-3737: Handle micrometeorite flashes (v2)

drlaw1558:
This PR addresses JP-3737 by adding an additional check in the jump step to look for micrometeorite flashes, which manifest as bursts of jumps across a large fraction of the detector.  This replaces prior PR https://github.com/spacetelescope/stcal/pull/308, updating it to work with the refactored jump step.  Corresponding jwst PR is https://github.com/spacetelescope/jwst/pull/9253

Kept in draft form for now pending potential future modifications.


braingram:
Closing based on discussion on JP ticket. Feel free to reopen and update this PR if the approach is still a candidate.

PR #352, opened 2025-03-11
Do not rescale resampled image intensities

mcara:
Resolves [JP-3907](https://jira.stsci.edu/browse/JP-3907)

Closes https://github.com/spacetelescope/jwst/issues/9246

This PR basically undoes changes introduced in https://github.com/spacetelescope/jwst/pull/7894. For more details, see https://github.com/spacetelescope/jwst/issues/9246

It also adds a unit tests that checks that surface brightness photometry shows an agreement between distorted (input) images and resampled (distortion-free) images regardless of the location of the "source" in the distorted image.

I expect numerous test failures in JWST pipeline that will need to be OKified.

`jwst` regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/13797969918
`romancal` regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/13797977338

## Tasks

</details>

### DIFF
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -515,74 +515,6 @@ class Resample:
         """
         return self._compute_err
 
-    def _get_intensity_scale(self, model):
-        """
-        Compute an intensity scale from the input and output pixel area.
-
-        For imaging data, the scaling is used to account for differences
-        between the nominal pixel area and the average pixel area for
-        the input data.
-
-        For spectral data, the scaling is used to account for flux
-        conservation with non-unity pixel scale ratios, when the
-        data units are flux density.
-
-        Parameters
-        ----------
-        model : dict
-            The input data model.
-
-        Returns
-        -------
-        iscale : float
-            The scale to apply to the input data before drizzling.
-
-        """
-        photom_pixel_area = model["pixelarea_steradians"]
-        wcs = model["wcs"]
-
-        if photom_pixel_area:
-            if 'SPECTRAL' in wcs.output_frame.axes_type:
-                # Use the nominal area as is
-                input_pixel_area = photom_pixel_area
-
-                # If input image is in flux density units, correct the
-                # flux for the user-specified change to the spatial dimension
-                if is_flux_density(model["bunit_data"]):
-                    iscale = 1.0 / math.sqrt(self.pixel_scale_ratio)
-                else:
-                    iscale = 1.0
-            else:
-                input_pixel_area = self.get_input_model_pixel_area(model)
-
-                if input_pixel_area is None:
-                    raise ValueError(
-                        "Unable to compute input pixel area from WCS of input "
-                        f"image {repr(_get_model_name(model))}."
-                    )
-
-                if self._pixel_scale_ratio is None:
-                    input_pscale = 3600.0 * np.rad2deg(
-                        math.sqrt(input_pixel_area)
-                    )
-
-                    self._pixel_scale_ratio = (
-                        self._output_pixel_scale / input_pscale
-                    )
-
-                    # update output model if "pixel_scale_ratio" was never
-                    # set previously:
-                    if (self._output_model is not None and
-                            self._output_model.get("pixel_scale_ratio") is None):
-                        self._output_model["pixel_scale_ratio"] = self._pixel_scale_ratio
-
-                iscale = math.sqrt(photom_pixel_area / input_pixel_area)
-
-        else:
-            iscale = 1.0
-
-        return iscale
-
     def reset_arrays(self, n_input_models=None):
         """ Initialize/reset `Drizzle` objects, output model and arrays,
         and time counters and clears the "finalized" flag. Output WCS and shape
@@ -742,7 +674,12 @@ class Resample:
                 f"Input model '{_get_model_name(model)}' is not a 2D image."
             )
 
-        iscale = self._get_intensity_scale(model)
+        if (group_id := model["group_id"]) not in self._group_ids:
+            self.update_time(model)
+            self._group_ids.append(group_id)
+            self.output_model["pointings"] += 1
+
+        iscale = 1.0
         log.debug(f'Using intensity scale iscale={iscale}')
 
         pixmap = calc_pixmap(
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -524,14 +524,17 @@ class Resample:
         For spectral data, the scaling is used to account for flux
         conservation with non-unity pixel scale ratios, when the
         data units are flux density.
+
         Parameters
         ----------
         model : dict
             The input data model.
+
         Returns
         -------
         iscale : float
             The scale to apply to the input data before drizzling.
+
         """
         photom_pixel_area = model["pixelarea_steradians"]
         wcs = model["wcs"]
@@ -549,7 +552,8 @@ class Resample:
         return iscale
 
     def _compute_pixel_scale_ratio(self, model):
-        if self._pixel_scale_ratio is None:
+        if (self._pixel_scale_ratio is None and
+                'SPECTRAL' not in model["wcs"].output_frame.axes_type):
             photom_pixel_area = model["pixelarea_steradians"]
             input_pscale = 3600.0 * np.rad2deg(
                 math.sqrt(photom_pixel_area)
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -548,6 +548,23 @@ class Resample:
 
         return iscale
 
+    def _compute_pixel_scale_ratio(self, model):
+        if self._pixel_scale_ratio is None:
+            photom_pixel_area = model["pixelarea_steradians"]
+            input_pscale = 3600.0 * np.rad2deg(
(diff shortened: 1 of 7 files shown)


melanieclarke (changes requested):
I think this needs discussion with INS before proceeding.

Also, there are two separate issues handled in the `_get_intensity_scale` function.  The imaging iscale is the part you are trying to undo, but the spectral iscale is necessary for flux conservation when the units are flux density.  At a minimum, that change needs to be separated out and preserved.

drlaw1558:
Just now catching up with this; yes, I'd like to discuss this further and whether it relates to the other currently open discussions of outstanding drizzle issues.

tests/resample/conftest.py
```diff
@@ -11,3 +21,71 @@ def wcs_gwcs():
     shape = (1000, 1000)
     pscale = 0.06 / 3600
     return make_gwcs(crpix, crval, pscale, shape)
+
+
+@pytest.fixture(scope="module")
+def nrcb5_wcs_wcsinfo():
+    """ Returns both wcs and wcsinfo """
+    path = Path(__file__).parent.parent / DATADIR / "nrcb5-wcs.asdf"
```
pllim, line 29: Does this package not use `astropy.utils.data.get_pkg_data_filename()`?

tests/resample/conftest.py
```diff
@@ -11,3 +21,71 @@ def wcs_gwcs():
...
+
+    patch_size = 21
+    patch_area = patch_size * patch_size
+    p2 = patch_size // 2
+    # add border so that resampled partial pixels can be isolated
+    # in the segmentation:
+    border = 4
+    pwb = patch_size + border
+
+    fwhm2sigma = 2.0 * math.sqrt(2.0 * math.log(2.0))
```
pllim, line 53: Just use this since `astropy` is already dependency?

https://github.com/astropy/astropy/blob/ea320a519eae450ec542f8009ac8d4a5a728a24f/astropy/stats/funcs.py#L58

tests/resample/conftest.py
```diff
@@ -11,3 +21,71 @@ def wcs_gwcs():
...
+    path = Path(__file__).parent.parent / DATADIR / "nrcb5-wcs.asdf"
+    with asdf.open(path, lazy_load=False) as asdf_file:
+        wcs = asdf_file.tree["wcs"]
+        wcs.bounding_box = wcs_bbox_from_shape(wcs.array_shape)
+        wcsinfo = asdf_file.tree["wcsinfo"]
+        return wcs, wcsinfo
+
+
+@pytest.fixture(scope="module")
+def nrcb5_many_fluxes(nrcb5_wcs_wcsinfo):
```
pllim, line 38: Actually, if you don't mind adding photutils as a *test dependency*, you can call this instead:

https://photutils.readthedocs.io/en/latest/user_guide/datasets.html#simulating-images

drlaw1558:
I think we also need to put back the computation of self.pixel_scale_ratio for the imaging case as well.  Even though this PR is removing iscale for this case, I believe this segment of code is still where the basic pixel_scale_ratio was getting calculated in some cases, and not having this information is what's causing failures in the new https://github.com/spacetelescope/stcal/pull/370 unit test even though it's applying ok to my test files.

tests/resample/conftest.py
mcara: @pllim How do you know all of this?

tests/resample/conftest.py
mcara: Not worth it (for me):
1. `Gaussian2DKernel` is simple;
2. I want two types of sources: uniform ("extended") and point sources spread across the image;
3. I want a grid so that I can check whether the position in input image affects the resampled flux (distortion effects);
4. I want pixels outside of the patch to be zero so they do not contribute (mix with) border pixels in the patch in order to test _as precise as possible_ flux conservation.

I don't know whether that function can do this simulation

melanieclarke:
> Could you please point me to where this is used and it would crash if not present? Thanks!

It's related to the changes for #370.  See here:
https://github.com/spacetelescope/stcal/blob/7a9c315735ed5fd38a945b7a8a7c845d3e74a1a1/src/stcal/resample/resample.py#L915
and here:
https://github.com/spacetelescope/stcal/blob/7a9c315735ed5fd38a945b7a8a7c845d3e74a1a1/src/stcal/resample/resample.py#L1100

The CI tests are failing on this branch because of this.

mcara:
@melanieclarke ready to be merged. downstream JWST tests will fail until https://github.com/spacetelescope/jwst/pull/9278 is merged. Romancal should not fail but current failures (skycell) are unrelated.

melanieclarke (commented):
A couple more questions and suggestions.  We will also need to rerun regression tests for romancal and jwst.

tests/resample/conftest.py
mcara: Maybe deprecate it in favor of all caps?

mcara: I'll open an issue

src/stcal/resample/resample.py
```diff
@@ -518,71 +518,53 @@ def compute_err(self):
...
             iscale = 1.0
 
         return iscale
 
+    def _compute_pixel_scale_ratio(self, model):
+        if self._pixel_scale_ratio is None:
+            photom_pixel_area = model["pixelarea_steradians"]
+            input_pscale = 3600.0 * np.rad2deg(
+                math.sqrt(photom_pixel_area)
+            )
```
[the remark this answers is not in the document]
drlaw1558: I think my preference would be to leave this as-is at the moment.  It should work to do both, but given the complexity of this part of the code and the reference files that it interacts with I'd be wary of introducing subtle new problems with such a change.  Let's keep it as-is and revisit it later if it proves to be necessary.

src/stcal/resample/resample.py
```diff
@@ -518,71 +518,53 @@ def compute_err(self):
...
+            # If input image is in flux density units, correct the
+            # flux for the user-specified change to the spatial dimension
+            iscale = 1.0 / math.sqrt(self.pixel_scale_ratio)
 
         else:
             iscale = 1.0
 
         return iscale
 
+    def _compute_pixel_scale_ratio(self, model):
```
melanieclarke, line 551: I just realized that this is called for both imaging and spectral data, but the implementation is only appropriate to imaging data.  Spectral data for jwst should always have _pixel_scale_ratio set by the calling code, but it would be clearer to make sure this code is only called for imaging data, or else add an implementation appropriate to spectral data.

melanieclarke: @mcara - deadline for this build is this Friday and it would be nice to get this one in and resolved.  Do you have time to get this cleaned up?

mcara: Yes, later today. You are right about imaging vs spectral issue here. Thanks for catching this. I'll think about a fix.

melanieclarke: Great, thank you!

melanieclarke: Right, I don't think it's necessary right now to add an implementation for spectral data because it should never be called with the current code structure, but it would be clearer and more future-proof if the code did not allow this block to be called for spectral data ever, because it's only appropriate to imaging.

melanieclarke (commented):
I think all comments have been addressed now.  I'll run a fresh set of regression tests.

mcara:
@melanieclarke I believe this PR is ready now. I fixed most failing tests in `romancal` but now there is one failing test that due to the fact that now, after #370 , error and variance arrays are rescaled by pixel scale ratio and this one has changed from being computed using mean pixel area of the input image to pixel area reported in the header. I checked and area has changed from 2.8069988166685943e-13 to 2.8083389953727505e-13 srad. So this test will now have to be OKified as well as JWST tests.

melanieclarke:
Thanks Mihai!  I'm rerunning the jwst regtests now because there were some unrelated failures in NIRSpec MOS modes -- I just want to make sure spectral data is definitely not affected by the changes here.

It also looks like there's a failure in an outlier_detection unit test for jwst, in addition to the ones I expected in resampling:
test_outlier_step_square_source_no_outliers: jwst.outlier_detection.tests.test_outlier_detection
Can you take a look at this one?

melanieclarke:
Thanks @schlafly - can you approve the PR for Roman, then?

melanieclarke (commented):
LGTM for JWST, but I'd like to hold off on merge until the last failing unit test on the jwst side is fixed in https://github.com/spacetelescope/jwst/pull/9278

schlafly (approved):

mcara:
> LGTM for JWST, but I'd like to hold off on merge until the last failing unit test on the jwst side is fixed in [spacetelescope/jwst#9278](https://github.com/spacetelescope/jwst/pull/9278)

I investigated this and discussed the findings with @melanieclarke . It seems the issue is in the unit test itself: data are all random and the pixel area is set to 1 srad.

melanieclarke (approved):
Unit test failure is understood - we'll mark it xfail in the jwst PR and fix the test data later.

I'll go ahead and get this merged now and will follow up with the jwst PR after.  

Thanks for your patience and for working this through, Mihai!

PR #356, opened 2025-03-18
JP-3916:  Accounting for bias in group 2 saturation flagging

hayescr:
Resolves [JP-3916](https://jira.stsci.edu/browse/JP-3916)

This PR addresses the special group 2 check for saturation in frame-averaged group data, where the counts from saturating CR in a later frame may be averaged down below the saturation threshold.  The PR updates the checks for such cases to account for non-zero bias levels.

This PR will be tied to JWST PR https://github.com/spacetelescope/jwst/pull/9302.

## Tasks

</details>


hayescr:
Moving out of draft and tagging @drlaw1558 in case you would like to review as well.

melanieclarke:
JWST regression tests with jwst PR + this branch here:
https://github.com/spacetelescope/RegressionTests/actions/runs/14473554984

Romancal tests with this branch here:
https://github.com/spacetelescope/RegressionTests/actions/runs/14473603523

hayescr:
Thanks, I'll take a look at the results to check whether they are expected changes.

hayescr:
Looking over the regtests, I'm only seeing changes that I would expect, similar to what was mentioned on the ticket.  Here are the main cases that showed changes more significant than header keyword changes, with a brief summary of the changes:

test_nirspec_irs2_detector1 - primarily only improvements (additional CRs/snowballs flagged, partially flagged snowballs now fully flagged, etc.).  As mentioned on the ticket, due to the interaction between the saturation flagging and snowball detection, in the rate image there are a couple of snowballs where the flagging was slightly different so the rates in the pixels around the snowball are slightly different at a low level (though not particularly at a concerning level of change)

test_nirspec_irs2_detector1_with_clean_flicker_noise - same dataset as above, so many of the changes are the same, but there also appears to be a slight change in the cleaning from clean_flicker_noise (typically within the noise).  It appears that the changes to saturation flagging have resulted in slightly different masks and backgrounds for the 1/f fitting, so I'm not surprised that that could introduce lots of small changes.

test_nirspec_verify - a few extra CRs caught, a couple small changes in snowball shapes (both real snowballs and ones that are masking saturating sources), recover some pixels that have a saturating CR in group 3 and were being flagged with group 2 saturation due even though no CR is visible in the second group.

test_nircam_image_sirs - Similar to the nirspec verify test are also a few cases of a saturating CR in group 3 that was resulting in group 2 sat flagging even though the CR isn't present in group 2 (and the branch changes are no longer mistakenly flagging group 2).  There are also a few (non-saturating) pixels around the core of a group 2 CR are no longer being flagged.  I checked the resulting rate files for this example and some of these pixels affected by the CR are masked by a jump detection snowball anyway, though for 1-2 CRs, there are a few CR affected pixels that now make it to the rate files.  I have a bit more of a discussion of cases like this on the ticket.

Let me know if it would help to see any examples.

melanieclarke:
Thanks for reviewing the diffs @hayescr - those sound reasonable to me.

I'm running regtests for jwst on main to confirm that this is not a breaking change for the current repository state:
https://github.com/spacetelescope/RegressionTests/actions/runs/14517788655

Assuming those come back clean, I think this looks good for JWST.  We will need romancal approval as well.

melanieclarke (approved):
Regtests on JWST main show the DQ diffs for the NIRCam image dataset, which I think is expected from the changes here.  The  NIRSpec IRS2 diffs will show up when the JWST PR is merged.  I think this looks fine to merge first, and follow up with the JWST PR when we're ready.

ddavis-stsci (approved):
LGTM the romancal tests pass
https://github.com/spacetelescope/RegressionTests/actions/runs/14518267943

PR #359, opened 2025-04-08
RCAL-968: add inverse of sky variance as an option for creating weight map in resample.

mairanteodoro:
Resolves [RCAL-968](https://jira.stsci.edu/browse/RCAL-968)

This PR adds another option for the creation of a weight map by using the inverse of the sky variance in resample.

## Tasks

</details>

### DIFF
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -223,7 +223,7 @@ class Resample:
         self.fillval = fillval
         self.good_bits = good_bits
 
-        if weight_type.startswith("ivm") or weight_type == "exptime":
+        if weight_type.startswith("ivm") or weight_type in ("exptime", "ivsky"):
             self.weight_type = weight_type
         else:
             raise ValueError("Unexpected weight type: '{self.weight_type}'")
@@ -964,9 +964,10 @@ class Resample:
             else:
                 mask = np.full_like(rn_var, False)
 
+            weight = np.ones(self.output_array_shape)
+
             # Set the weight for the image from the weight type
             if self.weight_type.startswith("ivm") and rn_var is not None:
-                weight = np.ones(self.output_array_shape)
                 weight[mask] = np.reciprocal(rn_var[mask])
 
             elif self.weight_type == "exptime":
--- a/src/stcal/resample/utils.py
+++ b/src/stcal/resample/utils.py
@@ -143,6 +143,24 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
     elif weight_type == 'exptime':
         exptime, s = get_tmeasure(model)
         result = exptime * dqmask
+        
+    elif weight_type == "ivsky":
+        var_sky = model["var_sky"]
+        if (
+            var_sky is not None and 
+            var_sky.shape == data.shape
+        ):
+            with np.errstate(divide="ignore", invalid="ignore"):
+                inv_sky_variance = var_sky**-1
+            inv_sky_variance[~np.isfinite(inv_sky_variance)] = 0
+        else:
+            warnings.warn(
+                "'var_sky' array is not available. "
+                "Setting drizzle weight map to 1",
+                RuntimeError,
+            )
+            inv_sky_variance = 1.0
+        result = inv_sky_variance * dqmask
 
     else:
         result = np.ones(data.shape, dtype=data.dtype) * dqmask
--- a/tests/resample/test_resample_utils.py
+++ b/tests/resample/test_resample_utils.py
@@ -120,7 +120,7 @@ def test_is_flux_density(unit, result):
     assert is_flux_density(unit) is result
 
 
-@pytest.mark.parametrize("weight_type", ["ivm", "exptime"])
+@pytest.mark.parametrize("weight_type", ["ivm", "exptime", "ivsky", ])
 def test_build_driz_weight(weight_type):
     """Check that correct weight map is returned of different weight types"""
 
@@ -129,6 +129,7 @@ def test_build_driz_weight(weight_type):
     model["dq"][0] = JWST_DQ_FLAG_DEF.DO_NOT_USE
     model["measurement_time"] = 10.0
     model["var_rnoise"] /= 10.0
+    model["var_sky"] /= 10.0
 
     weight_map = build_driz_weight(
         model,
--- a/tests/resample/helpers.py
+++ b/tests/resample/helpers.py
@@ -107,7 +107,7 @@ def make_input_model(shape, crpix=(0, 0), crval=(0, 0), pscale=2.0e-5,
         "subtracted": False,
     }
 
-    for arr in ["var_flat", "var_rnoise", "var_poisson"]:
+    for arr in ["var_flat", "var_rnoise", "var_poisson", "var_sky"]:
         model[arr] = np.ones(shape, dtype=np.float32)
 
     model["err"] = np.sqrt(3.0) * np.ones(shape, dtype=np.float32)


mcara:
I this weighting method is intended to be used only by Roman and not JWST, I think this should go into `romancal`. However,  currently there is no easy way to overwrite `build_driz_weight` using OOP.

I will make a PR to allow subclasses to customize weight computations.

tapastro:
> I this weighting method is intended to be used only by Roman and not JWST, I think this should go into `romancal`. However, currently there is no easy way to overwrite `build_driz_weight` using OOP.
> 
> I will make a PR to allow subclasses to customize weight computations.

It's true that JWST has not requested this weighting, but it could be useful to have it available for testing. The scientific motivation for Roman should be similarly applicable to JWST (though we don't have a method of providing sky variance yet). I'm not opposed to exposing it to JWST.

src/stcal/resample/utils.py
```diff
@@ -143,6 +143,24 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
     elif weight_type == 'exptime':
         exptime, s = get_tmeasure(model)
         result = exptime * dqmask
+        
+    elif weight_type == "ivsky":
+        var_sky = model["var_sky"]
+        if (
+            var_sky is not None and
```
kmacdonald-stsci, line 150: I recommend putting the `and` at the begging of the next line.

src/stcal/resample/utils.py
```diff
@@ -143,6 +143,24 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
...
         result = exptime * dqmask
+        
+    elif weight_type == "ivsky":
+        var_sky = model["var_sky"]
+        if (
+            var_sky is not None and 
+            var_sky.shape == data.shape
+        ):
+            with np.errstate(divide="ignore", invalid="ignore"):
+                inv_sky_variance = var_sky**-1
```
kmacdonald-stsci, line 154: I recommend just doing `1. / var_sky`, rather than `var_sky**-1`.

schlafly (commented):
This looks good to me.  I think we should combine the ivm / ivsky -> weight calculations, which are largely duplicated, but otherwise looks good.

src/stcal/resample/utils.py
```diff
@@ -143,6 +143,24 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
...
+            inv_sky_variance[~np.isfinite(inv_sky_variance)] = 0
+        else:
+            warnings.warn(
+                "'var_sky' array is not available. "
+                "Setting drizzle weight map to 1",
+                RuntimeError,
+            )
```
braingram, line 161:
```suggestion
            warnings.warn(
                "'var_sky' array is not available. "
                "Setting drizzle weight map to 1",
                RuntimeError,
            )
```
I think it would be more consistent with the current code if this was not a warning. Instead `validate_input_model` could require `var_sky` if `weight_type==ivsky` as is the case for `ivm` and `var_rnoise`:
https://github.com/spacetelescope/stcal/blob/e34115bf6d06bf5f636dd5ebe8a2b12974f0b023/src/stcal/resample/resample.py#L687-L689

mcara:
I would suggest renaming `ivsky` to something like `ivm-sky`. The reason is that this is a different class of weight compared of `exptime` and in several places in the code there is a check like `if weight_type.startswith('ivm')`. Of course we can adjust those checks but it is nicer overall to be able to identify the general type of weight.

src/stcal/resample/resample.py
```diff
@@ -111,12 +111,16 @@ def __init__(self, output_wcs, n_input_models=None, pixfrac=1.0,
...
             ``fillval`` value.
 
-        weight_type : {"ivm", "exptime"}, optional
+        weight_type : {"ivm", "exptime", "ivsky"}, optional
             The weighting type for adding models' data. For
             ``weight_type="ivm"`` (the default), the weighting will be
             determined per-pixel using the inverse of the read noise
             (VAR_RNOISE) array stored in each input image. If the
             ``VAR_RNOISE`` array does not exist, the variance is set to 1 for
+            all pixels (i.e., equal weighting). If ``weight_type="ivm"``,
```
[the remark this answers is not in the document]
mairanteodoro: Following @mcara's suggestion, I changed it to `ivm-sky`, so that line now should read `weight_type="ivm-sky"`.

src/stcal/resample/utils.py
mairanteodoro: Done.

mairanteodoro:
> This looks good to me. I think we should combine the ivm / ivsky -> weight calculations, which are largely duplicated, but otherwise looks good.

Done.

src/stcal/resample/utils.py
mairanteodoro: Done.

schlafly (commented):
Thanks Mairan.  One comment inline.

Do you think anyone will use the different warning types?  I think I would tend to remove that argument and just always use RuntimeWarning or whatever.  I also don't think there's a reason for ivm and ivm-sky to use different fill values---I suspect that's a bug for the ivm weighting that we should address, but I've asked Tyler to weigh in there.  I think I'd get rid of the array name from the warning message and just say 'requested array not available' to save an additional argument, so that _get_inverse_variance would end up just having var, shape arguments, which would be simpler.

src/stcal/resample/utils.py
```diff
@@ -114,26 +117,29 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
         dtype=np.uint8,
         flag_name_map=flag_name_map,
     )
+    
+    if weight_type == "ivm":
+        inv_variance = _get_inverse_variance(
+            model.get("var_rnoise"),
+            data.shape, "var_rnoise",
+            1,
```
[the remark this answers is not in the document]
tapastro: I think for purposes of variance resampling, 0 seems like the correct choice. I'm not aware of missing (0 or NaN) rnoise variances being a common situation, so the change in behavior doesn't worry me too much as causing a large diff in outputs, but I'll tag @drlaw1558 in case he has reservations for JWST behavior.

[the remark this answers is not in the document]
drlaw1558: On principle I'm inclined to agree that 0 makes more sense; if the RN variance is 0/NaN you don't want to be using that pixel, so an inverse variance weight of 0 would achieve that.

I remember a somewhat related issue coming up previously in https://github.com/spacetelescope/jwst/pull/8437  VAR_POISSON can be zero even for good pixels if the slope is negative (consistent within uncertainty with zero signal), but VAR_RN should be defined.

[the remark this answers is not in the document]
mairanteodoro: I refactored `_get_inverse_variance()` to always return a zeroed array when an attribute is missing.

src/stcal/resample/utils.py
```diff
@@ -423,3 +422,29 @@ def is_flux_density(bunit):
...
+    inv : numpy.ndarray
+        Inverse variance array. If input array is missing or has wrong shape,
+        returns an array filled with zeros of shape `data_shape`.
+    """
+    if array is not None and array.shape == data_shape:
+        with np.errstate(divide="ignore", invalid="ignore"):
+            inv = 1.0 / array
+        inv[~np.isfinite(inv)] = 0
+    else:
+        inv = np.full(data_shape, 0, dtype=np.float32)
```
schlafly, line 449:
```suggestion
        inv = np.full(data_shape, 1, dtype=np.float32)
```
Let's retain the old behavior for this failure case (we asked for some weighting that didn't exist and we just provide all 1).

src/stcal/resample/utils.py
```diff
@@ -74,15 +74,16 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
...
-        ``VAR_RNOISE`` array does not exist,
-        the variance is set to 1 for all pixels (i.e., equal weighting).
-        If ``weight_type="exptime"``, the weight will be set equal
-        to the measurement time when available and to
+        (VAR_RNOISE) array stored in each input image. If
+        ``weight_type="ivm-sky"``, the weighting will be determined
+        by the inverse of the sky variance ``VAR_SKY``. If the ``VAR_SKY``
+        array does not exist, the variance is set to 0 for all pixels
+        (i.e., equal weighting).  If ``weight_type="exptime"``, the 
+        weight will be set equal to the measurement time when available and to
```
schlafly, line 86: For this and the corresponding documentation in resample.py let's merge the ivm and ivm-sky documentation.
i.e.,
The weighting type for adding models' data.  For weight_type = ivm and weight_type = ivm-sky, the weighting will be determined per pixel using the inverse of either the read noise or sky noise arrays, respectively (VAR_RNOISE or VAR_SKY).  If the array does not exist, the weight is set to 1 for all pixels (i.e., equal weighting).

src/stcal/resample/utils.py
mairanteodoro: Done.

src/stcal/resample/utils.py
mairanteodoro: Done.

src/stcal/resample/utils.py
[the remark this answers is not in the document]
mairanteodoro: The latest changes do the following:
- handle "ivm" and "Ivm-sky" cases identically:
  - for "ivm": calculate 1/var_rnoise, setting non-finite to 0, and if missing or shape mismatch, return all ones (equal weighting);
  - for "ivm-sky": calculate 1/var_sky, setting non-finite to 0, and if missing or shape mismatch, return all ones (equal weighting);
- always multiply the result by the DQ mask;
- the fallback for a missing variance array is an array filled with ones (equal weighting).

src/stcal/resample/utils.py
```diff
@@ -74,20 +74,18 @@ def build_driz_weight(model, weight_type=None, good_bits=None,
...
-        the model. The default value of `None` will
-        set weights to 1 for pixels not flagged in the DQ array of the model.
-        Pixels flagged as "bad" in the DQ array will have thier weights
-        set to 0.
+    weight_type : {"exptime", "ivm", "ivm-sky"}, None, optional
+        The weighting type for adding models' data. For ``weight_type="ivm"``
+        and ``weight_type="ivm-sky"``, the weighting will be determined
+        per-pixel using the inverse of either the read noise (VAR_RNOISE) or
+        sky variance (VAR_SKY) arrays, respectively. If the array does not
+        exist, the weight is set to 1 for all pixels (i.e., zero weighting).
```
schlafly, line 82:
```suggestion
        exist, the weight is set to 1 for all pixels (i.e., equal weighting).
```

schlafly (approved):
I think this is good to go, modulo one minor docstring update I left in-line.  @tapastro , mind taking another look?

mairanteodoro:
> I think this is good to go, modulo one minor docstring update I left in-line. @tapastro , mind taking another look?

I updated the docstring.

tapastro (approved):
Looks good!

tests/resample/test_resample_utils.py
```diff
@@ -155,3 +156,22 @@ def test_unsupported_weight_type(weight_type):
...
+    assert inv[1, 1] == 1.0
+    # Wrong shape
+    with pytest.warns(RuntimeWarning, match=f"'{array_name}' array not available."):
+        inv2 = _get_inverse_variance(None, (2, 2), array_name)
+    assert np.all(inv2 == 1)
+    with pytest.warns(RuntimeWarning, match=f"'{array_name}' array not available."):
+        inv3 = _get_inverse_variance(np.ones((1, 1)), (2, 2), array_name)
+    assert np.all(inv3 == 1)
+
+
```
pllim, line 177:
```suggestion
```

tests/resample/test_resample_utils.py
```diff
@@ -155,3 +156,22 @@ def test_unsupported_weight_type(weight_type):
...
+    assert inv[0, 1] == 0
+    assert inv[1, 0] == 0
+    assert inv[1, 1] == 1.0
+    # Wrong shape
+    with pytest.warns(RuntimeWarning, match=f"'{array_name}' array not available."):
+        inv2 = _get_inverse_variance(None, (2, 2), array_name)
+    assert np.all(inv2 == 1)
+    with pytest.warns(RuntimeWarning, match=f"'{array_name}' array not available."):
+        inv3 = _get_inverse_variance(np.ones((1, 1)), (2, 2), array_name)
+    assert np.all(inv3 == 1)
```
pllim, line 175: I think this will give easier to read failure message? 🤷‍♀️ 
```suggestion
    np.testing.assert_array_equal(inv3, 1)
```

melanieclarke:
> The changes to general variance resampling (outside of var_sky) on handling non-finite pixels is causing downstream test failures. I think the change is good, but it should be documented - can the change be added to the changelog entry?

Just a note for documentation/follow up: this change looks like it will resolve [JP-3991](https://jira.stsci.edu/browse/JP-3991)

mairanteodoro:
The two failing tests are being caused because they are run against the `main` branch instead of the temp branch.

schlafly (approved):
Still looks good to me, one doc change in line.

PR #321, opened 2024-11-27
JP-3805: Change prior-group flagging to DO_NOT_USE instead of SATURATION

drlaw1558:
This PR addresses the issue raised by JP-3805 wherein the flagging of saturation in group-averaged data by estimating the counts in each frame can bias the finding and flagging of snowballs downstream.  This does so by changing the flagging of such 'pre-saturation' to use DO_NOT_USE instead of SATURATED.

## Tasks
- [ ] update or add relevant tests
- [ ] update relevant docstrings and / or `docs/` page
- [ ] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [ ] write news fragment(s) in `changes/`: `echo "changed something" > changes/<PR#>.<changetype>.rst` (see below for change types)
  - [ ] run regression tests with this branch installed (`"git+https://github.com/<fork>/stcal@<branch>"`)
    - [ ] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) 
    - [ ] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml)

<details><summary>news fragment change types...</summary>

- ``changes/<PR#>.apichange.rst``: change to public API
- ``changes/<PR#>.bugfix.rst``: fixes an issue
- ``changes/<PR#>.general.rst``: infrastructure or miscellaneous change
</details>

PR #362, opened 2025-04-11
JP-3164: Update STCAL CI Tests

kmacdonald-stsci:
Resolves [JP-3164](https://jira.stsci.edu/browse/JP-3164)

This PR cleans up CI test removing or updating skipped or xfail'd tests.

## Tasks

</details>

### DIFF
--- a/tests/test_saturation.py
+++ b/tests/test_saturation.py
@@ -4,7 +4,6 @@ Unit tests for saturation flagging
 
 """
 from enum import IntEnum
-import pytest
 import numpy as np
 
 from stcal.saturation.saturation import flag_saturated_pixels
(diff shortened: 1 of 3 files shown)


tests/test_saturation.py
```diff
@@ -76,7 +75,8 @@ def test_read_pattern_saturation_flagging():
     )
 
     # Make sure that groups after the third get flagged
-    assert np.all(gdq[0, 2:, 5, 5] == DQFLAGS["SATURATED"])
+    assert np.all(gdq[0, 3:, 5, 5] == DQFLAGS["SATURATED"])
```
tapastro, line 78: I think the behavior introduced by #321 is not desirable long term; if we opt to remove the xfail here, we should at least add a comment describing the ideal state for future edits to the code and this test. I believe the long-term fix will be tracked under https://jira.stsci.edu/browse/JP-3835

emolter (commented):
Overall these look good to me Ken, but what is the status of https://github.com/spacetelescope/stcal/blob/b4131aa56f556f888047d930520855abb4dcf546/tests/test_ramp_fitting_gls_fit.py#L196 ?  Are you well-versed enough in ramp fitting to take a look at that one and add it to this PR too, or should we try to get someone else to do it?

PR #364, opened 2025-04-23
Generic variances

schlafly:
This PR attempts to treat the various variance arrays in resample more symmetrically while not changing the behavior very much.  The basic idea is that all of the different variance arrays should be resampled in the same way.  It simplifies the code somewhat and allows other variance arrays to be specified for resampling, without adding more code to stcal/resample.  It does this by extending the "enable_var" argument to talk a list of variance arrays that should be resampled, but retaining the current behavior when enable_var = True.

Currently all romancal and most jwst tests pass.  The two failing jwst tests involve tests that the resample does the right thing when enable_var is True and the ivm weight type is selected, but the images do not include the var_rnoise weight type needed for ivm.  I can update this PR to handle that case if desired; it's not obvious to me that erroring out is not appropriate behavior when people ask for ivm weights but var_rnoise is not available!

Is this something that we would like to have in stcal?  I'll add a changelog entry if so.

Regression tests are here:
Webb: https://github.com/spacetelescope/RegressionTests/actions/runs/15025203836/job/42224326282
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/14623806303/job/42214787253

### DIFF
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -211,7 +211,7 @@ class Resample:
 
         include_var_in_err : list or None, optional
             Indicates which variances to include in the ``err`` computation.
-            This field is only used if ``compute_err`` is ``from_var``.  If
+            This field is used only if ``compute_err`` is ``from_var``.  If
             None, all variances are included in the error computation.
 
         """
@@ -930,21 +930,19 @@ class Resample:
             self.output_model["err"] = self._driz_error.out_img
             del self._driz_error
 
-        elif self._enable_var:
+        elif self._enable_var and (self._compute_err == "from_var"):
             # compute error from variance arrays:
             var_components = [
                 self._output_model[x] for x in self._include_var_in_err]
-            if self._compute_err == "from_var":
-                self.output_model["err"] = np.sqrt(
-                    np.nansum(var_components, axis=0)
-                )
+            self.output_model["err"] = np.sqrt(
+                np.nansum(var_components, axis=0)
+            )
 
-                # nansum returns zero for input that is all NaN -
-                # set those values to NaN instead
-                all_nan = np.all(np.isnan(var_components), axis=0)
-                self._output_model["err"][all_nan] = np.nan
-                del all_nan
-
+            # nansum returns zero for input that is all NaN -
+            # set those values to NaN instead
+            all_nan = np.all(np.isnan(var_components), axis=0)
+            self._output_model["err"][all_nan] = np.nan
+            del all_nan
             del var_components
 
         self.finalize_time_info()
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -4,6 +4,7 @@ import warnings
 import sys
 
 import numpy as np
+from collections import defaultdict
 
 from drizzle.utils import calc_pixmap
 from drizzle.resample import Drizzle
@@ -53,22 +54,25 @@ class Resample:
 
     """
     # supported output arrays (subclasses can add more):
-    output_array_types = {
+    output_array_types_explicit = {
         "data": np.float32,
         "wht": np.float32,
         "con": np.int32,
         "var_rnoise": np.float32,
         "var_flat": np.float32,
         "var_poisson": np.float32,
-        "var_sky": np.float32,
         "err": np.float32,
     }
+    # add default of float32 to support generic variances conveniently
+    output_array_types = defaultdict(lambda: np.float32)
+    output_array_types.update(output_array_types_explicit)
 
     dq_flag_name_map = None
 
     def __init__(self, output_wcs, n_input_models=None, pixfrac=1.0,
                  kernel="square", fillval=0.0, weight_type="ivm", good_bits=0,
-                 enable_ctx=True, enable_var=True, compute_err=None):
+                 enable_ctx=True, enable_var=True, compute_err=None,
+                 include_var_in_err=None):
         """
         Parameters
         ----------
@@ -205,6 +209,11 @@ class Resample:
                 At this time, output error array is not equivalent to
                 error propagation results.
 
+        include_var_in_err : list or None, optional
+            Indicates which variances to include in the ``err`` computation.
+            This field is only used if ``compute_err`` is ``from_var``.  If
+            None, all variances are included in the error computation.
+
         """
         # to see if setting up arrays and drizzle is needed
         self._finalized = False
@@ -220,6 +229,11 @@ class Resample:
                 self._enable_var = []
         else:
             self._enable_var = enable_var
+
+        if include_var_in_err is None:
+            self._include_var_in_err = self._enable_var
+        else:
+            self._include_var_in_err = include_var_in_err
 
 
         # these attributes are used only for informational purposes
@@ -505,7 +519,7 @@ class Resample:
 
     @property
     def enable_var(self):
-        """ Indicates whether variance arrays are resampled. """
+        """ List of variance arrays to resample. """
         return self._enable_var
 
     @property
@@ -917,7 +931,8 @@ class Resample:
 
         elif self._enable_var:
             # compute error from variance arrays:
-            var_components = [self._output_model[x] for x in self._enable_var]
+            var_components = [
+                self._output_model[x] for x in self._include_var_in_err]
             if self._compute_err == "from_var":
                 self.output_model["err"] = np.sqrt(
                     np.nansum(var_components, axis=0)
@@ -939,12 +954,14 @@ class Resample:
         weights. """
         shape = self.output_array_shape
 
-        self.variance_info = dict()
+        self._variance_info = dict()
         for noise_type in self._enable_var:
+            # note: output_array_types is a defaultdict, so this will succeed
+            # even when noise_type is not in output_array_types
             var_dtype = self.output_array_types[noise_type]
             wsum = np.full(shape, np.nan, dtype=var_dtype)
             wt = np.zeros(shape, dtype=var_dtype)
-            self.variance_info[noise_type] = dict(
+            self._variance_info[noise_type] = dict(
                 wsum=np.full(shape, np.nan, dtype=var_dtype),
                 wt=np.zeros(shape, dtype=var_dtype),
             )
@@ -1024,14 +1041,16 @@ class Resample:
                     model=model,
                     **pars,
                 )
-                self.variance_info[varname]['var'] = var
+                self._variance_info[varname]['var'] = var
 
         # Set the weight for the image from the weight type
         if self.weight_type.startswith("ivm"):
-            rn_var = self.variance_info['var_rnoise']['var']
-            mask = (rn_var > 0) & np.isfinite(rn_var)
-            weight = np.zeros(self.output_array_shape, dtype='f4')
-            weight[mask] = np.reciprocal(rn_var[mask])
+            rn_var_info = self._variance_info.get('var_rnoise', dict())
+            rn_var = rn_var_info.get('var', None)
+            if rn_var is not None:
+                mask = (rn_var > 0) & np.isfinite(rn_var)
+                weight = np.zeros(self.output_array_shape)
+                weight[mask] = np.reciprocal(rn_var[mask])
 
         elif self.weight_type == "exptime":
             t, _ = get_tmeasure(model)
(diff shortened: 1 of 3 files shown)


braingram:
I think treating `enable_var` as a list of enabled variances is a simple and effective way to handle this and agree that it simplifies the code (and will allow roman to more easily drop `var_flat`).

Perhaps @tapastro or @melanieclarke can weight in on the jwst unit test failures.

Looking back at the test it was added by @mcara in https://github.com/spacetelescope/jwst/pull/6305

src/stcal/resample/resample.py
```diff
@@ -1014,80 +1017,42 @@ def resample_variance_arrays(self, model, pixmap, iscale,
...
-                "var_flat",
-                model=model,
-                **pars,
-            )
-            if flat_var is not None:
-                mask = (flat_var >= 0) & np.isfinite(flat_var) & (weight > 0)
-                self._var_flat_wsum[mask] = np.nansum(
+        # Set the weight for the image from the weight type
+        if self.weight_type.startswith("ivm"):
+            rn_var = self.variance_info['var_rnoise']['var']
```
tapastro, line 1031: The downstream tests are failing because of these lines - the failing tests have empty or incorrectly sized variance arrays, which fail L1021's `self._check_var_array`, leading to no `var` key definition for the `var_rnoise` entry. Then L1031 attempts to access the `var` key.

schlafly: Right, I understand why they're failing.  My question is more whether the test we have there is a good test in the first place---in my opinion, if someone specifies inverse variance read noise weighting and then gives input models without read noise variances, it's appropriate to fail.  I can write the code to handle this case, but I thought I'd check that this was really something we want to be able to handle.

tapastro: Apologies, I read the message that tagged me in, but neglected to read the first comment. I agree that these tests are covering a case that should probably return an error - I'm not sure it's a case that we should expect to see with any regularity, and if so, that it should only return a warning.

This will require a small jwst PR to update.

schlafly: @mcara , do you want to weigh in on whether to remove the offending jwst test or update this PR to handle the case that ivm weighting is selected but the input models don't include read noise variances?

src/stcal/resample/resample.py
```diff
@@ -1109,38 +1074,13 @@ def finalize_resample_variance(self, output_model):
...
-                self._var_poisson_weight,
-                self._var_flat_weight,
-            )
+            for varname in self._enable_var:
+                varwsum = self.variance_info[varname]['wsum']
+                weight = self.variance_info[varname]['wt']
+                output_model[varname] = (varwsum / (weight * weight)).astype(
+                    dtype=self.output_array_types[varname]
+                )
+            del self.variance_info
```
mcara, line 1083: I don't like when class attributes can disappear (especially public ones). Maybe iterate over dict items and delete those? and also make this a private attribute?

src/stcal/resample/resample.py
```diff
@@ -1109,38 +1074,13 @@ def finalize_resample_variance(self, output_model):
...
-
-            del (
-                self._var_rnoise_wsum,
-                self._var_poisson_wsum,
-                self._var_flat_wsum,
-                self._var_rnoise_weight,
-                self._var_poisson_weight,
-                self._var_flat_weight,
-            )
+            for varname in self._enable_var:
```
mcara, line 1077:
```suggestion
            for varname in self.variance_info:
```

src/stcal/resample/resample.py
mcara: With regard to this particular crash, would replacing `startswith` with strict equality solve the issue?

mcara: I think the test is doing what it is supposed to do: check that the code behaves "correctly" when inputs are incorrect. The code on `main` in `stcal` is reproducing old/original behavior in `jwst` pipeline (and I assume in `romancal` as well): https://github.com/spacetelescope/jwst/blob/6902a68f15d0a762d30915830384f9fdc02e9271/jwst/resample/resample.py#L298-L304

I do not know why the code was written to behave this way (log a message and proceed) but I assume it is because pipeline is not supposed to crash (and I had to remove multiple `raise Exception` from my `tweakreg` and other code to make it not fail). Maybe this has changed but it is for @tapastro to make a decision.

src/stcal/resample/resample.py
```diff
@@ -1014,80 +1017,42 @@ def resample_variance_arrays(self, model, pixmap, iscale,
...
-                **pars,
-            )
-            if flat_var is not None:
-                mask = (flat_var >= 0) & np.isfinite(flat_var) & (weight > 0)
-                self._var_flat_wsum[mask] = np.nansum(
+        # Set the weight for the image from the weight type
+        if self.weight_type.startswith("ivm"):
+            rn_var = self.variance_info['var_rnoise']['var']
+            mask = (rn_var > 0) & np.isfinite(rn_var)
+            weight = np.zeros(self.output_array_shape, dtype='f4')
```
mcara, line 1033: This was `float64` in the original code, I believe. I think for these computations we want increased precision for intermediate computations.

src/stcal/resample/resample.py
mcara: However, the code currently in `stcal` is also buggy as `weight` would not be defined if check `var_rnoise` fails but other arrays pass. I think the cleanest would be to skip resampling of variance arrays when `var_rnoise` is missing and `weight_type` is (or stars with) "ivm" (depending on whether a particular weighting needs `rnoise`).

tapastro: I fully support avoiding error returns whenever possible - our code is first and foremost meant to support automated operations which deliver science products, and graceful handling of unexpected circumstances (especially during operations on arrays that are not the flux array) is preferable. I would suggest that the code log a warning message if it is unable to proceed with variance array resampling in the manner it expects and continues.

I don't know if Roman has different expectations.

schlafly: Note that in this case the flux array is impacted---the user has asked for ivm resampling, but var_rnoise does not exist, and so the weights used for resampling end up being trivial.  But I'll tweak this to avoid the error.

src/stcal/resample/resample.py
schlafly: Thanks.  I've renamed this to _variance_info.  I didn't delete all of the dictionary entries individually, though I could---I'll admit it's unclear to me when I'm happy to let the garbage collector do its collection on its own vs. when I want to remove things explicitly.

schlafly:
I've addressed the review comments.  Significant changes:
* I now handle the weird case that was causing an error in the jwst unit tests before.  In that test ivm weighting is requested but no var_rnoise is available.  If you work hard enough you could still trigger an error (e.g., enable_var = True, var_flat in the models, but var_rnoise not in the models), but the old code didn't handle this case either and it's not tested.
* I changed the output_array_types dictionary to be a "defaultdict" with a default float32 type.  This lets one in principle define a new var_aoeu type and have resample resample it without needing to subclass, while keeping the type definitions in one place.
* I added a new include_var_in_err argument that lets one specify which variances should be summed to produce err in the case that weight_type = "from_var".  If not set, retain the default behavior of using all variances.
* I added a little test that at least makes sure that the new include_err_in_var argument is accepted, but did not include any correctness checks there.

I also introduced some new error where mypy is complaining that I didn't annotate the defaultdict right.  I don't see any other type annotations in that codebase so I'm confused as to whether I should actually include new type annotations or what; I'd appreciate advice there.

src/stcal/resample/resample.py
```diff
@@ -202,15 +209,33 @@ def __init__(self, output_wcs, n_input_models=None, pixfrac=1.0,
                 At this time, output error array is not equivalent to
                 error propagation results.
 
+        include_var_in_err : list or None, optional
+            Indicates which variances to include in the ``err`` computation.
+            This field is used only if ``compute_err`` is ``from_var``.  If
+            None, all variances are included in the error computation.
+
```
mcara, line 216: I would prefer to:
1. not make `enable_var` a list - just boolean;
2. not make `include_var_in_err` as a parameter.

Instead, I am proposing to add two class variables:
1. `variance_array_names = ["var_rnoise"]`
2. `error_from_variances = None  # default None means use all from `variance_array_names`

Subclasses (in JWST and Roman pipelines) can customize these lists.

Alternatively, even a more flexible approach (although too much flexibility may have its own drawbacks), similar fields could be added to each input data "mode" (dictionary).

Alternatively, when `enable_var` is `True`, we could resample all variance arrays whose name stars with `"var_"`.


In particular I do not like currently proposed approach due to line 227 (see below):
```python
self._enable_var = ['var_rnoise', 'var_poisson', 'var_flat']
```
It is quite difficult (not nice) to modify this in a subclass.

src/stcal/resample/resample.py
```diff
@@ -202,15 +209,33 @@ def __init__(self, output_wcs, n_input_models=None, pixfrac=1.0,
...
         self._finalized = False
         self._n_res_models = 0
 
         self._enable_ctx = enable_ctx
-        self._enable_var = enable_var
         self._compute_err = compute_err
 
+        if isinstance(enable_var, bool):
+            if enable_var:
+                self._enable_var = ['var_rnoise', 'var_poisson', 'var_flat']
```
mcara, line 227: This is hardcoded and that is why I prefer class variables.

src/stcal/resample/resample.py
```diff
@@ -202,15 +209,33 @@ def __init__(self, output_wcs, n_input_models=None, pixfrac=1.0,
...
+            else:
+                self._enable_var = []
+        else:
+            self._enable_var = enable_var
+
+        if include_var_in_err is None:
+            self._include_var_in_err = self._enable_var
+        else:
+            self._include_var_in_err = include_var_in_err
```
mcara, line 236: Too convoluted, IMO.

src/stcal/resample/resample.py
```diff
@@ -62,12 +63,16 @@ class Resample:
         "var_poisson": np.float32,
         "err": np.float32,
     }
+    # add default of float32 to support generic variances conveniently
+    output_array_types = defaultdict(lambda: np.float32)
```
braingram, line 67:
```suggestion
    output_array_types:defaultdict[str, DTypeLike] = defaultdict(lambda: np.float32)
```

These suggestions should appease mypy.

src/stcal/resample/resample.py
```diff
@@ -4,6 +4,7 @@
 import sys
 
 import numpy as np
+from collections import defaultdict
```
braingram, line 7:
```suggestion
from collections import defaultdict

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from numpy.typing import DTypeLike
```

isort/ruff will likely want to change this to move the check after the imports but I can't make a suggestion on those lines.

src/stcal/resample/resample.py
schlafly: Okay, I've now adopted your suggestion.

src/stcal/resample/resample.py
schlafly: I've now adopted your suggestion and do this via variance_array_names.

src/stcal/resample/resample.py
schlafly: Thanks.  I've now adopted your approach.

src/stcal/resample/resample.py
```diff
@@ -53,7 +54,7 @@ class Resample:
 
     """
     # supported output arrays (subclasses can add more):
-    output_array_types = {
+    output_array_types_explicit = {
```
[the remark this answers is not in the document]
schlafly: I got rid of output_array_types_explicit and put the new keywords directly into output_array_types.

mcara (approved):

schlafly:
Okay, I've updated the regression tests here:
Webb: https://github.com/spacetelescope/RegressionTests/actions/runs/15025203836/job/42224326282
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/14623806303/job/42214787253
which are all green.  Mihai has approved.  Anyone else want to look at this or shall we merge?

tapastro (approved):

ddavis-stsci (approved):
LGTM

PR #367, opened 2025-05-07
Add interpolation to gwcs pixel map

t-brandt:
Resolves [JP-3997](https://jira.stsci.edu/browse/JP-3997)

Closes #

This PR addresses the mosaic pipeline for both JWST and Roman.  It demonstrates an approach to dramatically decrease the amount of time spent in computing coordinate transformations, which currently dominates runtime, with negligible loss of accuracy assuming the distortion corrections are all smooth.  The PR itself is not intended to be merged straightaway; the intention is to change some combination of the routine adjusted here and a similar routine in drizzle
https://github.com/spacetelescope/drizzle/blob/63749f487b766e814edcafc50f5ceed1e470d71e/drizzle/utils.py#L10 ,
and to use a single routine if possible for maintainability.

## Tasks

</details>


mcara:
I may be wrong but I am very concerned about this approach for the following reason: `drizzle` computes coordinates in the resampled image by interpolating `pixmap` values. Your approach would make drizzle use interpolated values to interpolate again (interpolate interpolated values).

Although the effects of a coarser grid may not be visible in science data, they are clearly visible in the weights of resampled images (from my past experience with a similar option in the `drizzlepac` that also allows adjusting coarseness of the grid).

schlafly:
@mcara , what do you think next steps are here?  I think it will make a big difference for the mosaic step for Roman.

mcara:
@schlafly Since this PR would not affect `drizzle` package itself and you are satisfied with it, I have no problem at all.

emolter (commented):
I wonder if it might make sense to expose the interpolation factor as an optional input?  That way we're covered if we realize that, say, one of the JWST modes needs a finer interp grid.

Could even consider replacing the "fast" flag such that interp<=1 falls back to the old "slow" version

mcara (changes requested):
1. `AstroDrizzle` in the `DrizzlePac` has a parameter `stepsize`: https://drizzlepac.readthedocs.io/en/deployment/astrodrizzle.html. It really functions as `fast` from this PR: when `stepsize=1` => `fast=False`. When `stepsize>1` => `fast=True` and you can control how much "faster". We could have a parameter `pixmap_step` to do the same. No need for `fast`, IMO.
2. I am worried about using cubic splines because of their tendency to wiggle. Inverse of pixmap is used to compute scan regions via polygon intersections (not directly by the resampling algorithm) and this wiggling is bound to create issues (for example, you may discover that some pixels near the edges of the images may miss). I kind of strongly believe we should stay with 1st order for now. Maybe do some investigation and see if interpolating functions (for the X and Y coordinates) are  wiggly both for the current Roman distortions and also for realistic JWST models. If @schlafly thinks that it's OK to use 3rd order considering potential issues, I will approve it. Even then, I think we need a parameter that can select order.
3. This change really needs a unit test on some typical distortion model. Make sure computed values with `fast=True` (or `pixmap_step=5 or 10`) are close to those with `fast=False` to within expected accuracy (1e-3? 1e-4?)

src/stcal/outlier_detection/utils.py
```diff
@@ -283,8 +289,36 @@ def calc_gwcs_pixmap(in_wcs, out_wcs, in_shape):
     bb = wcs_bbox_from_shape(in_shape)
     log.debug("Bounding box from data shape: {}".format(bb))
 
-    grid = gwcs.wcstools.grid_from_bounding_box(bb)
-    return np.dstack(reproject(in_wcs, out_wcs)(grid[0], grid[1]))
+    if fast and len(in_shape) == 2:
```
mcara: Why is `len(in_shape) == 2` important here? How is the code different from the case of `fast=False`? That is, would the code handle more than 2 axes when `false=False`?

src/stcal/outlier_detection/utils.py
```diff
@@ -283,8 +289,36 @@ def calc_gwcs_pixmap(in_wcs, out_wcs, in_shape):
...
+
+        # Make the 1D grids we want the results on, then make
+        # coarser versions with 10% of the points (minimum 10).
+
+        x = gwcs.wcstools.grid_from_bounding_box(bb[1])
+        y = gwcs.wcstools.grid_from_bounding_box(bb[0])
+
+        x_coarse = np.linspace(x[0], x[-1], max(len(x)//10, 10))
+        y_coarse = np.linspace(y[0], y[-1], max(len(y)//10, 10))
```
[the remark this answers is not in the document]
mcara: Also, the grid seems to be mostly _but not totally_ uniform: the last column/row can be quite narrow (as little as 1 pixel). I would suggest using 10 (or `pixmap_step`) as an initial approximation => number of nodes => uniform step (`<=pixmap_step` to err on the side of accuracy).

mcara (changes requested):
I don't know whether my previous review comment will be preserved so I will repost it here (sorry if it happens to be a duplicate):

1. `AstroDrizzle` in the `DrizzlePac` has a parameter `stepsize`: https://drizzlepac.readthedocs.io/en/deployment/astrodrizzle.html. It really functions as `fast` from this PR: when `stepsize=1` => `fast=False`. When `stepsize>1` => `fast=True` and you can control how much "faster". We could have a parameter `pixmap_step` to do the same. No need for `fast`, IMO.
2. I am worried about using cubic splines because of their tendency to wiggle. Inverse of pixmap is used to compute scan regions via polygon intersections (not directly by the resampling algorithm) and this wiggling is bound to create issues (for example, you may discover that some pixels near the edges of the images may miss). I kind of strongly believe we should stay with 1st order for now. Maybe do some investigation and see if interpolating functions (for the X and Y coordinates) are  wiggly both for the current Roman distortions and also for realistic JWST models. If @schlafly thinks that it's OK to use 3rd order considering potential issues, I will approve it. Even then, I think we need a parameter that can select order.
3. This change really needs a unit test on some typical distortion model. Make sure computed values with `fast=True` (or `pixmap_step=5 or 10`) are close to those with `fast=False` to within expected accuracy (1e-3? 1e-4?)

src/stcal/outlier_detection/utils.py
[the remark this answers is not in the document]
t-brandt: No, the step would not be 400 pixels.  The number of points is the maximum of 10 and len(x)//10: no fewer than 10 steps, and no more than about 10 pixel gaps between steps.

src/stcal/outlier_detection/utils.py
```diff
@@ -287,8 +298,39 @@ def calc_gwcs_pixmap(in_wcs, out_wcs, in_shape):
...
+
+        # Finally, fill in the rest of the pixels using bilinear
+        # interpolation and return the result.
+
+        fx = interpolate.RectBivariateSpline(x_coarse, y_coarse,
+                                             pixmap_coarse[0], kx=order, ky=order)
+        fy = interpolate.RectBivariateSpline(x_coarse, y_coarse,
+                                             pixmap_coarse[1], kx=order, ky=order)
+
+        return np.dstack([fx(x, y), fy(x, y)])
```
mcara:
```suggestion
        # evaluate spline on the full grid:
        x = np.arange(shape[1])
        y = np.arange(shape[0])
        return np.dstack([fx(x, y), fy(x, y)])
```

src/stcal/outlier_detection/utils.py
```diff
@@ -279,6 +279,17 @@ def calc_gwcs_pixmap(in_wcs, out_wcs, in_shape):
...
 
+    stepsize : int, optional
+        If stepsize>1, perform the full calculation on a sparser grid
+        and use interpolation to fill in the rest of the pixels.
+        Recommended if the underlying distortion correction is smooth.
+        Default 1.
+
+    order : int, optional
+        Order of the 2D spline to interpolate the sparse pixel mapping
+        if stepsize>1.  Should be either 1 (bilinear) or 3 (bicubic).
```
mcara:
```suggestion
        if ``stepsize>1``.  Supported values are: 1 (bilinear) or 3 (bicubic).
```
although, I believe, `scipy` supports orders 2 and maybe higher than 3

src/stcal/outlier_detection/utils.py
```diff
@@ -287,8 +298,39 @@ def calc_gwcs_pixmap(in_wcs, out_wcs, in_shape):
...
+
+        pixmap_coarse = reproject(in_wcs, out_wcs)(*sparsegrid)
+
+        # Finally, fill in the rest of the pixels using bilinear
+        # interpolation and return the result.
+
+        fx = interpolate.RectBivariateSpline(x_coarse, y_coarse,
+                                             pixmap_coarse[0], kx=order, ky=order)
+        fy = interpolate.RectBivariateSpline(x_coarse, y_coarse,
+                                             pixmap_coarse[1], kx=order, ky=order)
```
mcara: `RectBivariateSpline` also takes `bbox` parameter that may be helpful. I am not sure.

mcara (commented):
This PR really needs some unit test as well.

t-brandt:
Withdrawing this PR in favor of https://github.com/spacetelescope/stcal/pull/488.

PR #368, opened 2025-05-12
JP-3906: Extrapolate darks when science data is longer than provided dark

tapastro:
Resolves [JP-3906](https://jira.stsci.edu/browse/JP-3906)

This PR adds functionality to the dark_current algorithm that will extrapolate a provided dark to fully cover the science data frame length, when the provided dark does not cover enough frames/groups. 

Ideally this code will not impact general data processing until instrument teams opt to deliver truncated darks; these darks should have a difference between the final two frames that is a high-SNR rate to be used for extrapolation to arbitrary dark lengths.

That said, this code will be enabled for any science data that exceeds the existing dark lengths; the existing behavior is to issue a warning, skip dark_current subtraction and continue processing.

## Tasks

</details>


drlaw1558:
@tapastro Looks like this code is working as expected; using the short MIRI darks I created I'm seeing the same extrapolated values here as I get in the code that I used to construct the extrapolated dark files.  These files themselves may not be perfect, but the algorithm lines up with expectations.

One caveat though with regard to how the ERR array is being handled.  Right now the code requires dark.sci and dark.err to have the same shape.  This is doubly pointless though; the ERR values in the reference files are not meaningful, and the pipeline never uses them anyway (uncertainty in the dark is handled in the ramp fitting step via the average_dark parameter).  Indeed, https://jira.stsci.edu/browse/JP-3827 just removed the 4D error array from the ramp model entirely.

Short MRS darks will still have at least 30 groups to cover the non-linear portion of the ramp, and it thus seems unnecessary to have 30 groups of meaningless ERR arrays inflating the file size.  At the same time, it would be tedious to have to redeliver ALL darks in order to remove the ERR array from the data model.

I'd thus be inclined to remove all mention of the ERR arrays from dark_sub.py, and clean up the datamodels by removing them from the dark data model.  If I'm remembering correctly, it would then be ok from a file validation perspective if existing darks had ERR arrays (which would just be 'extra' information), but newly-delivered files could choose to drop the ERR extension completely?

Thoughts?  Tagging @t-brandt as well since this would affect Roman too.

tapastro:
I had meant to bring up the error arrays in the darks but it slipped my mind - I agree that they don't appear to be used at all, despite them being pushed around in this step. I'm all for removing them from the step if they don't provide any benefit. I also understand the datamodel situation to be that the extra error array is considered valid but unnecessary, and no validation issue should arise, but I can open an stdatamodels PR and check.

I'm not sure there is any overlap with Roman at this point - they use their own datamodels and no longer use the stcal dark_current step. See https://stsci.slack.com/archives/C01FP900P38/p1746727930210279

schlafly:
I agree that there are no Roman consequences.  We can talk about bringing these back together, but the Roman dark current treatment is very simple and didn't need shipping out to stcal.

melanieclarke (approved):
This looks good to me.  Can you please link here regtests for jwst and romancal?

I think this should not need the stdatamodels PR to merge first, but will need to be coordinated with the jwst PR, since it is a breaking change.

tapastro:
JWST RTs: https://github.com/spacetelescope/RegressionTests/actions/runs/15053943380 - this shows the expected failure that will need okifying once both stcal and jwst PRs are merged.

Romancal RTs are passing : https://github.com/spacetelescope/RegressionTests/actions/runs/15146550952

The downstream unit tests are failing due to xdist issues, I think.

schlafly (approved):
Approving for rcal (which doesn't use dark_current).

PR #370, opened 2025-05-23
JP-3680: Scale drizzle errors and variances

drlaw1558:
This PR resolves [JP-3680](https://jira.stsci.edu/browse/JP-3680) (incorrect drizzle errors at non-standard output scales) by scaling the errors and variances by the square root of the input/output pixel area ratio.

## Tasks

</details>

### DIFF
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -1103,9 +1103,9 @@ class Resample:
                 scaling = 1.0
             else:
                 if is_imaging_wcs(self.output_wcs):
-                    scaling = self.pixel_scale_ratio
+                    scaling = self.pixel_scale_ratio ** 2
                 else:
-                    scaling = 1. / np.sqrt(self.pixel_scale_ratio)
+                    scaling = 1. / self.pixel_scale_ratio
 
             for varname in self.variance_array_names:
                 varwsum = self._variance_info[varname]['wsum']


src/stcal/resample/resample.py
```diff
@@ -1085,8 +1085,9 @@ def finalize_resample_variance(self, output_model):
             for varname in self.variance_array_names:
                 varwsum = self._variance_info[varname]['wsum']
                 weight = self._variance_info[varname]['wt']
-                output_model[varname] = (varwsum / (weight * weight)).astype(
-                    dtype=self.output_array_types[varname]
+                output_model[varname] = (varwsum / (weight * weight)
```
kmacdonald-stsci, line 1088: I recommend not doing `x/y/z`.  Visually, this is ambiguous and could mean `x/(y/z)` or `(x/y)/z`.  It's the latter and mathematically equivalent to `x/(y*z)`.  I recommend changing the above `x/y/z` to `x/(y*z)` because it's easier to read and computationally a double divide is more expensive than one multiplication and one divide.

kmacdonald-stsci (approved):

src/stcal/resample/resample.py
drlaw1558: Done as part of the refactor to handle imaging vs spectroscopic modes, see if you think this is more clear now.

melanieclarke:
Looks like there's a copy-paste error in your last change - the var scaling should be squared.

drlaw1558:
Yup, that would do it.  Fixed now.

melanieclarke:
> The fundamental problem with this particular implementation is that this scaling is done in the `finalize` method and it uses a single `pixel_scale_ratio` value (likely from the last input image or so) and it applies it to quantities computed from multiple input images.

I think that should no longer be an issue once we get #352 done -- all we need is the user-set ratio between the output scales and input scales, which should be the same for all input images processed together.

mcara:
The only place would be in `add_image` because it can treat each input image individually.

> I think it's still probably close enough for the approximate error purposes for most use cases

This is likely the case for "standard" processing (archive?) But for custom processing by the end user, there is no guarantee that pixel scales are not quite different (but I do not know that - I just wanted to mention this possible problem). Also `tweakreg` can modify nominal scales too (just something else to keep in mind).

PR #372, opened 2025-06-03
JP-3891: Use the zeroframe in likelihood fitting

melanieclarke:
Resolves [JP-3891](https://jira.stsci.edu/browse/JP-3891)

If a zeroframe is downlinked separately, use it in likelihood ramp fitting as a separate datapoint from the first group.

## Tasks

</details>


src/stcal/ramp_fitting/likely_fit.py
```diff
@@ -62,22 +47,48 @@ def likely_ramp_fit(ramp_data, readnoise_2d, gain_2d):
 
     readtimes = get_readtimes(ramp_data)
 
+    # Check for a zeroframe
```
melanieclarke, line 50: @t-brandt - Can you please check my implementation here (lines 50-84)?  The rest of this PR is just some minor clean up for things I noted in passing.  These lines add the changes for the zeroframe handling.

I think this is implemented as described in the ticket, but would appreciate you double checking my interpretation.

melanieclarke:
jwst regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/15541145016
romancal regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/15541178480

src/stcal/ramp_fitting/likely_fit.py
```diff
@@ -62,22 +47,48 @@ def likely_ramp_fit(ramp_data, readnoise_2d, gain_2d):
...
+    if ramp_data.zeroframe is not None:
+        # Read time for the zeroframe is the first entry in the first group
+        rt_zero = readtimes[0][0]
+
+        # Remove the zero readtime from the first group
+        readtimes[0] = readtimes[0][1:]
```
t-brandt, line 56: If the zero frame is downlinked separately, the first group should always have >1 reads, but maybe we want to catch an exception to the above with an error message that the zero frame exists *and* the first element of readtimes does not have length >1?

src/stcal/ramp_fitting/likely_fit.py
```diff
@@ -62,22 +47,48 @@ def likely_ramp_fit(ramp_data, readnoise_2d, gain_2d):
...
     for integ in range(nints):
-        data = ramp_data.data[integ, :, :, :]
-        gdq = ramp_data.groupdq[integ, :, :, :].copy()
-        pdq = ramp_data.pixeldq[:, :].copy()
+        data = ramp_data.data[integ]
+        gdq = ramp_data.groupdq[integ].copy()
+        pdq = ramp_data.pixeldq.copy()
+
+        # Use the zeroframe if available
+        if ramp_data.zeroframe is not None:
```
t-brandt, line 72:
```suggestion
        if hasattr(ramp_data, 'zeroframe') and ramp_data.zeroframe is not None:
```
As before, to make it easier to use this with Roman

src/stcal/ramp_fitting/likely_fit.py
```diff
@@ -62,22 +47,48 @@ def likely_ramp_fit(ramp_data, readnoise_2d, gain_2d):
...
+            # Zeroframe image for this integration (2D)
+            zf = ramp_data.zeroframe[integ]
+
+            # Borrow the DQ image for the first group
+            zfdq = gdq[0].copy()
+
+            # Subtract the zeroframe data from the first group
+            first_group = (data[0] * ramp_data.nframes - zf) / (ramp_data.nframes - 1)
```
t-brandt, line 80: Should we explicitly catch the case where `nframes=1`?  E.g., 
`if ramp_data.nframes == 1:`
`    raise ValueError("Not expecting the zero frame to be separated when there is one frame per group.")`
This should never happen, so it's me being a little paranoid.  Without the catch the code results in division by zero.

t-brandt (commented):
Looks great, thank you.  I apologize for the delay in my review.

melanieclarke:
@tapastro - I think this is ready for review now.  Do you want to try to get it in today or hold off for next build?

tapastro:
I think it's fine to get in - it's an update on an optional algorithm, so I don't expect any issues to be caused for DMS. I'll review.

tapastro (approved):

PR #375, opened 2025-06-10
fix downstream by not running in parallel

braingram:
This PR fixes the downstream jwst and romancal jobs by not running them in parallel. This will cause these jobs to take longer but will work around 2 issues:

- pytest-cov and pytest-xdist issue handling warnings: https://github.com/pytest-dev/pytest-cov/issues/693 (since jwst now [turns warnings into errors](https://github.com/spacetelescope/jwst/pull/9490))
- crds issues due to parallel reference file fetching: https://github.com/spacetelescope/crds/issues/930

There is also a somewhat related issue where stray files from opencv are also causing run files See https://github.com/spacetelescope/jwst/pull/9512 for more details but since the coverage configuration for the jwst job uses the pyproject.toml in this directly we need to duplicate those jwst changes here. jwst also either includes or produces a "rmap" file. This is also ignored for coverage.

## Tasks

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,6 @@ jobs:
         CRDS_CLIENT_RETRY_COUNT: 3
         CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
       envs: |
-        - linux: py312-jwst-cov-xdist
-        - linux: py312-romancal-cov-xdist
+        - linux: py312-jwst-cov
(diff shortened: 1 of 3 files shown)


braingram:
> LGTM. Should the rests directory be omitted from coverage?

My vote would be to keep the tests directory as part of the coverage measurement. It can be useful for things like duplicated tests (which otherwise pytest just ignores).

PR #381, opened 2025-07-09
Add caching to get_catalog

WilliamJamieson:
The `http://gsss.stsci.edu/webservices` server we use to grab the catalog information for tweakreg can sometimes be overwhelmed by requests for catalogs made during CI runs, especially if there are a bunch of PRs all running tweakreg tests at around the same time. This is making a lot of the testing flaky.

This PR addresses this by assuming that the catalogs are not going to be updated during any particular run of jwst or romancal so it caches catalog requests made during the lifetime of those runs. This is a port of spacetelescope/romancal#1833 to `stcal`.

Note this could be simplified to using the `functools.lru_cache`; however, that is not multiprocessing-safe (but it is thread-safe). I know we sometimes use multiprocessing and it isn't clear if we would be using it during `tweakreg` related things, so I went ahead and made it multiprocessing-safe, which makes it thread-safe as well.

Note I also transferred several unit tests that should have been running on the stcal code but were running on a dead copy of that code in the `romancal` tests which have now been removed. See spacetelescope/romancal#1835.

## Tasks

</details>

### DIFF
--- a/changes/381.general.rst
+++ b/changes/381.general.rst
@@ -1,0 +1,4 @@
+Add caching to the ``tweakreg.astrometric_utils.get_catalog`` function to avoid
+repeated calls to the VO web service for the same parameters. This will improve
+performance and reduce the load on the service.  Also, testing from romancal for
+``get_catalog`` has been moved to stcal.
--- a/src/stcal/tweakreg/astrometric_utils.py
+++ b/src/stcal/tweakreg/astrometric_utils.py
@@ -1,12 +1,19 @@
+from __future__ import annotations
+
 import os
+from multiprocessing import Manager
+from typing import TYPE_CHECKING
 
 import requests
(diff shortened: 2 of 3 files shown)


schlafly (commented):
Looks good to me.  I am happy to also reduce the number of queries via things like:
- fewer-to-no "garbage" queries (test_get_catalog_using_invalid_parameters)
- test_catalog_using_epoch to 2 epochs and one location instead of 3x3 grid
- test_create_catalog_variable_num_sources to just one or two num_sources instead of 3
- test_create_catalog_different_epochs --> delete, already covered by test_catalog_using_epoch

But the caching will clearly make a big impact.

WilliamJamieson:
The testing suggestions are fine, I simply did a direct port of the former romancal tests here.

PR #384, opened 2025-08-08
Move the default timeout to the call level

stscieisenhamer:
WIP [RCAL-1130](https://jira.stsci.edu/browse/RCAL-1130)

This PR addresses an issue where the call to the VO catalog service times out due to crowded fields in the Roman FOV. The current solution is a stop-gap measure that provides access to the timeout argument in tweakreg.astrometric_utils.create_astrometric_catalog. Basically the default timeout provided in the module is moved to the call of `get_catalog`, allowing external modification of the module-level TIMEOUT parameter.

This avoids api changes at a number of levels while a more permanent solution is determined.

## Tasks

</details>

### DIFF
--- a/src/stcal/tweakreg/astrometric_utils.py
+++ b/src/stcal/tweakreg/astrometric_utils.py
@@ -90,7 +90,9 @@ def create_astrometric_catalog(
         fiducial[1],
         epoch=epoch,
         search_radius=radius,
-        catalog=catalog)
+        catalog=catalog,
+        timeout=TIMEOUT
+    )
     if len(ref_dict) == 0:
         return ref_dict
 


braingram:
How about instead we:
- add `timeout` as an argument to [absolute_align](https://github.com/spacetelescope/stcal/blob/879edb8deb12c5ecea3b79f6bcdc5231bd4788dd/src/stcal/tweakreg/tweakreg.py#L116)
- provide a different timeout in romancal
- update [_parse_ref_cat](https://github.com/spacetelescope/stcal/blob/879edb8deb12c5ecea3b79f6bcdc5231bd4788dd/src/stcal/tweakreg/tweakreg.py#L218) and [create_astrometric_catalog](https://github.com/spacetelescope/stcal/blob/879edb8deb12c5ecea3b79f6bcdc5231bd4788dd/src/stcal/tweakreg/astrometric_utils.py#L32) to pass the timeout to `get_catalog`

That would avoid the global state change currently added in https://github.com/spacetelescope/romancal/pull/1884

stscieisenhamer:
I'm good with all that; just didn't want to walk through all the api changes until all are happy.

braingram:
Thanks. Overall looks good to me. Would you run the jwst and romancal regression tests?

PR #385, opened 2025-08-12
add velocity aberration calculations

stscieisenhamer:
Resolves [RCAL-1123](https://jira.stsci.edu/browse/RCAL-1123)

Required by romancal [PR#1891](https://github.com/spacetelescope/romancal/pull/1891)

This PR implements the common base velocity aberration calculation used by the command `roman_set_velocity_aberration`. 

## Tasks

</details>

### DIFF
--- a/changes/385.general.rst
+++ b/changes/385.general.rst
@@ -1,1 +1,1 @@
-add velocity aberration calculations
+Add utility functions for calculating velocity aberration.
--- a/changes/385.general.rst
+++ b/changes/385.general.rst
@@ -1,0 +1,1 @@
+add velocity aberration calculations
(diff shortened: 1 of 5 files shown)


schlafly (commented):
Thanks Jonathan.  The romancal failures look transient and the regtests pass, so I think this is good from a romancal perspective.

The thought here is that this is common code that makes more sense in stcal than in romancal, and Webb code could migrate to use this as well at some point in the future.  But if folks don't like that we can stick this in romancal instead.

changes/385.general.rst
```diff
@@ -0,0 +1 @@
+add velocity aberration calculations
```
emolter, line 1:
```suggestion
Add utility functions for calculating velocity aberration.
```

schlafly (approved):

PR #407, opened 2025-09-22
Add code for combining sregions

emolter:
Helps to resolve [JP-4100](https://jira.stsci.edu/browse/JP-4100) along with https://github.com/spacetelescope/jwst/pull/9856

This PR adds methods for combining footprints from input images into a combined output footprint using Shapely, and converting that footprint into an S_REGION keyword that MAST can understand.

There are examples of this code running on synthetic and real data on [this JIRA ticket](https://jira.stsci.edu/browse/JP-4100), as well as some discussion of why the Shapely dependency was needed here and why `spherical_geometry` was found to be unsuitable as a replacement.

Please note that `photutils` requires `shapely`, so it is already an indirect dependency of both `romancal` and `jwst`.

## Tasks

</details>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
       - id: text-unicode-replacement-char
   # lint and format
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.11
+    rev: v0.13.3
     hooks:
       - id: ruff-check
         args:
@@ -77,21 +77,21 @@ repos:
   #     - id: numpydoc-validation
   # format code examples in docstrings
   - repo: https://github.com/adamchainz/blacken-docs
-    rev: 1.19.1
+    rev: 1.20.0
     hooks:
       - id: blacken-docs
         additional_dependencies:
           - black==22.12.0
   # sort imports in Cython code
   - repo: https://github.com/pycqa/isort
-    rev: 6.0.1
+    rev: 6.1.0
     hooks:
       - id: isort
         name: isort (cython)
         types: [cython]
   # lint Cython code
   - repo: https://github.com/MarcoGorelli/cython-lint
-    rev: v0.16.7
+    rev: v0.17.0
     hooks:
       - id: cython-lint
       - id: double-quote-cython-strings


src/stcal/alignment/resample_utils.py
```diff
@@ -39,3 +40,184 @@ def calc_pixmap(in_wcs, out_wcs, shape=None):
...
+        combined_footprints = [combined_footprints]
+    elif isinstance(combined_footprints, shapely.geometry.MultiPolygon):
+        combined_footprints = combined_footprints.geoms
+    combined_polys = []
+    for poly in combined_footprints:
+        x, y = poly.exterior.coords.xy
+        combined_poly = np.vstack([x, y]).T
+        combined_poly = combined_poly[:-1]  # remove duplicate last point
+        combined_poly = _simplify_by_angle(combined_poly)
+        combined_polys.append(combined_poly)
```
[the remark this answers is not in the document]
emolter: let me know how https://github.com/spacetelescope/stcal/pull/407/commits/13de0546d09cb10d73a2c4a3afdcc6e90bf96bb2 looks

src/stcal/alignment/resample_utils.py
```diff
@@ -39,3 +40,184 @@ def calc_pixmap(in_wcs, out_wcs, shape=None):
...
+    # combine footprints with Shapely
+    combined_polygons = combine_footprints(footprints_pixels)
+
+    # intersect with output WCS footprint
+    if intersect_footprint is not None:
+        x, y = world2det(intersect_footprint[:, 0], intersect_footprint[:, 1])
+        intersect_footprint_pixels = np.vstack([x, y]).T
+        final_polygons = _intersect_with_bbox(combined_polygons, intersect_footprint_pixels)
+        if not final_polygons:
+            raise ValueError("No overlap between input s_regions and intersection footprint")
```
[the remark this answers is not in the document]
emolter: added in https://github.com/spacetelescope/stcal/pull/407/commits/13de0546d09cb10d73a2c4a3afdcc6e90bf96bb2

tests/test_combine_sregions.py
```diff
@@ -0,0 +1,418 @@
...
+            [10.0, 11.0],
+        ]
+    )
+    sregion_list = _footprints_to_sregion_list(complex_footprint_set)
+    with pytest.raises(ValueError) as excinfo:
+        combine_sregions(sregion_list, det2world, intersect_footprint=bbox)
+    assert str(excinfo.value) == "No overlap between input s_regions and intersection footprint"
+
+
+def test_duplicate_tolerance():
```
emolter, line 399: @tapastro This test is to ensure that we follow the tolerance requirements for duplicated points set by MAST.

RE your comment about repeatability of testing, I agree but I think it's fine to basically make any choice as long as it's deterministic.  The way the code is currently written, it will always just keep the first point in the polygon.  The Shapely union orders the points clockwise around the polygon.

schlafly (approved):

src/stcal/alignment/resample_utils.py
```diff
@@ -39,3 +41,211 @@ def calc_pixmap(in_wcs, out_wcs, shape=None):
...
+    -------
+    str
+        The combined s_region.
+    
+    Raises
+    ------
+    ValueError
+        If there is no overlap between the input s_regions and the intersection footprint.
+    """
+    footprints = np.array([util.sregion_to_footprint(sregion) for sregion in sregion_list])
```
braingram, line 73:
```suggestion
    footprints = np.array([util.sregion_to_footprint(sregion) for sregion in sregion_list])
```
Would you pull this conversion from strings to footprints outside of this function or make it conditional? That way if romancal wants to use this it can provide the the [footprints computed from L2 files](https://github.com/spacetelescope/romancal/blob/450348798730a7e11a88cb7383dd512ecdc0ccdf/romancal/resample/resample_utils.py#L77) without needing to first convert them to S_REGION strings (with a loss of precision).

emolter: Yeah I can definitely do that.  What would be the most useful?

Would it make sense to have a method `combine_footprints_world(footprints, det2world)` that takes in a list of footprints and outputs a list of combined footprints?  And then make `_polygons_to_sregion` public, so if you did want to go all the way to an output s_region you could call that one?

Would it make sense to make the current `combine_footprints()` private, since it operates only in pixel space?

I think the easiest thing would be to make it a conditional but that solution seems clunky to me.

emolter: ok to keep it consistent with `wcs_from_sregions` I just made it conditional.  How does https://github.com/spacetelescope/stcal/pull/407/commits/7f5ede5bd28c8b7934c2b90a3b7ce4fa4e81f236 look?  Unit test for it is here https://github.com/spacetelescope/stcal/pull/407/commits/a998995f19f9e3b1d615d848eadf7f3fa6512e13

braingram: Looks good to me. Thanks!

PR #410, opened 2025-09-24
Jump two-point difference calc early return case has too many items

tapastro:
This PR addresses a bug found in testing, where input data triggering this if: https://github.com/spacetelescope/stcal/blob/101ba66a426db0afff7f9130e8128a99cf715ad1/src/stcal/jump/twopoint_difference.py#L63-L65

attempts to return five items into an explicit four item tuple: https://github.com/spacetelescope/stcal/blob/101ba66a426db0afff7f9130e8128a99cf715ad1/src/stcal/jump/jump.py#L90

 Haven't checked codecov, but I'm guessing we don't have test coverage of these lines?

## Tasks

</details>


braingram (approved):
Thanks for finding that return I missed. LGTM.

braingram:
Would you add a changelog since the release 1.15.1 version has the bug fixed in this PR?

PR #412, opened 2025-09-25
Rescale resample kernels by pixel scale ratio

mcara:
Resolves [JP-4137](https://jira.stsci.edu/browse/JP-4137)

This PR addresses ...

## Tasks

</details>


src/stcal/resample/resample.py
```diff
@@ -813,7 +813,9 @@ def add_model(self, model):
             'ymax': ymax,
         }
 
-        self._driz.add_image(data, **add_image_kwargs)
+        f = (iscale / self.pixel_scale_ratio)**2
```
nden, line 816: Can you rename `f` to something more descriptive?

src/stcal/resample/resample.py
```diff
@@ -802,7 +802,7 @@ def add_model(self, model):
         add_image_kwargs = {
             'exptime': model["exposure_time"],
             'pixmap': pixmap,
-            'scale': iscale,
+            'scale': self.pixel_scale_ratio,
```
nden, line 805: If `scale` is now scale ratio, at least add a comment to it to avoid confusion in the future.

src/stcal/resample/resample.py
nden: I'm not very familiar with the code and I don't understand this.  
For flux density images `iscale` is defined as `1/sqrt(pixel_scale_ratio)` so in this case 
`f = 1 / pixel_scale_ratio**3`. For flux units, `f = 1 / pixel_scale_ratio**2`.  And the data is scaled by this. In the pipeline we set `pixel_scale_ratio=1` so it doesn't matter in the default case. How about user supplied scales?

mcara:
> No impact on Roman regression tests, so I'm approving. But I'll admit I'm not following the code here at all. drizzle.add_image(..., scale=) is supposed to specify the pixel scale of the input image, or other things depending on the algorithm; it's weird to me that iscale (the old code) was ever right there? Presumably then "f" is taking out the factor of the scale squared that the change to the "scale" argument is putting in, though in my head that would be something that drizzle is directly responsible for rather than the stcal wrapper. And then all of the add_image calls need to get that same factor.

A lot of discussion related to scale was in https://github.com/spacetelescope/jwst/issues/9246. However, there I missed that some kernels use `scale` also to compute kernel size. So, by forcing `scale=iscale=1` (at least for the imaging) we make a mistake in kernel size when actual pixel scale ratio is not 1. This PR fixes that by supplying `scale=pixel_scale_ratio` which is correct for kernel sizing but now input image would be multiplied by `scale^2` and so I divide them by the same factor before passing them to `drizzle` so that overall input images are divided and multiplied by `scale^2` resulting in a total multiplicative factor of 1 (for the imaging case). For spectral, the correction factor is the original `iscale` (multiplicative) divided by `scale^2` resulting in the old multiplicative factor `iscale`.

mcara:
@schlafly

> Maybe one conceptual question---I had understood that a goal of drizzle was to preserve surface brightness; if an image with constant surface brightness equal to X goes in, one comes out with value equal to X everywhere. If "scale" changes the output image scale, that means to me that that isn't right, and the input image is assumed to be closer to flux units rather than surface brightness units, and therefore needs to get scaled by 'scale' to get to surface brightness units. Maybe this is elementary, sorry!

I believe the above statement is correct.

PR #418, opened 2025-10-20
Rescale resample kernels by pixel scale ratio - V2

mcara:
Resolves [JP-4137](https://jira.stsci.edu/browse/JP-4137)
Closes spacetelescope/stcal#412

This PR fixes a bug in the resample code which is caused by the fact that we fix the `iscale` to be 1 while `drizzle` uses it in some kernels (turbo, gaussian, and lanczos) not only to apply a scale to input image data but also to compute kernel size in the input image frame from sizes in output image (that is, `drizzle` expects it to be pixel scale ratio). 

This PR takes a different approach from spacetelescope/stcal#412: `iscale` and (old) `kscale` (now `pixel_scale_ratio`) have been decoupled in https://github.com/spacetelescope/drizzle/pull/203 and this PR makes necessary code changes in `stcal` to work with the upcoming `drizzle` release that will include https://github.com/spacetelescope/drizzle/pull/203

## Tasks

    - [x] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/runs/18666316914)

</details>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ classifiers = [
 ]
 dependencies = [
   "astropy >=6.0.0",
-  "drizzle@git+https://github.com/mcara/drizzle.git@decouple-scales",
+  "drizzle >=2.2.0",
   "scipy >=1.14.1",
   "scikit-image>=0.21.0",
   "numpy >=1.25.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ classifiers = [
 ]
 dependencies = [
   "astropy >=6.0.0",
-  "drizzle >=2.0.1",
+  "drizzle@git+https://github.com/mcara/drizzle.git@decouple-scales",
   "scipy >=1.14.1",
   "scikit-image>=0.21.0",
   "numpy >=1.25.0",
--- a/src/stcal/outlier_detection/utils.py
+++ b/src/stcal/outlier_detection/utils.py
@@ -269,7 +269,7 @@ def gwcs_blot(median_data, median_wcs, blot_shape, blot_wcs, pix_ratio, fillval=
         pixmap=pixmap,
         out_img=outsci,
         fillval=fillval,
-        iscale=1.0,
+        iscale=1.0 / (pix_ratio * pix_ratio),
         interp="linear",
         sinscl=1.0,
     )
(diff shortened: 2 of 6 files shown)


braingram (changes requested):
Just marking a requested change for the dependency change so we don't commit a git dependency to main.

kmacdonald-stsci (approved):

mcara:
Actually, there may be improved memory usage from the drizzle update. In https://github.com/spacetelescope/drizzle/pull/208/files#diff-4f0a4f497f200684bb05f3b851ca50a436e435e6488340614704a3793a64f47b my hope was that it may reduce making array copies. But I am not sure. If you want, wait until tomorrow and re-run tests and compare memory.

@braingram I rebased `romancal` PR and restarted tests.

schlafly (approved):
Thanks Mihai!

src/stcal/outlier_detection/utils.py
```diff
@@ -263,15 +263,20 @@ def gwcs_blot(median_data, median_wcs, blot_shape, blot_wcs, pix_ratio, fillval=
...
+        data=median_data,
+        pixmap=pixmap,
+        out_img=outsci,
+        fillval=fillval,
+        # scaling of the input pixel is unnecessary since outlier detection
+        # is based on SNR (ratio being a key word here). However, to preserve
+        # the same accuracy loss as before, we keep the scaling in order for
+        # the regression tests to pass.
+        # TODO: Consider setting iscale=1 to avoid accuracy loss and simplify
+        # the logic.
```
braingram: Would you open up a follow-up jwst ticket/issue for this?

braingram (approved):
LGTM. Thanks for making the updates.

pyproject.toml
```diff
@@ -11,7 +11,7 @@ classifiers = [
 ]
 dependencies = [
   "astropy >=6.0.0",
-  "drizzle >=2.0.1",
+  "drizzle >=2.2.0",
```
braingram: @mcara would you open a jwst PR to allow for this new version. jwst currently has an upper pin for drizzle that will cause issues when this stcal PR is merged:
https://github.com/spacetelescope/jwst/blob/471c68ed97f484455f86da11f52556c28425c3de/pyproject.toml#L25

mcara: https://github.com/spacetelescope/jwst/pull/10150

emolter (approved):
I think this looks good, since all JWST regression tests pass.  I found one typo (repeated text).

I don't really understand the reason why it's a bad idea to divide by the pixel scale ratio, or what the differences are in practice.  But since that change/bugfix isn't applied to JWST data here, no worries.

Issue #335, opened 2025-02-03
Codecov bot is not ignoring conftest.py


emolter:
A rule should be added to the Codecov bot's ignore list to ignore `conftest.py` files, e.g. the one [added here](https://github.com/spacetelescope/stcal/pull/320/files/c0b9971de130372880ef4877a110635742913857..d8a1d3b15a828524c1bea138a07c01664ff52df6).  These are part of the testing suite, so they don't themselves need test coverage.

braingram:
I'm in favor of having coverage measurement for test files. It helps to find unused fixtures, logic branches in tests that aren't exercised and duplicate tests. For example, pytest won't complain if you define 2 tests with the same name but will only run the last one defined. Having coverage for test files would show the first test is "uncovered".

PR #426, opened 2025-10-31
TST: Improve coverage reporting

pllim:
This PR addresses coverage weirdness I saw in #425 . It was checking for coverage of a test module, which it is not supposed to do, among other things.

No need for change log nor RT.

Maybe related:

* https://github.com/spacetelescope/stcal/issues/335

## Tasks

</details>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -207,8 +207,6 @@ archs = ["x86_64", "arm64"]
 [tool.cibuildwheel.linux]
 archs = ["auto", "aarch64"]
 
-[tool.coverage]
-
 [tool.coverage.run]
 omit = ["config.py", "config-3.py", "*.rmap", "tests/*", "*/tests/*"]
 


pllim:
The Python 3.14 failures are unrelated.

pllim:
> duplicate test names and unused fixtures

Do you really need coverage to catch these? Can't things like ruff or pre-commit do these checks? Even flake8 complains about duplicate function names if you need a really simple tool. A grep can probably smoke out unused fixtures from time to time.

I think coverage should only touch functional code, as test code coverage is meaningless to users.

What do others think?

melanieclarke:
I don't think it's common practice to include tests in your coverage statistics.  I'm in favor of leaving them out of the CI report to reduce clutter.  If you're interested in checking coverage for tests, that could easily be done locally.

pllim:
I see his point but I am not that convinced of the benefits. As you can see here, the coverage drops from 89.66% to 70.66% just by omitting the tests. 89% looks very good but 70% is start to be worrying. We thought we had good coverage of the stcal code but do we really?

braingram:
> I see his point but I am not that convinced of the benefits. As you can see here, the coverage drops from 89.66% to 70.66% just by omitting the tests. 89% looks very good but 70% is start to be worrying. We thought we had good coverage of the stcal code but do we really?

I see 85% with this PR :-/ https://github.com/spacetelescope/stcal/pull/426/checks?check_run_id=54626149077
Where is the 70 from? Is it perhaps from a report prior to the jwst and romancal tests finishing? Those are currently used to measure coverage and often do not finish before the first coverage report.

pllim:
Ehhh it was 70% earlier. What happened? 😆

pllim:
Oh OK. I must have not waited for downstream. It jumped from 70% to 77% when romancal-cov completed. 

Does that means it is also counting downstream coverage (downstream code + downstream test code)?

https://github.com/spacetelescope/stcal/blob/49ee10e7da13fa1fff13909e84a2660c54b55486/tox.ini#L66

pllim:
OK now that I wait till everything finishes, it says `85.36% (-4.32%)`. That is still a minus but not as dramatic as I claimed earlier.

@melanieclarke what do you think?

melanieclarke:
I still don't see much benefit to including the tests in the coverage, but it's not terribly important to me either way.  I do think it would be nice to increase the local code coverage, but that's a different PR.

pllim:
OK then, given Brett G has the strongest opinion here, I am just going to close this without merge. Thanks, all!

PR #441, opened 2025-11-10
JP-3964: create_astrometric_catalog() needs more control for stars without Gaia proper motions (affects TweakReg)

pllim:
Resolves [JP-3964](https://jira.stsci.edu/browse/JP-3964)

This PR addresses https://github.com/spacetelescope/jwst/issues/8168

Closes #446

## Tasks

### DIFF
--- a/src/stcal/tweakreg/astrometric_utils.py
+++ b/src/stcal/tweakreg/astrometric_utils.py
@@ -145,7 +145,6 @@ def get_catalog(
     search_radius=0.1,
     catalog="GAIADR3",
     timeout=TIMEOUT,
-    apply_pm=True,
     pm_err_max=None,
 ):
     """Extract catalog from VO web service.


src/stcal/tweakreg/astrometric_utils.py
```diff
@@ -170,6 +171,17 @@ def get_catalog(
     timeout : float, optional
         Timeout in seconds to wait for the catalog web service to respond. Default: 30.0 s
 
+    apply_pm : bool, optional
+        If True, apply proper motion when available. Otherwise,
+        do not apply proper motion at all. Default: True
```
mcara: What would be the epoch?

mcara: ... if true?

pllim: Not sure. As per ticket, True means whatever the pipeline is already doing. So what is jwst doing?

mcara: What about selecting/using only those sources with proper motion information?

mcara: > s-goldman pointed out alternative ways to filter at https://drizzlepac.readthedocs.io/en/latest/mast_data_products/pipeline-astrometric-calibration-description.html#fit-quality-and-selection-of-wcs-solution 

I don't think that link talks about selecting sources from a returned astrometric catalog.

PR #446, opened 2025-11-19
JP-3964: Filter out bad proper motion from get_catalog

pllim:
Resolves [JP-3964](https://jira.stsci.edu/browse/JP-3964)

Closes #441

This PR allows `get_catalog` to filter out rows with bad proper motion info. This would be the new default behavior. ref https://github.com/spacetelescope/jwst/issues/8168

Might get superseded by https://jira.stsci.edu/browse/AL-1006 anyway in the future.

## TODO

- [x] Buy-in from INS; they have to find edge cases and perform science case testing, etc.
- [ ] Are romancal RT failures acceptable? Eddie and/or Mihai should review.

## Tasks

  - [x] run regression tests with this branch installed (`stcal@git+https://github.com/pllim/stcal.git@tweak-and-shout-less`) 
    - [x] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/19516312167, https://github.com/spacetelescope/RegressionTests/actions/runs/20979629034
    - [x] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/19516360882, https://github.com/spacetelescope/RegressionTests/actions/runs/20979652406

### DIFF
--- a/src/stcal/tweakreg/astrometric_utils.py
+++ b/src/stcal/tweakreg/astrometric_utils.py
@@ -1,5 +1,7 @@
 import os
+import warnings
 
+import numpy as np
 import requests
 from astropy import units as u
 from astropy.coordinates import SkyCoord
@@ -28,6 +30,7 @@ def create_astrometric_catalog(
     table_format="ascii.ecsv",
     num_sources=None,
     timeout=TIMEOUT,
+    strict_cols=("pmra", "pmdec"),
 ):
     """Create an astrometric catalog that covers the inputs' field-of-view.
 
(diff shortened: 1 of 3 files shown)


pllim:
I'll keep this in draft form until we have buy-in from INS. FYI.

src/stcal/tweakreg/astrometric_utils.py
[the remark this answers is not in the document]
pllim: OK I think I get it now and pushed a commit that hopefully address your request. Sorry for the confusion and thank you for your patience!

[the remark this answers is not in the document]
tapastro: Yes, looks good! Hopefully Varun/David have some feedback for us soon, but I think your PR addresses the issue that was raised by the ticket. 👍

tests/test_tweakreg.py
```diff
@@ -395,11 +403,29 @@ def test_get_catalog_raises_connection_error(monkeypatch):
 
 @pytest.mark.parametrize("epoch", [None, 2025.1])
 def test_catalogs_match(epoch):
-    gsss = amutils.get_catalog(10, 10, search_radius=0.1, epoch=epoch, catalog="GAIADR3")
+    gsss = amutils.get_catalog(10, 10, search_radius=0.1, epoch=epoch, catalog="GAIADR3", strict_cols=None)
```
pllim: Same problem as https://github.com/spacetelescope/stcal/pull/446/changes#r2688683728 . Ideally, I should not need to set `strict_cols=None` here but not for `s3` below.

src/stcal/tweakreg/astrometric_utils.py
```diff
@@ -158,10 +172,14 @@ def get_catalog(
     timeout : float, optional
         Timeout in seconds to wait for the catalog web service to respond. Default: 30.0 s
 
+    strict_cols : tuple or None, optional
```
braingram: I'm in favor of not adding this argument and instead just using the epoch to determine if all corrected or all uncorrected sources are returned.

I think it's a bug that we're returning corrected and uncorrected sources when an epoch is provided. This function could be updated to fix that bug by only returning corrected sources when an epoch is provided and returning all (uncorrected) sources when an epoch is not provided.

Is there an expected pipeline use of `strict_cols`?

Removing this option would simplify this function and the tests required to support it.

pllim: Option is for backward compatibility if INS needs the old behavior for whatever reason. If you want the API to change again, I would have to defer to @tapastro and/or @drlaw1558 for another iteration of development.

drlaw1558: Sounds like both approaches have the same end result scientifically; either require PMs exist through strict_cols or require that PMs have been used when epoch was provided.  Seems like more a question of which is easier to maintain from a dev standpoint as new catalogs such as S3 are added.

braingram: Here's an alternative that doesn't add a `strict_cols` argument: https://github.com/spacetelescope/stcal/pull/493

PR #447, opened 2025-11-20
Add `GAIADR3_S3` catalog support in tweakreg

braingram:
Closes https://jira.stsci.edu/browse/AL-1006

This adds support for a new catalog `GAIADR3_S3` in tweakreg. The motivation is 2 fold:
- large queries often fail for the current gsss server (which has been problematic for romancal verification)
- there is discussion of shutting down the gsss server

This PR updates tweakreg to use `s3://stpubdata/gaia/gaia_dr3/public/hats/gaia/` when the `GAIADR3_S3` catalog is requested.

Note that this does not change the default catalog and jwst and romancal will both require updates to allow for the new catalog name.

Romancal regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/19876427312
JWST regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/19876432597
(failures are unrelated)

I ran a script locally that queries a 12x6 grid of ra/decs and queried gsss and s3 with a 0.1 search radius. The queries sometimes have a small difference in number of sources (since gsss returns sources outside of the search cone). The gsss queries took on average 0.7 seconds whereas the s3 queries took on average 67 seconds (with several taking longer than 2 minutes).

### DIFF
--- a/docs/stcal/tweakreg/description.rst
+++ b/docs/stcal/tweakreg/description.rst
@@ -46,6 +46,11 @@ of the input data.
 
 If ``abs_refcat`` is set to 'GAIADR3_S3' the `HATS <https://hats.readthedocs.io/en/stable/>`_
 package will be used to access an S3 bucket containing a partitioned Gaia data release 3 catalog.
+When using the S3 catalog and a specified epoch, source positions are corrected internally by
+propagating the position to the given epoch using the Gaia proper motions and parallaxes
+(and throwing away any sources without computed proper motions and parallaxes).
+Parallaxes are computed as if the observer were located at the Earth barycenter.
+
 The combined source catalog 
 then gets cross-matched and fit to this astrometric reference catalog.
 The results of this one fit then gets back-propagated to all the


braingram:
A small update on this.

Brian McLean confirmed that gsss expects epoch to be in decimal year format in utc scale (which is what we're currently providing on main). However if I compare the gsss results to the official gaia archive results the values differ.

Using source 5918660719983560192 (which has a large proper motion) if I query the gaia archive and gsss using equivalent epochs (providing the expected jyear/tcb to gaia and decimalyear/utc to gsss) the results differ by over 90 mas:
```python
from astropy.time import Time
from stcal.tweakreg.astrometric_utils import get_catalog
from astroquery.gaia import Gaia

epoch = 2025.0
sid = 5918660719983560192  # source with large proper motion

query = f"SELECT ESDC_EPOCH_PROP(ra, dec, parallax, pmra, pmdec, radial_velocity, 2016.0, {epoch}) FROM gaiadr3.gaia_source WHERE source_id = {sid}"
job = Gaia.launch_job_async(query)
results = job.get_results()
#print(results)
ra = results[0][0][0]
dec = results[0][0][1]
print(f"(deg) gaia ra:{ra} dec:{dec}")

# the above epoch is julian year tcb convert to decimal year utc
epoch_jyear = Time(epoch, format='jyear', scale='tcb')
epoch_decimal_year = Time(epoch_jyear, format='decimalyear', scale='utc')
catalog = "GAIAREFCAT"

t = get_catalog(ra, dec, epoch=epoch_decimal_year.value, catalog=catalog)
ra2 = t[t['objID'] == sid][0]['ra']
dec2 = t[t['objID'] == sid][0]['dec']
print(f"(deg) gsss ra:{ra2} dec:{dec2}")

print(f"(deg) RA diff: {ra - ra2}")
print(f"(deg) DEC diff: {dec - dec2}")
print(f"Epoch: {repr(epoch_decimal_year)}")
print(f"Catalog: {catalog}")
```
Produces:
```
INFO: Query finished. [astroquery.utils.tap.core]
(deg) gaia ra:266.6282715407062 dec:-57.328446746442594
(deg) gsss ra:266.628290426716 dec:-57.3284212141562
(deg) RA diff: -1.8886009797824954e-05
(deg) DEC diff: -2.5532286393570303e-05
Epoch: <Time object: scale='utc' format='decimalyear' value=2024.9993140093616>
Catalog: GAIAREFCAT
```
Put another way the gsss epoch propagation does not match the results from the gaia archive. Is there one that we should target? If it's the gsss results what algorithm do they use for propagation?

braingram:
> Brian McLean confirmed that gsss expects epoch to be in decimal year format in utc scale (which is what we're currently providing on main). However if I compare the gsss results to the official gaia archive results the values differ.

I was now told that the format is not decimal year and is instead julian year (so the provided epochs on main are incorrect). For scale he thinks that utc is more appropriate. I will open a separate issue for this.

tapastro:
Is it expected that the S3 queries take so long?

Is it unavoidable that pyarrow now becomes a dependency for all users of stcal? It doesn't appear to be a lightweight package. Not the end of the world, but thought I'd ask.

PR #459, opened 2025-12-17
JP-4209: Fix compute_axes comparison to enforce jump_data.max_extended_radius correctly

pllim:
Resolves [JP-4209](https://jira.stsci.edu/browse/JP-4209)

This PR addresses https://github.com/spacetelescope/jwst/issues/10090

## Tasks

  - [x] run regression tests with this branch installed (`"stcal@git+https://github.com/pllim/stcal@full-and-a-half-width"`)
    - [x] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/20313673878 (all passed)
    - [x] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml) https://github.com/spacetelescope/RegressionTests/actions/runs/20313700698 (all passed)

### DIFF
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -118,7 +118,7 @@ def detect_jumps_data(jump_data):
 def twopoint_diff_multi(jump_data, twopt_params, data, gdq, readnoise_2d, n_slices):
     """
     Split data for jump detection multiprocessing.
-    
+
     Parameters
     ----------
     jump_data : JumpData
@@ -130,7 +130,7 @@ def twopoint_diff_multi(jump_data, twopt_params, data, gdq, readnoise_2d, n_slic
     data : ndarray
         The science data, 4D array float.
 
-    gdq : ndarray 
+    gdq : ndarray
         The group DQ, 4D array uint8.
 
     readnoise_2d : ndarray
@@ -236,7 +236,7 @@ def slice_data(twopt_params, data, gdq, readnoise_2d, n_slices):
     data : ndarray
         The science data, 4D array float.
 
-    gdq : ndarray 
+    gdq : ndarray
         The group DQ, 4D array uint8.
 
     readnoise_2d : ndarray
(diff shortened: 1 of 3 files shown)


src/stcal/jump/jump.py
```diff
@@ -968,10 +970,10 @@ def compute_axes(expand_by_ratio, ellipse, expansion, jump_data):
     else:
         axis1 = ellipse[1][0] + expansion
         axis2 = ellipse[1][1] + expansion
-    axis1 = min(axis1, jump_data.max_extended_radius)
-    axis2 = min(axis2, jump_data.max_extended_radius)
+    axis1 = min(axis1 / 2, jump_data.max_extended_radius)
+    axis2 = min(axis2 / 2, jump_data.max_extended_radius)
 
-    return (round(axis1 / 2), round(axis2 / 2))
```
braingram: Would you add a unit test verifying this fix?

pllim: Added `test_ellipse_compute_axes_max_radius` in the test module. Without this patch, it fails as follows:

```
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 1 / 2 (50%)
E           Max absolute difference among violations: 2
E           Max relative difference among violations: 0.5
E            ACTUAL: array([2, 2])
E            DESIRED: array([4, 2])
```

This is because it halved `max_extended_radius` on return, treating it as diameter instead.

melanieclarke:
> Is there a corresponding jwst PR to fix the usage in jwst? jwst multiplies the spec-provided `max_extended_radius` by 2 (converting it to a "diameter"):

Thanks for pointing this out @braingram.  I wasn't tracking this issue closely, but looking at it now, I agree that it looks like the stcal code is self-consistently using this value as a full width, not a radius.  To reduce confusion, I think a better fix would be to just change the name of the JumpData attribute instead of changing how the value is used everywhere.  `max_extended_width` would work.

melanieclarke:
> > max_extended_width
> 
> Is this a public API change on both stcal and jwst with a deprecation period? 🤪

No need for a change on the jwst side.  The step parameter can still be specified as a radius and multiplied by 2 before being passed to JumpData.  I think the change of attribute name is an internal implementation detail for stcal, and should not need a deprecation period.  It is set internally via the `set_shower_info` method.

pllim:
Okay. Since there are conflicts anyway and the proposed solution is completely, different, I will just close this one without merge and open a new PR. Thanks, all.

PR #462, opened 2025-12-30
Align style checks with jwst

braingram:
Updates ruff and pre-commit checks to more closely match those in jwst:
- https://github.com/spacetelescope/jwst/blob/16981d163e3eb2546d48625dd4206cfe31d20a5f/.ruff.toml
- https://github.com/spacetelescope/jwst/blob/16981d163e3eb2546d48625dd4206cfe31d20a5f/.pre-commit-config.yaml

There are a few remaining checks that jwst runs that stcal does not:
- numpydoc: enabling this will require many manual documentation updates. If we want to do this I suggest we do so in a separate PR after we've clarified what is "public" in this package.
- stcal uses 110 line length, jwst uses 100. I don't think we should make this change as part of this PR. It introduces another ~3k line changes just from automatic fixes.

There are no API changes that impact downstream in this PR. I made updates to docstrings etc where obvious and added `noqa` where:
- I was unsure about docstring changes
- the rule violation would change API (for example renaming an argument)

If we go forward with this PR I will open up follow issues for rule violations that were `noqa`ed that we may want to address.

JWST regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/20603739253
romancal regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/20603745181
all pass with no differences

## Tasks


zacharyburnett (approved):
LGTM, good idea to separate shared and non-share checks in the config

pyproject.toml
```diff
@@ -92,77 +92,92 @@ markers = ["soctests"]
...
 ignore = [
+  "D100",   # missing docstring in public module
+  "D105",   # missing docstring in magic method
+  "E741",   # ambiguous variable name (O/0, l/I, etc.)
+  "UP015",  # unnecessary open(file, "r"). no harm being explicit
+  "TRY003", # prevents custom exception messages not defined in exception itself.
+  "TRY400", # enforces log.exception instead of log.error in except clause.
+
+  # below ignores are not shared with jwst
+  "B905",    # zip without strict
```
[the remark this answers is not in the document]
braingram: I agree that this one is worth a follow-up PR. Sorting out if `strict=True` changes behavior will require a more involved look than what was done as part of this style PR.

src/stcal/alignment/util.py
```diff
@@ -323,7 +327,7 @@ def wcs_from_footprints(
     ref_wcs: gwcs.wcs.WCS,
     ref_wcsinfo: dict,
     transform: astropy.modeling.models.Model | None = None,
-    bounding_box: Sequence | None = None,
+    bounding_box: Sequence | None = None,  # noqa: ARG001, ARG002
```
braingram: Follow up issue: https://github.com/spacetelescope/stcal/issues/464 we should probably just remove this function.

src/stcal/jump/jump.py
```diff
@@ -163,7 +165,7 @@ def twopoint_diff_multi(jump_data, twopt_params, data, gdq, readnoise_2d, n_slic
     return reassemble_sliced_data(real_result, jump_data, gdq, yinc)
 
 
-def reassemble_sliced_data(real_result, jump_data, gdq, yinc):
+def reassemble_sliced_data(real_result, jump_data, gdq, yinc):  # noqa: ARG001, ARG002
```
braingram: Follow-up issue: https://github.com/spacetelescope/stcal/issues/465

src/stcal/ramp_fitting/utils.py
```diff
@@ -72,10 +71,11 @@ def set_if_total_integ(final_dq, integ_dq, flag, set_flag):
     final_dq[all_set] = np.bitwise_or(final_dq[all_set], set_flag)
 
 
-def dq_compress_sect(ramp_data, num_int, gdq_sect, pixeldq_sect):
+def dq_compress_sect(ramp_data, num_int, gdq_sect, pixeldq_sect):  # noqa: ARG001
```
braingram: Follow-up issue: https://github.com/spacetelescope/stcal/issues/467

src/stcal/resample/resample.py
```diff
@@ -432,53 +427,41 @@ def check_output_wcs(self, output_wcs, estimate_output_shape=True):
...
             A dictionary of data model attributes and values.
 
         """
-        assert self._output_wcs is not None
-        assert np.array_equiv(
-            self._output_wcs.array_shape,
-            self._output_array_shape
-        )
-        assert self._output_pixel_scale
+        assert self._output_wcs is not None  # noqa: S101
```
braingram: Follow-up issue: https://github.com/spacetelescope/stcal/issues/469

PR #471, opened 2025-12-31
explicitly define `zip(strict)`

### LINKED ISSUE #463: Use `strict=True` for zip and unskip B905

zacharyburnett:
Closes #463 

define `zip(strict=True)` for instances of `zip()` without `strict`; depending on the usages and inputs, sometimes it might need to be `False`

## Tasks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -159,3 +159,6 @@ src/stcal/_version.py
 
 # VSCode stuff
 .vscode
+
+# lockfiles
+uv.lock
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,7 +144,6 @@ ignore = [
   "TRY400", # enforces log.exception instead of log.error in except clause.
 
   # below ignores are not shared with jwst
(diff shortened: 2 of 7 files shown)


uv.lock
braingram: Would you add `uv.lock` to `.gitignore`? If we never want this file in a repo maybe it's worth adding it to your global gitignore (if you have one)?

braingram:
Thanks! I looked at the uses and these all seem reasonable given the arguments provided to `zip`.

Would you run regtests with this PR?

zacharyburnett:
regtests:
- https://github.com/spacetelescope/RegressionTests/actions/runs/20999118539
- https://github.com/spacetelescope/RegressionTests/actions/runs/20999123554

braingram (approved):
Thanks! Regtests all passed (I had to restart the jwst ones because they failed due to a worker going down).

schlafly (approved):

PR #483, opened 2026-01-06
remove all resetwarnings

### LINKED ISSUE #477: find_faint_extended changes global warning filters
https://github.com/spacetelescope/stcal/blob/17f997245912ea72c5c23b0025e3b1f3f88517c6/src/stcal/jump/jump.py#L811

The ignore should have been a context manager with only relevant code within. As it stands, theoretically, any warnings that come after that line is executed would be silenced. Such behavior could hide actual warnings outside of that function that we might care about.

Also see:

* #150

braingram:
Removes all calls to `warnings.resetwarnings` and replaces them with `warnings.catch_warnings` to avoid global changes to the warning filters.

roman regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/20750781061
(failures unrelated)
jwst regtests: https://github.com/spacetelescope/RegressionTests/actions/runs/20750785751
all passed

Closes https://github.com/spacetelescope/stcal/issues/477

## Tasks

### DIFF
--- a/changes/483.bugfix.rst
+++ b/changes/483.bugfix.rst
@@ -1,0 +1,1 @@
+Replace ``warnings.resetwarnings`` with ``warnings.catch_warnings`` to avoid global changes to warning filters.
--- a/src/stcal/jump/jump.py
+++ b/src/stcal/jump/jump.py
@@ -822,35 +822,34 @@ def find_faint_extended(indata, ingdq, pdq, readnoise_2d, jump_data, min_diffs_f
 
     all_ellipses = []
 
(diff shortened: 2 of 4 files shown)


pllim (approved):
Numpy RuntimeWarning can also be controlled via https://numpy.org/doc/stable/reference/generated/numpy.errstate.html but this is good enough. Thanks!

emolter (approved):
LGTM

melanieclarke (approved):
Looks good to me too.  Thanks for the clean up!

PR #485, opened 2026-01-06
Deprecate unused functions

### LINKED ISSUE #347: Possible duplicate reproject() function
Is there any meaningful difference between `stcal.alignment.util.reproject()` and `stcal.outlier_detection.utils.reproject()`?  Could the one in outlier detection be removed?  @mcara @nden @braingram

t-brandt:
Resolves [JP-4216](https://jira.stsci.edu/browse/JP-4216)

Closes #347 

This PR deprecates unused functions in `stcal.alignment` and `stcal.outlier_detection`.  For computing a pixel transformation map, `drizzle.utils.calc_pixmap` is now used, matching the behavior in `stcal.resample`.  Functions `reproject` should now be unused either in `stcal`, `jwst`, or `romancal` and have been deprecated.

## Tasks

### DIFF
--- a/changes/485.removal.rst
+++ b/changes/485.removal.rst
@@ -1,0 +1,1 @@
+Deprecate ``reproject`` from both ``alignment.util`` and ``outlier_detection.utils``, deprecate ``calc_pixmap`` from ``alignment.util`` and ``calc_gwcs_pixmap`` from ``outlier_detection.utils`` in favor of ``drizzle.utils.calc_pixmap``
--- a/changes/485_removal.rst
+++ b/changes/485_removal.rst
@@ -1,1 +1,0 @@
-Deprecate ``reproject`` from both ``alignment.util`` and ``outlier_detection.utils``, deprecate ``calc_pixmap`` from ``alignment.util`` and ``calc_gwcs_pixmap`` from ``outlier_detection.utils`` in favor of ``drizzle.utils.calc_pixmap``
--- a/changes/485_removal.rst
+++ b/changes/485_removal.rst
@@ -1,0 +1,1 @@
+Deprecate ``reproject`` from both ``alignment.util`` and ``outlier_detection.utils``, deprecate ``calc_pixmap`` from ``alignment.util`` and ``calc_gwcs_pixmap`` from ``outlier_detection.utils`` in favor of ``drizzle.utils.calc_pixmap``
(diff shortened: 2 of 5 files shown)


braingram:
Renaming the changelog fragment to 485.removal.rst should fix the changelog failures (which is causing docs build to fail).

I wouldn't object to removing these functions instead of deprecating them since stcal essentially has no stable API. 

If the warnings are kept would you add an appropriate stack level to fix the style failures?

There are downstream failures that appear to be from gwcs_blot usage in romancal and jwst triggering the deprecation warning. Either way would you update gwcs_blot to use non-deprecated functions?

emolter (commented):
This looks good to me!  I'm holding off on approving until regression tests are run for jwst.  I've gone ahead and started them here: https://github.com/spacetelescope/RegressionTests/actions/runs/20785713860

t-brandt:
@mcara right now, stcal uses both drizzle's calc_pixmap:
https://github.com/spacetelescope/stcal/blob/ee4fb8501ddcbce92aaceb660545efef4f12f815/src/stcal/resample/resample.py#L9
https://github.com/spacetelescope/stcal/blob/ee4fb8501ddcbce92aaceb660545efef4f12f815/src/stcal/resample/resample.py#L757
and its own version of the same thing:
https://github.com/spacetelescope/stcal/blob/ee4fb8501ddcbce92aaceb660545efef4f12f815/src/stcal/outlier_detection/utils.py#L281
https://github.com/spacetelescope/stcal/blob/ee4fb8501ddcbce92aaceb660545efef4f12f815/src/stcal/outlier_detection/utils.py#L254
This PR is meant to clean that up and to avoid having the two routines (in stcal and in drizzle) diverge.

Implementing the ability to interpolate in both outlier detection and drizzling would match the behavior of drizzlepac and would speed up the mosaic pipeline in romancal by nearly a factor of 3.  We can't afford to leave that on the table.  I am not proposing changing the default behavior of drizzle to match that of drizzlepac, but only exposing options to the user to have it behave differently under the appropriate circumstances, i.e., to do what drizzlepac does when the user wants it to.  If we don't change drizzle then we will need to either repeat the calc_pixmap functions entirely within stcal to avoid any references to them in drizzle, keeping multiple versions of the same routines, or we will need to repeat them yet again in romancal.  Both of these seem worse to me from the standpoint of maintainability and ease of doing what we want.

t-brandt:
Withdrawing this PR in favor of https://github.com/spacetelescope/stcal/pull/488

PR #486, opened 2026-01-08
add `Breaking Changes` change log category and rename `General` to `Other Changes`

### LINKED ISSUE #287: should `stcal` use steps as the top-level sections in the change log, like `jwst` does?
The manually-made changelog before #284 had `General`, `Bug Fixes`, and `Changes to API`. as top-level headers, and then steps as subheaders underneath that:
https://github.com/spacetelescope/stcal/blob/c1a8bab38cdfd5221c74df870000183921127e3b/CHANGES.rst?plain=1#L15-L46

However, `towncrier` only supports creating one level of header, so in #284 I kept the top-level ones and recommended prepending `[ramp_fitting]`, `[jump]`, etc. to the front of the change log entry to make it clearer which entry affects which step:
https://github.com/spacetelescope/stcal/blob/b9b0f8dac9707da71b242fc7e8954e05393ac447/changes/278.apichange.rst?plain=1#L1

Should we change the change log sections (fragment types) to be more like `jwst`'s change log (steps as top-level headers)? 

This would change the fragment filename from `changes/299.bugfix.rst` to `changes/299.ramp_fitting.rst`, and involve changing these types to every possible step:
https://github.com/spacetelescope/stcal/blob/c1a8bab38cdfd5221c74df870000183921127e3b/pyproject.toml#L261-L268

zacharyburnett:
Closes #287 

- https://github.com/spacetelescope/jwst/pull/10119
- https://github.com/spacetelescope/romancal/pull/2129

## description

- explicitly set order of change log categories instead of using alphabetical order (by using TOML arrays)
- defines a `Breaking Changes` change log category to the top of the change log, along with instructions in the `changes/README.rst`. The idea here is that if a PR change breaks existing API or functionality, the dev should ALSO add a `changes/<PR#>.breaking.rst` news fragment that summarizes the breaking change and how to migrate to the new functionality / API
- changes `General` category to `Other Changes`
- defines changelog categories for every step

## Tasks


pyproject.toml
```diff
@@ -241,14 +241,67 @@ ignore = [".gitkeep"]
...
+[[tool.towncrier.type]]
+directory = "docs"
+name = "Documentation"
+showcontent = true
 
-[tool.towncrier.fragment.general]
-name = "General"
+[[tool.towncrier.type]]
+directory = "misc"
+name = "Micellaneous"
```
tapastro: > Miscellaneous

emolter (approved):
I'm not 100% familiar with the toml specifics, I'm assuming this works from a technical standpoint.

This looks good to me, it appears to achieve what we agreed on in the meeting RE breaking changes.  Reorganizing by step is a nice bonus.

melanieclarke (commented):
#483 just added in a new change that needs to be renamed, otherwise I think this looks good.  Let me know when that's done so we can get this in before more changes arrive!

zacharyburnett:
all done, I'll merge this now before more changes are merged

PR #491, opened 2026-01-13
Added snowball flagging to stcal likely_fit

t-brandt:
Resolves [RCAL-1312](https://jira.stsci.edu/browse/RCAL-1312)

Closes #

This PR adds the snowball-flagging capability to the likelihood fitting algorithm.  This enables the user to reproduce the important jump detection features entirely within the likelihood-based ramp fitting/jump detection step.

## Tasks


zacharyburnett (approved):
this looks good to me as long as the new test passes; perhaps we could also add a regression test fitting against a larger dataset, but I'll leave that up to you

Would this be considered a breaking change if it alters the existing behavior of the pipeline with `jump.expand_large_events`? Maybe not?

t-brandt:
I slightly expanded the unit test to verify that the expanded jump flags were being applied correctly when ramp fitting.

schlafly (approved):
Thanks Tim!

melanieclarke (approved):
Looks good!  No impact to jwst.

PR #497, opened 2026-01-20
[SCSB-234] use Python 3.13 to build docs with Sphinx 9

zacharyburnett:
Resolves [SCSB-234](https://jira.stsci.edu/browse/SCSB-234)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,7 +53,6 @@ extensions = [
     "sphinx_automodapi.automodsumm",
     "sphinx_automodapi.autodoc_enhancements",
     "sphinx_automodapi.smart_resolver",
-    "sphinx_asdf",
     "sphinx.ext.mathjax",
 ]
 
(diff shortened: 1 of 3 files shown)


zacharyburnett:
1 warning treated as error blocking the build:
https://app.readthedocs.org/projects/stcal/builds/31243216/#304492810--184
```
generating indices... <unknown>:1: WARNING: py:class reference target not found: numpy._typing.DTypeLike [ref.class]
```

braingram:
Would it be too far of a scope stretch to also drop the unused `sphinx-asdf` as part of this PR? by removing:
https://github.com/spacetelescope/stcal/blob/56a3f89c5a4d13f94107c66df536e4062a9fbf5f/docs/conf.py#L56
https://github.com/spacetelescope/stcal/blob/56a3f89c5a4d13f94107c66df536e4062a9fbf5f/pyproject.toml#L36

PR #500, opened 2026-01-21
pin gwcs<1

braingram:
Will also require pinning in jwst: https://github.com/spacetelescope/jwst/pull/10164

## Tasks


melanieclarke (dismissed):
Thanks @braingram, this seems like a reasonable first step to get the tests passing.  Do we know where the error is coming from?

braingram:
> The tests with the jwst PR should be sufficient, thanks! Have you run romancal regtests?

Thanks.

I queued them up https://github.com/spacetelescope/RegressionTests/actions/runs/21224251373

FWIW romancal currently has a gwcs upper pin that doesn't allow gwcs 1:
https://github.com/spacetelescope/romancal/blob/6494d21ab34cf90aaff9a16d0d058ec71e589b50/pyproject.toml#L35

WilliamJamieson:
Note spacetelescope/gwcs#684 will fix the issue this is solving. I will publish a bugfix version when that PR is merged.

melanieclarke (approved):
Regtests all passing; thanks for the fix.

PR #506, opened 2026-01-28
Handle group averaging in non-linearity correction

schlafly:
This PR significantly refactors and updates the linearity correction step in order to allow for handling of the fact that linearity corrections are derived using data where every read is available, but are applied to data which average the reads into resultants.

Following the approach of @t-brandt , this PR makes an initial linearity correction in the traditional way to make a simple estimate for the count rate of each pixel.  It then uses that to simulate the individual reads entering into each group, and applies an inverse linearity polynomial to see how different that simulated group was from the observed group.  It applies the difference to the simulated reads, leaving one with a set of individual reads that average to the observed group and also increase with count rate in the correct way.  It then applies the linearity correction to these reads and averages the result to get the final linearity-corrected group.

This approach requires additional arguments.  Critically, it requires inverse linearity polynomials and the read pattern, so that it knows how to make linearity-affected reads and groups from those linearity affected reads.  Optionally, it also accepts the saturation level.  If provided, it caps individual reads at 1.2x saturation.  This is not expected to see much use since resultants marked as saturated are also restored to their original pre-linearity-correction values, but it feels appropriate to avoid calling the linearity / inverse linearity functions much outside their ranges.

Regression tests with romancal and jwst show no changes, since these pipelines don't provide the arguments necessary to trigger the new behavior.
https://github.com/spacetelescope/RegressionTests/actions/runs/21636507600
https://github.com/spacetelescope/RegressionTests/actions/runs/21636512363
For Roman we expect to turn on the new behavior soon.

This PR also provides a hook to allow a pipeline to include an "additional correction" beyond the classical non-linearity polynomials.  This correction is applied to the individual reads that are simulated as part of this PR.  This feature will be used to support the integral non-linearity correction that will be done for Roman.

I changed more than I originally wanted to as part of this PR to try to remove some code duplication; I apologize for the long diff.  The biggest change there was to remove the separate zero frame handling in the linearity correction branch, and to not use separate linearity coefficients and linearity dqs for the zero frame processing, which didn't make sense to me since the linearity corrections were the same (and does not seem to impact regression tests).

There's a new "prepare_coefficients" function which handles changing problematic coefficients to no-ops that was previously part of linearity_correction_branch.  linear_correct_plane ended up being renamed apply_polynomial and having many of its arguments made optional.

The new tests show that the new linearity correction gives a more accurate result for a particular set of linearity / inverse linearity polynomial coefficients that I made up.

## Tasks

- [X] update or add relevant tests
- [X] Does this PR change any API used downstream? (if not, label with `no-changelog-entry-needed`)
  - [X] run regression tests with this branch installed (`"git+https://github.com/<fork>/stcal@<branch>"`)
    - [X] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/runs/21448690317)
    - [X] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/runs/21448669635)


src/stcal/linearity/linearity.py
```diff
@@ -344,3 +307,238 @@ def ben_coeffs(lin_coeffs):
...
+
+    Returns
+    -------
+    data_linearized : ndarray
+        The corrected data (ngroups, nrows, ncols).
+    """
+    ngroups, nrows, ncols = data.shape
+
+    # If no inverse linearity coefficients, do simple correction in place
+    if ilin_coeffs is None:
```
tapastro: I had a discussion with @t-brandt about this work, and he suggested that while JWST is unlikely to provide the integral nonlinearity corrections, we do have an existing bug in our nonlinearity corrections for group-averaged data; we perform the correction on the averaged data, which introduces a bias. 

If I understand this PR (and that discussion), this PR's approach would not allow JWST to utilize the correct method for group-averaged data. Is that right? If so, could this be re-written in some way to allow group-averaged data to be corrected on a per-read basis, then averaged back into the original shape, all without ilin_coeffs provided? I don't recall the details of how this would work - perhaps performing an initial coeffs determination from the group-averaged data, then using that as a first guess for generating the per-read counts to re-derive a better set of coeffs? I could very well be speaking out of my rear, I am not that familiar with the details.

src/stcal/linearity/linearity.py
```diff
@@ -344,3 +307,238 @@ def ben_coeffs(lin_coeffs):
...
+
+    # Total number of reads
+    nreads = max([max(reads) for reads in read_pattern])
+
+    # Calculate read times
+    readtimes = np.arange(nreads, dtype=float)
+
+    # Calculate resultant times (mean time of reads in each resultant)
+    t_resultants = np.array([np.mean([readtimes[r - 1] for r in reads]) for reads in read_pattern])
+
```
t-brandt:
```suggestion
    # Calculate mean read number of each resultant (e.g. reads [4, 5, 6, 7] is 5.5)
    mean_read_resultant = np.array([np.mean(reads) for reads in read_pattern])

```

tapastro (approved):
I think this looks good - I'm not too worried about API changes to linearity methods, and this approach allows JWST to try out the group-averaged linearity changes.

src/stcal/linearity/linearity.py
```diff
@@ -344,3 +307,238 @@ def ben_coeffs(lin_coeffs):
...
+    iy, ix = np.meshgrid(np.arange(nrows), np.arange(ncols), indexing="ij")
+
+    # Create index tuple for last valid resultant of each pixel
+    last_idx = (lastvalidresultant[iy, ix], iy, ix)
+
+    # Linearize last valid resultant for each pixel
+    lastread_lin = apply_polynomial(data[last_idx], lin_coeffs, gdq[last_idx], dqflags)
+
+    dt = t_resultants[last_idx[0]] - t_resultants[0]
+    countrate = (lastread_lin - firstread_lin) / dt
```
t-brandt:
```suggestion
    d_reads = mean_read_resultant[last_idx[0]] - mean_read_resultant[0]
    # countrate is in units of counts/read
    countrate = (lastread_lin - firstread_lin) / d_reads
```

src/stcal/linearity/linearity.py
```diff
@@ -344,3 +307,238 @@ def ben_coeffs(lin_coeffs):
...
+    for i in range(ngroups):
+        reads_for_resultant = read_pattern[i]
+        n_reads_this = len(reads_for_resultant)
+
+        # Allocate array for just these reads
+        reads_linearized = np.zeros((n_reads_this, nrows, ncols), dtype=data.dtype)
+
+        # Reconstruct linearized counts for these reads
+        for j, read_idx in enumerate(reads_for_resultant):
+            reads_linearized[j] = firstread_lin + countrate * readtimes[read_idx - 1]
```
t-brandt:
```suggestion
        reads_this_resultant = np.array(read_pattern[i])

        # Reconstruct linearized counts for these reads
        reads_linearized = firstread_lin + countrate * reads_this_resultant[:, None, None]
```

src/stcal/linearity/linearity.py
```diff
@@ -344,3 +307,238 @@ def ben_coeffs(lin_coeffs):
...
+        if satval is not None:
+            np.putmask(
+                reads_unlinearized,
+                reads_unlinearized > 1.2 * satval[np.newaxis, :, :],
+                1.2 * satval[np.newaxis, :, :],
+            )
+
+        # Apply classic linearity to the reads
+        reads_corrected = np.zeros_like(reads_unlinearized)
+        for j in range(n_reads_this):
```
t-brandt:
```suggestion
        for j in range(reads_corrected.shape[0]):
```

schlafly:
@t-brandt , I think I addressed all of your comments.  I reran regression tests, which are still green.  If you're happy, I'll merge this.
https://github.com/spacetelescope/RegressionTests/actions/runs/21636507600
https://github.com/spacetelescope/RegressionTests/actions/runs/21636512363

t-brandt (approved):

PR #516, opened 2026-02-05
AL-991: Add support for DQ flag propagation in resample

mcara:
Resolves [AL-991](https://jira.stsci.edu/browse/AL-991)

This PR adds support for propagating DQ flags in resample. It is off by default so it will not affect JWST.

## Tasks

### DIFF
--- a/changes/516.resample.rst
+++ b/changes/516.resample.rst
@@ -1,0 +1,1 @@
+Added support for propagating DQ flags in resampling. DQ flags are propagated by bitwise OR of all input DQ flags that contribute to a given output pixel.
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -92,6 +92,7 @@ class Resample:
         enable_ctx=True,
         enable_var=True,
         compute_err=None,
+        propagate_dq=False,
     ):
         """
         Initialize Resample.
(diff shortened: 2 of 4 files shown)


schlafly (approved):
Thanks Mihai.  This looks good to me.

tapastro (approved):
I don't have any real issue with the code, but I don't know that an OR-joined resampled DQ array is going have much value for JWST - I expect most of the DQ planes will be either nearly full or nearly empty.

I imagine the drizzle method is optimized for OR-ing bits and would not be immediately compatible with a more nuanced approach; that said, my first thought would be to AND all input pixels that meet some overlap threshold, i.e. >=0.5. That is likely going to be much slower than whatever has been written...

schlafly:
> I don't have any real issue with the code, but I don't know that an OR-joined resampled DQ array is going have much value for JWST - I expect most of the DQ planes will be either nearly full or nearly empty.

While many planes are not valuable, e.g., I expect people will be interested in knowing whether any partially saturated pixels contributed to a target pixel.  For Roman we'll also be marking the locations of diffraction spikes where this will be valuable (and presumably would have value in Webb as well).  I agree that we'll have to mask out the planes we don't care about when using this feature.

mcara:
Please merge at will

PR #520, opened 2026-02-11
Fix ambiguous behavior for identical chi squared improvements

t-brandt:
Resolves [JP-4245](https://jira.stsci.edu/browse/JP-4245)

Closes #

This PR fixes ambiguous and unpredictable behavior when operating on noiseless data (as in tests), where the chi squared improvements could be identical when dropping one and two group differences because the extra difference was a perfect fit to a line.  The code assumed that a chi squared improvement being identical to a trial value was sufficient, but this adds an additional check to ensure that the correct number of group differences is masked.

## Tasks


schlafly (approved):
This resolves the issue in romancal---thank you!

braingram:
> This looks good to me.
> 
> Regtests: romnacal: https://github.com/spacetelescope/RegressionTests/actions/runs/21962510669/job/63443796177 jwst: https://github.com/spacetelescope/RegressionTests/actions/runs/21962499466/job/63443780897

Both show differences that would need okifying. @t-brandt are these differences expected? Is the jwst data noiseless?

t-brandt:
Those JWST data are not noiseless, but they are 16-bit unsigned integer at the outset.  Looking at it a little more closely it does seem plausible that a very, very small fraction of pixels would hit this edge case.

tapastro (approved):

PR #526, opened 2026-02-25
JP-4097: Allow 4-D reference file arrays to support multistripe data

tapastro:
Partially resolves [JP-4097](https://jira.stsci.edu/browse/JP-4097)

This PR makes changes to level 1 step algorithms to allow 4-dimensional reference file arrays where 2-dimensional arrays were assumed up to this point. This is meant to solve issues caused by multistripe data from JWST, where distinct, possibly disjoint regions of detector space are read separately, such that each integration will not be situated at the same location in the detector frame. A 2-d reference array could be used, but the complex logic needed to reproduce the multistripe detector readout pattern would need to be replicated in stcal code, which I wanted to avoid.

The simplest solution I could find was to build a 4-d reference array in JWST step code which should be useable in a 1:1 fashion to the science data array. This required modifying some steps which to this point could assume that the reference array would always be of a shape representing the two-dimensional detector array.

## Tasks

### DIFF
--- a/changes/526.saturation.rst
+++ b/changes/526.saturation.rst
@@ -1,0 +1,1 @@
+Allow 4D input arrays in the saturation step, matching the dimensions of the input data. This is intended for support of striping modes that may cover different areas of the detector in each integration.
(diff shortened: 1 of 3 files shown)


melanieclarke (commented):
This looks reasonable, for supporting superstripe mode. I agree it's preferable to including any more complicated logic for reference file extraction here in stcal.

kmacdonald-stsci (dismissed):
It is not clear to me why the various arrays could have different dimensions. Please update the docstrings to clarify that arrays could be different dimensions and under what conditions.

schlafly (approved):
Thanks!

PR #531, opened 2026-03-12
Reduce memory usage of linearity correction when read-level correction is used

schlafly:
This PR reduces the memory usage of the linearity step, in particular when using the new read-level correction for Roman.  The results are numerically identical to the old results, as demonstrated by the Webb and Roman regression tests:
Webb: https://github.com/spacetelescope/RegressionTests/actions/runs/23003338371
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/23003355163
Note that the Webb tests show failures, but they are the same failures as on main and relate to not finding particular MIRI files rather than to numerical mismatches.  

The main change is that instead of doing the entire image at once in the read-level correction we chunk it into groups of 64 rows and do each chunk of rows individually.  This naturally reduces the peak memory usage dramatically.

As part of my clean up I did also fill out the docstrings in some unrelated linearity routines and delete a couple unused lines.  I also slightly rewrote the core 'apply_polynomial' function and some of the read-level correction to remove some unneeded temporary copies of arrays.

I did not update any tests since everything was covered by existing tests.

For romancal, for one particular default MA table used in our regression tests this is the most expensive step, using 12 GB resident / 9 GB heap.  With this PR, those numbers become 7 GB resident / 3 GB heap.

## Tasks

- [X] run regression tests with this branch installed (`"git+https://github.com/<fork>/stcal@<branch>"`)
  - [X] [`jwst` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/jwst.yml)
  - [X] [`romancal` regression test](https://github.com/spacetelescope/RegressionTests/actions/workflows/romancal.yml)

### DIFF
--- a/changes/531.linearity.rst
+++ b/changes/531.linearity.rst
@@ -1,0 +1,1 @@
+Reduce memory usage of linearity correction, particularly for read-level corrections.
--- a/src/stcal/linearity/linearity.py
+++ b/src/stcal/linearity/linearity.py
@@ -71,8 +71,8 @@ def linearity_correction(
         A callable that takes the non-linear counts array and returns a correction
         array to be added to it. This allows mission-specific corrections (e.g.,
         integral non-linearity) to be applied. The callable should accept a 3D
-        array (nreads, nrows, ncols) and return a correction array of the same
-        shape.
+        array (nreads, tile_rows, ncols) and return a correction array of the same
+        shape.  The last axis must span the complete detector width.
 
     read_pattern : list of lists or None
         The pattern of reads entering into groups. Each element is a list of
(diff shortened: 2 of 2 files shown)


zacharyburnett (commented):
I don't have the expertise to approve this but did have a question, does this open up the possibility of scaling linearity correction with multiprocessing?

braingram (approved):
The regtest runners were not picking up the dependency override (due to a workflow bug).
https://github.com/spacetelescope/RegressionTests/actions/runs/23156755721
is a run with the bug fixed and shows expected memory reductions for linearity and elp tests.

Thanks for the updates.

tapastro (approved):
No objections, LGTM.

PR #421, opened 2025-10-24
JP-3967: SATURATION flag for partially saturated pixels should be propagated to the _rate DQ

pllim:
Resolves [JP-3967](https://jira.stsci.edu/browse/JP-3967)

This PR addresses https://github.com/spacetelescope/jwst/issues/8124 by reverting a bit of #125 

xref

* https://github.com/spacetelescope/stcal/pull/156

Target build: 12.3

## Notes from Ken MacDonald

There are several past PRs related to your ticket:

* https://github.com/spacetelescope/stcal/pull/133
* https://github.com/spacetelescope/stcal/pull/131
* https://github.com/spacetelescope/stcal/pull/125
* https://github.com/spacetelescope/stcal/pull/112

In particular, the SAT flag for an integration was set only if the entire integration was set. This has a couple of consequences. When SAT is set, so is DNU. This invalidates the data, making their values NaN.

Unfortunately, due to the "LIKELY" algorithm being done in Python, this is done in two parts of the code:

* `utils.py` for the “LIKELY” and
* `slope_fitter.c` for the “OLS” algorithm.

Below are links to the relevant parts of the C extension.

The initial set up for SAT flagging is done here (done on intial read):

https://github.com/spacetelescope/stcal/blob/b3263575bdf1181d2461cba0c465dc32d43a5fd0/src/stcal/ramp_fitting/src/slope_fitter.c#L1891

Notice there is a `cnt_sat` variable that gets updated. This makes it easy to fine fully saturated ramps.

The setting of the integration level SAT is done here (done during processing):

https://github.com/spacetelescope/stcal/blob/b3263575bdf1181d2461cba0c465dc32d43a5fd0/src/stcal/ramp_fitting/src/slope_fitter.c#L1295

Here is the seeting of pedestal data based on the integration level saturation:

https://github.com/spacetelescope/stcal/blob/b3263575bdf1181d2461cba0c465dc32d43a5fd0/src/stcal/ramp_fitting/src/slope_fitter.c#L1989

Invalid data is defined as `DNU | SAT`:

https://github.com/spacetelescope/stcal/blob/b3263575bdf1181d2461cba0c465dc32d43a5fd0/src/stcal/ramp_fitting/src/slope_fitter.c#L2148

If a ramp is flagged as SAT, set its slope to NaN:

https://github.com/spacetelescope/stcal/blob/b3263575bdf1181d2461cba0c465dc32d43a5fd0/src/stcal/ramp_fitting/src/slope_fitter.c#L2807

[...]

PR #538, opened 2026-04-15
JP-4000: Updating Ramp Fitting Based on new Usage of Persistence Flag

kmacdonald-stsci:
Resolves [JP-4000](https://jira.stsci.edu/browse/JP-4000)

This PR addresses the persistence flag handling in the ramp fit step. The new persistence step flags groups now, instead of at the pixel level. These group level flags interfered with ramp fit processing, but shouldn't have any effect on ramp fitting. This PR is designed to identify groups that have been flagged with persistence and ensure they do not effect ramp fitting. The effects were in the suppression of one group ramps and the computation of segments within a ramp.

## Tasks


kmacdonald-stsci:
Here is a clean regression test run, using this PR:

https://github.com/spacetelescope/RegressionTests/actions/runs/24677667046

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -287,10 +287,14 @@ def suppress_one_good_group_ramps(ramp_data):
...
     nints, ngroups, nrows, ncols = dq.shape
     dnu_flag = ramp_data.flags_do_not_use
+    pers_flag = np.uint8(ramp_data.flags_persistence)
+    npers = ~pers_flag
 
     for integ in range(nints):
         # In the current integration find ramps with only one group.
-        intdq = dq[integ, :, :, :]
+        intdq = dq[integ, :, :, :].copy()
+        intdq[:, :, :] &= npers
```
melanieclarke: Is this data copy necessary? I think you should be able to check for any flag other than persistence in the dq array directly, in the `good_index` logic below.

src/stcal/ramp_fitting/src/slope_fitter.c
melanieclarke, line 1: A general question -- since we expect that persistence flags are only ever informational and shouldn't impact the ramp fitting, could we pass in a DQ array with persistence flags removed, instead of modifying the slope fitter code to expect them?

I'm wondering in part because it seems possible we might want more information-only flags in the group DQ in the future, and it seems like a larger maintenance burden to update the C code every time.

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: This isn't about just checking for flags other than persistence. It is about finding ramps that have one good group, with respect to all flags except the persistence flag. The persistence flag is temporarily removed in order to sum the number of ramps with no flags, then check for the ramps where that sum is 1. Since the removal is temporary, a copy is used to prevent permanent removal persistence flag.

src/stcal/ramp_fitting/src/slope_fitter.c
kmacdonald-stsci: The group DQ array is made up of 8 bit `uchar`'s and all 8 bits are now defined. No other group DQ flag can be defined now. This won't have to be done again.

src/stcal/ramp_fitting/ramp_fit.py
melanieclarke: But can't you make that check without actually removing the flag?  Like `dq[integ, :, row, col] & npers == 0` instead of `dq[integ, :, row, col] == 0`?

src/stcal/ramp_fitting/src/slope_fitter.c
melanieclarke: Okay, but even so, it seems like a simpler change to handle it on the Python side instead of the C.  Did you explore that route?

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: This raises an ambiguity error, with and without the parenthesis.  And that is not the only place the copy is used.

Screenshot 2026-04-21 at 12 55 27 PM

melanieclarke: Sorry, I meant inside the np.where, line 313.

src/stcal/ramp_fitting/src/slope_fitter.c
kmacdonald-stsci: To do this an array of the persistence flags would have to be kept around to add them to the rate and rateints product after returning from the slope fitter. The persistence flag is easily accounted for in the slope fitter for both the segment computation and the propagation of flags to the return products.

melanieclarke: Fair enough!

src/stcal/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I consistently got feedback for years from both Nadia and Howard to avoid using `np.where`.

kmacdonald-stsci: Also, the copied array gets used in multiple places, as I explained above. This simplifies the code, makes it easier to read, and add modest temporary overhead.

tapastro:
In testing this, I find that persistence flags present in the input RampModel's groupdq array are not propagated to the output rate or rateints dq arrays. We want the same behavior here as for the saturated flag, as of the latest changes from https://github.com/spacetelescope/stcal/pull/421/ .

kmacdonald-stsci:
> In testing this, I find that persistence flags present in the input RampModel's groupdq array are not propagated to the output rate or rateints dq arrays. We want the same behavior here as for the saturated flag, as of the latest changes from #421 .

This has now been fixed.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -300,13 +304,14 @@ def suppress_one_good_group_ramps(ramp_data):
...
         wh1_cols = wh_one[1]
+
         for n in range(len(wh1_rows)):
             row = wh1_rows[n]
             col = wh1_cols[n]
             # Find ramps that have good 0th group, but the rest of the
             # ramp flagged.
-            good_index = np.where(ramp_data.groupdq[integ, :, row, col] == 0)
-            if ramp_data.groupdq[integ, good_index, row, col] == 0:
+            good_index = np.where(intdq[:, row, col] == 0)
```
tapastro: I understood Melanie's point above to be that intdq is not necessary, as it is only used in locations where it is checked against a value of 0. Replacing `good_index = np.where(intdq[:, row, col] == 0)` here with `good_index = np.where(intdq[:, row, col] & npers == 0)`, with the equivalent addition in the below line, removes the need for a copy of an array of up to 10x2048x2048 ints.

kmacdonald-stsci: I will try a couple of alternative ways to change the way the processing occurs to see if I can address this adequately.

kmacdonald-stsci: I added two functions that use smaller memory footprints. Both work. 

The difference would be performance. The `*_smallest` function has three explicit python loops, so will run slower, but the memory footprint is only the size of an individual pixel ramp, i.e., a `uint8` with dimension `(ngroups)`. The other function uses implicit numpy looping, so performs faster, but will have a memory foot print of `uint8` with dimensions `(ngroups, nrows, ncols)`. 

The latter is a trade off of memory vs performance, but if you want the smallest memory consumption, the `*_smallest` function works and will likely only be a modest performance hit. Let me know which you prefer implemented and I will remove the deprecated function, `*_dep` and the one not desired.

src/stcal/ramp_fitting/ramp_fit.py
```diff
@@ -273,12 +273,71 @@ def ramp_fit_data(ramp_data, save_opt, readnoise_2d, gain_2d, algorithm, weighti
...
+    good_groups = np.zeros((ngroups, nrows, ncols), dtype=ramp_data.groupdq.dtype)
+    for integ in range(nints):
+        dq = ramp_data.groupdq[integ, :, :, :]
+        good_groups[:, :, :] = 0
+        good_groups[(dq[:, :, :] & npers) == 0] = 1
+
+        wh_one = np.where(np.sum(good_groups, axis=0) == 1)
+        if len(wh_one[0]) > 0:
+            for idx in range(len(wh_one[0])):
+                dq[:, wh_one[0][idx], wh_one[1][idx]] |= dnu_flag
```
melanieclarke: This version looks good!  I think you could still go one step further and skip making the good_groups array too:

```suggestion
    for integ in range(nints):
        dq = ramp_data.groupdq[integ, :, :, :]
        wh_one = np.where(np.sum((dq[:, :, :] & npers) == 0, axis=0) == 1)
        if len(wh_one[0]) > 0:
            for idx in range(len(wh_one[0])):
                dq[:, wh_one[0][idx], wh_one[1][idx]] |= dnu_flag
```

kmacdonald-stsci: I updated this, but because I find one liners like this difficult to read and debug, I left the old code in a comment, along with text explaining what that line does.

tapastro (approved):

tapastro:
Romancal RTs here: https://github.com/spacetelescope/RegressionTests/actions/runs/27356252439

braingram (approved):
Approving for roman.

schlafly (approved):

PR #542, opened 2026-05-07
Update resampled output image's background level and status

mcara:
Resolves [AL-1057](https://jira.stsci.edu/browse/AL-1057)

When attempting to resample (again) an already resampled image, the resulting double-resampled image often has an unexplained background offset compared to expected (single-resampled) image. After analysis, this is caused by resample populating most meta of the output resampled image from an input cal image and not updating background-related fields.

## Tasks

### DIFF
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -506,6 +506,9 @@ class Resample:
             "start_time": None,
             "end_time": None,
             "duration": 0.0,
+            # background level
+            "level": 0.0,
+            "subtracted": True,
         }
 
         if self._enable_var:
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -507,8 +507,8 @@ class Resample:
             "end_time": None,
             "duration": 0.0,
             # background level
-            "level": 0.0,
-            "subtracted": True,
+            "level": None,
+            "subtracted": False,
         }
 
         if self._enable_var:
--- a/changes/542.resample.rst
+++ b/changes/542.resample.rst
@@ -1,0 +1,2 @@
+Corrected the "level" and "subtracted" keys in the output model dictionary to
+show correct values for the resampled image.
--- a/tests/resample/test_resample.py
+++ b/tests/resample/test_resample.py
@@ -95,6 +95,8 @@ def test_resample_mostly_defaults(weight_type):
 
     assert resample.output_model["pointings"] == nmodels
     assert resample.output_model["exposure_time"] == ttime
+    assert resample.output_model["level"] == 0.0
+    assert resample.output_model["subtracted"] is True
 
     # next assert assumes constant IVM
     assert np.allclose(
--- a/tests/resample/test_resample.py
+++ b/tests/resample/test_resample.py
@@ -95,8 +95,8 @@ def test_resample_mostly_defaults(weight_type):
 
     assert resample.output_model["pointings"] == nmodels
     assert resample.output_model["exposure_time"] == ttime
-    assert resample.output_model["level"] == 0.0
-    assert resample.output_model["subtracted"] is True
+    assert resample.output_model["level"] == None
+    assert resample.output_model["subtracted"] is False
 
     # next assert assumes constant IVM
     assert np.allclose(


mcara:
Running regression tests on this PR is not needed, IMO, but I can start them if so desired.

braingram:
This PR only has a changelog entry. @mcara is this missing local files? I'll convert it back to draft until:
- all files are added
- tests pass

zacharyburnett:
> This PR only has a changelog entry. @mcara is this missing local files? I'll convert it back to draft until:
> 
> * all files are added
> * tests pass

the perfect PR! Can't introduce any bugs if there's no new code!

mcara:
@zacharyburnett This is a documentation‑driven development PR. Code? No. But the story is amazing.

braingram:
> Running regression tests on this PR is not needed, IMO, but I can start them if so desired.

Thanks for updating this. Please do run the regression tests for jwst and romancal when this is ready.

Would you add a test for this fix?

mcara:
`JWST` regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/25517766433
`romancal` regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/25517815534

mcara:
> Thanks for updating this. Please do run the regression tests for jwst and romancal when this is ready.
> 
> Would you add a test for this fix?

Done (both)

emolter (approved):
I tested locally that setting "subtracted" to True does fix the background level offset.  Code looks straightforward enough.

mcara:
Updated background values based on the discussion here: https://github.com/spacetelescope/jwst/pull/10525#issuecomment-4434541500

src/stcal/resample/resample.py
```diff
@@ -506,6 +506,9 @@ def create_output_model(self):
             "start_time": None,
             "end_time": None,
             "duration": 0.0,
+            # background level
+            "level": None,
+            "subtracted": False,
```
emolter, line 511: I thought the whole point was to set subtracted=True?

emolter (changes requested):

src/stcal/resample/resample.py
melanieclarke: I wondered that too, but setting the value to None should have the same effect:
https://github.com/spacetelescope/stcal/blob/3ce70fbe3bd4f65ea01033aa9ee44845a0b7cba6/src/stcal/resample/resample.py#L820-L821

Given that, I think subtracted=False is less confusing.

emolter: subtracted=None also?

melanieclarke: As in, we should unset the subtracted boolean as well as the level value?  That ought to work as well.

emolter: yes that is what I meant

mcara: are you saying a boolean is allowed to be None in schema? If so, yes, we can do that too. However, I do not find anything confusing about `subtracted=False` when `level=None`. Has the level value `None` been subtracted? No! (hence `False`). I think it's pretty logical. What does `subtracted=True` or `subtracted=False` mean? It's pretty obvious. However, `subtracted=None` is puzzling (it is not `True` and it's not `False`; what is it?). The question "was the value in `level` subtracted?" can have only a binary answer, IMO.

mcara: I think the value of `None` for the `level` quite clearly indicates that sky level was not evaluated => no need to subtract it.

emolter: @mcara I can see your point, I am fine with subtracted=False.

In a datamodel, setting a metadata attribute to None means that it will not get preserved through a save-load cycle at all, i.e., the fits header keyword won't even be there in the file.  Which was the reason for my suggestion.

emolter (approved):
I think we are splitting hairs at this point, and I'm fine with whatever option we go with

src/stcal/resample/resample.py
melanieclarke: I'm also fine with the current state.

mcara: Hm, it's a good point. However, I think I would prefer to leave this PR as is and, instead, adjust the JWST PR to set `subtract` to `None`  to avoid writing the keywords.

melanieclarke:
I think this is ready to go but needs romancal approval.  @schlafly?

@mcara - can you please also rerun the romancal regtests?  The last run from a few weeks ago has some unrelated failures.

schlafly (approved):

mcara:
I re-run https://github.com/spacetelescope/RegressionTests/actions/runs/25517815534 and there are 4 failures that I do not think are related to this PR.

schlafly:
Those are fine and I just okified them; I've also approved this PR.

melanieclarke (approved):
Thanks all.  I will go ahead and merge now.

PR #543, opened 2026-05-07
Allow sregion_to_footprint to accept sregions containing multiple polygons

### LINKED ISSUE #10513: Resample s_region combiner errors out if input s_region has multiple polygons
_Issue [JP-4349](https://jira.stsci.edu/browse/JP-4349) was created on JIRA by [Ned Molter](https://jira.stsci.edu/secure/ViewProfile.jspa?name=emolter):_

Resample attempts to combine the input file S_REGION values into a single combined output footprint.  However if any of the input S_REGIONs itself contains multiple polygons, the code doesn't know how to handle it and gives a hard-to-understand error that looks like
```
shapely.errors.GEOSException: TopologyException: side location conflict at 2318.534796107444 1271.0075765540262. This can occur if the input geometry is invalid.```
This was noticed when I was trying to re-resample an i2d file into a new WCS, which is an edge case and probably a bad idea in the first place, so there's unlikely any effect on pipeline products.  Nevertheless, the rest of resample has no trouble handling this case and it really should be straightforward to support the s_region part of it too.

### PR DESCRIPTION
**Pull request by emolter**
Resolves [JP-4349](https://jira.stsci.edu/browse/JP-4349)

Closes spacetelescope/jwst#10513

This PR:
- Allows `sregion_to_footprint` to accept as input S_REGIONs that are themselves multiple polygons.  In this case a list of footprint arrays is returned instead of a single footprint array.
- Fixes a bug in `combine_sregions` where it expected all input footprints to have the same number of vertices.

I have also confirmed that the original use-case from the JP ticket passes with this stcal version.

## Tasks

  - [x] if your change breaks existing functionality, also add a `changes/<PR#>.breaking.rst` news fragment

### DIFF
--- a/src/stcal/alignment/util.py
+++ b/src/stcal/alignment/util.py
@@ -323,9 +323,6 @@ def sregion_to_footprint(s_region: str) -> np.ndarray | list[np.ndarray]:
     """
     Parse the s_region string and return the footprint as a list of Nx2 arrays.
 
-    This needs to be a list to account for the case that the input string contains
-    multiple polygons.
-
     Parameters
     ----------
     s_region : str
@@ -334,8 +331,8 @@ def sregion_to_footprint(s_region: str) -> np.ndarray | list[np.ndarray]:
     Returns
     -------
     footprint : np.ndarray or list[np.ndarray]
-        If the S_REGION contains a single polygon, returns a 2D array of shape (N, 2).
-        If the S_REGION contains multiple polygons, returns a list of such arrays.
+        If the input contains a single polygon, returns a 2D array of shape (N, 2).
+        If the input contains multiple polygons, returns a list of such arrays.
     """
     polygons = re.split(r"POLYGON\s+ICRS\s*", s_region, flags=re.IGNORECASE)
     footprints = []
--- a/src/stcal/alignment/util.py
+++ b/src/stcal/alignment/util.py
@@ -319,9 +319,12 @@ def calc_rotation_matrix(roll_ref: float, v3i_yangle: float, vparity: int = 1) -
     return [pc1_1, pc1_2, pc2_1, pc2_2]
 
 
-def sregion_to_footprint(s_region: str) -> np.ndarray:
+def sregion_to_footprint(s_region: str) -> np.ndarray | list[np.ndarray]:
     """
-    Parse the s_region string and return the footprint as an Nx2 array.
+    Parse the s_region string and return the footprint as a list of Nx2 arrays.
+
+    This needs to be a list to account for the case that the input string contains
+    multiple polygons.
 
     Parameters
     ----------
@@ -330,11 +333,21 @@ def sregion_to_footprint(s_region: str) -> np.ndarray:
 
     Returns
     -------
-    footprint : np.ndarray
-        A 2D array of the footprint of the region, shape (N, 2)
+    footprint : np.ndarray or list[np.ndarray]
+        If the S_REGION contains a single polygon, returns a 2D array of shape (N, 2).
+        If the S_REGION contains multiple polygons, returns a list of such arrays.
     """
-    no_prefix = re.sub(r"[a-zA-Z]", "", s_region)
-    return np.array(no_prefix.split(), dtype=float).reshape(-1, 2)
+    polygons = re.split(r"POLYGON\s+ICRS\s*", s_region, flags=re.IGNORECASE)
+    footprints = []
+    for polygon in polygons:
+        polygon = polygon.strip()
+        if not polygon:
+            continue
+        coords = np.array(polygon.split(), dtype=float).reshape(-1, 2)
+        footprints.append(coords)
+    if len(footprints) == 1:
+        return footprints[0]
+    return footprints
 
 
 def _compute_bounding_box_with_offsets(
@@ -492,9 +505,17 @@ def wcs_from_sregions(
         The WCS object corresponding to the combined input footprints.
 
     """
-    footprints = [
-        sregion_to_footprint(s_region) if isinstance(s_region, str) else s_region for s_region in footprints
-    ]
+    parsed_footprints = []
+    for s_region in footprints:
+        if isinstance(s_region, str):
+            result = sregion_to_footprint(s_region)
+            if isinstance(result, list):
+                parsed_footprints.extend(result)
+            else:
+                parsed_footprints.append(result)
+        else:
+            parsed_footprints.append(s_region)
+    footprints = parsed_footprints
     fiducial = _calculate_fiducial(footprints, crval=crval)
     crval = fiducial
     v3yangle = np.deg2rad(ref_wcsinfo["v3yangle"])
--- a/src/stcal/alignment/util.py
+++ b/src/stcal/alignment/util.py
@@ -319,7 +319,7 @@ def calc_rotation_matrix(roll_ref: float, v3i_yangle: float, vparity: int = 1) -
     return [pc1_1, pc1_2, pc2_1, pc2_2]
 
 
-def sregion_to_footprint(s_region: str) -> np.ndarray | list[np.ndarray]:
+def sregion_to_footprint(s_region: str) -> list[np.ndarray]:
     """
     Parse the s_region string and return the footprint as a list of Nx2 arrays.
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/stcal/alignment/util.py [hunk 1]
```diff
@@ -330,11 +333,21 @@ def sregion_to_footprint(s_region: str) -> np.ndarray:
 
     Returns
     -------
-    footprint : np.ndarray
-        A 2D array of the footprint of the region, shape (N, 2)
+    footprint : np.ndarray or list[np.ndarray]
+        If the S_REGION contains a single polygon, returns a 2D array of shape (N, 2).
+        If the S_REGION contains multiple polygons, returns a list of such arrays.
```
emolter: It might have been more straightforward in some ways to always return a list, with that list being single-element in the (most-oft-used) case that this is a single polygon.  However, there are [downstream usages](https://github.com/search?q=org%3Aspacetelescope%20sregion_to_footprint&type=code) and this is the least invasive in terms of API changes.

If you think we should go ahead and change the behavior though, I can update the JWST usages to reflect that - I think all JWST usages are in tests.

emolter:
jwst regtests https://github.com/spacetelescope/RegressionTests/actions/runs/25515764155
romancal regtests https://github.com/spacetelescope/RegressionTests/actions/runs/25515798548

JWST tests all passing
Romancal failures seem to be due to a difference in the CRDS rmap, unrelated to this change.

src/stcal/alignment/util.py
zacharyburnett, line 1: I don't like how `sregion_to_footprints` output can be a list or a single entry (which is indexable like a list), that feels like a bug in the making and forces users to check what the return type is

Could we always output a list?

emolter: yes I can do that if preferred, that was the "alternative suggestion" in https://github.com/spacetelescope/stcal/pull/543#discussion_r3203867793.  It will make this require a coordinated JWST PR but it won't be too hard to make

zacharyburnett: oh sorry I should have seen that 😅 my opinion would be to do that, yes

emolter: https://github.com/spacetelescope/stcal/pull/543/commits/d4fad6deeb7948903510bfd82a2bb20c14188998 changes that behavior and https://github.com/spacetelescope/jwst/pull/10532 updates all usages in JWST.  It doesn't look like there are any usages elsewhere in the org.  New jwst regression tests are at https://github.com/spacetelescope/RegressionTests/actions/runs/25694678936

tapastro:
Should this have a breaking changelog fragment? It's an API change to a function listed in __all__, though it is a helper more than a top-level algorithm. I'd lean towards yes, but I'm not sure.

zacharyburnett:
better to be safe than sorry, I would add one especially since we don't know ALL the downstream usages and it's technically public

emolter:
@tapastro @zacharyburnett I added the breaking changelog entry.

I know @braingram had some reservations about this PR, will wait for his review here

braingram:
I'm still trying to get up to speed on this. There are questions I have about:
- the format of the multiple region sregions
- if we expect these will be used by roman
- how we handle these changes and how that impacts the allowed versions

To start on the last point, the current solution links this PR to jwst changes which (if I'm thinking of this correctly) will trigger a stcal 2.0.0 version (needed since these changes aren't compatible with jwst 2.0.0). WRT roman the changes aren't of any functional consequence (since roman doesn't use `combine_sregions` and doesn't use `sregion_to_footprint` directly or indirectly although I could be wrong here as it's hard to follow).

I think reason the failure was triggered could be fixed by updating jwst to provide only single region sregions to stcal by updating code here:
https://github.com/spacetelescope/jwst/blob/a3708dbf99da3b1a38eb856a23a8bbb89dd30b25/jwst/resample/resample_utils.py#L125
That relates to the 2nd point above of do we expect roman to use multiple region sregions? If the answer is no that would be reason to keep the stcal code focused on single region sregions. As far as I can tell there is no way for romancal to produce a multi-region sregion and there hasn't been discussion of adding this. It would be good for @schlafly to chime in here. If the answer is "yes" and roman wants to add multiple region sregions than we should consider the first point above about the format of these regions.

AFAIK these are intended to be STC-S encoded regions.
https://www.ivoa.net/Documents/Notes/STC-S/20091030/NOTE-STC-S-1.33-20091030.html#_Toc184557732
For jwst I wasn't able to find any documentation on the sregion format beyond what's in the keyword dictionary (which doesn't describe multiple-region strings). Is there documentation for the jwst sregion format?
For roman the documentation is similarly lacking, however we do claim that these are STC-S encoded for some usage:
https://github.com/spacetelescope/rad/blob/7fa8290f8654f2c18c0989a31d24262a20a35f46/latest/CCSP/ccsp_custom_product.yaml#L126
For the usual (single region) case:
```
POLYGON ICRS 0.0 0.0 1.0 0.0 1.0 1.0 0.0 1.0
```
This is (as far as I can tell) a valid STC-S encoded string.
For a multi-region case I think the STC-S format should be:
```
Union ICRS (POLYGON 0.0 0.0 1.0 0.0 1.0 1.0 0.0 1.0 POLYGON 2.0 2.0 3.0 2.0 3.0 3.0 2.0 3.0)
```
but that's not what we're expecting here which is more like:
```
POLYGON ICRS 0.0 0.0 1.0 0.0 1.0 1.0 0.0 1.0 POLYGON ICRS 2.0 2.0 3.0 2.0 3.0 3.0 2.0 3.0
```
(which is also as far as I can tell what HST produces and what the archive expects).

To try and summarize, if this can be addressed by jwst-only changes I'm in favor of that approach since:
- we're expecting non-standard STC-S strings for sregions for jwst
- I'm not aware of any plans to use multi-region sregions for roman

schlafly:
Sorry, the multiple S_REGION case is for something like you have two detectors with a gap between them, and a coadd that covers the whole region, and you want to specify S_REGION as the actual filled in region (two squares separated by a gap) rather than the enclosing rectangle (four corners of the coadd)?

If that's right, for ops we don't currently plan this.  That said, one could imagine making our coadd s_regions more detailed to exclude gaps, which would require multiple polygon support.  It isn't high enough on my priority list that I would want to drive significant extra development work to support, though.

braingram:
Given the unstandard format of the multi-polygon sregions, difficulty in changing that at this point for jwst, and no planned multi-polygon sregions for romancal, how about this as an alternative approach?

- modify jwst to separate multi-polygon sregions into separate strings (1 polygon per string) before passing the single polygon strings to stcal
- deprecate `combine_sregions` and associated code
- move `combine_sregions` and associated code to jwst and fix the bug

Moving `combine_sregions` to jwst has the benefit of allowing changes to made there without linking PRs and although it might be moot the nircam-derived defaults:
https://github.com/spacetelescope/stcal/blob/3ce70fbe3bd4f65ea01033aa9ee44845a0b7cba6/src/stcal/alignment/resample_utils.py#L173
may not match what we want for roman.

braingram:
> Sorry, the multiple S_REGION case is for something like you have two detectors with a gap between them, and a coadd that covers the whole region, and you want to specify S_REGION as the actual filled in region (two squares separated by a gap) rather than the enclosing rectangle (four corners of the coadd)?

@schlafly I missed your question. @emolter would know best so perhaps he can chime in. It's my understanding that these can come from the WFSS mode or from gaps between detectors/exposures used for resampling.

emolter:
@schlafly @braingram Yes that's right, the reason it's useful to support this is to make the S_REGION limited to only the regions where there is real data, i.e., exclude detector gaps.  My understanding is that this helps CRDS be more accurate with their region searches.

I'm happy to do whatever is best for Romancal here that also fixes the bug in jwst.  If that's limiting the combiner code to just single-polygon S_REGION strings in stcal, then that's fine by me.

I'm a bit confused though about the suggestion to move `combine_sregions` to jwst.  I reached out to Eddie when making the original PR that added this, and we agreed that stcal was a good place for it.  As written it's usable for other observatories, and while I don't understand Roman products very well my understanding is that could include Roman in the future.  If things have changed we can discuss.

emolter:
Regarding the "NIRCam-derived results", there are easy ways to make this more general without removing the entire functionality from stcal.

schlafly:
Thanks Ned.  I think your description and Brett's of Roman's plans here are right; Roman could use this but doesn't have immediate plans to.  I'm not super worried about the precision of the approximation for defining the polygons; Roman might pick a bit worse precision but we won't complain about those defaults.  I think Brett mostly wants to avoid making the non-standards-compliant format something stcal has to support indefinitely.

emolter:
Ok sounds good.

in conclusion I should leave the stcal code as-is and handle this on the jwst side by parsing the multi-polygons there.  Is there consensus that that is the plan?

braingram:
> Ok sounds good.
> 
> in conclusion I should leave the stcal code as-is and handle this on the jwst side by parsing the multi-polygons there. Is there consensus that that is the plan?

That's what I vote for given:
- the evolving discussion on the roman side of "what is a valid s-region"
- uncertainty about what is a valid STC-S string
- for romancal operations s-regions are currently all single polygon STC-S strings

PR #546, opened 2026-05-20
JP-4313: Use explicit read_times in JWST model in ramp fit

melanieclarke:
Toward [JP-4313](https://jira.stsci.edu/browse/JP-4313)

For coordination with https://github.com/spacetelescope/jwst/pull/10534:
Use explicit read times for JWST data if available.  This is intended to support new multistripe modes with repeated in-frame reads of the same detector area.

## Tasks

### DIFF
--- a/changes/546.ramp_fitting.rst
+++ b/changes/546.ramp_fitting.rst
@@ -1,0 +1,2 @@
+Use explicit read times for likelihood fits with JWST data, if available in the input model.
+This is intended to support new multistripe modes with repeated in-frame reads of the same detector area.
--- a/src/stcal/ramp_fitting/ramp_fit.py
+++ b/src/stcal/ramp_fitting/ramp_fit.py
@@ -98,8 +98,12 @@ def create_ramp_fit_class(model, algorithm, dqflags=None, suppress_one_group=Fal
     )
 
     ramp_data.algorithm = algorithm
-    if hasattr(model.meta.exposure, "read_pattern"):
-        ramp_data.read_pattern = [list(reads) for reads in model.meta.exposure.read_pattern]
+
+    # Set the read pattern from input read times if available
+    read_times = getattr(model.meta.exposure, "read_times", None)
+    if read_times is not None and len(read_times) > 0:
+        log.debug("Using explicit read times")
+        ramp_data.read_pattern = list(read_times)
 
     ramp_data.set_dqflags(dqflags)
     ramp_data.start_row = 0


melanieclarke:
Regtests with jwst on main:
https://github.com/spacetelescope/RegressionTests/actions/runs/26599529754

Regtests with romancal on main:
https://github.com/spacetelescope/RegressionTests/actions/runs/26599567262

Regtests with the jwst branch + stdatamodels update:
https://github.com/spacetelescope/RegressionTests/actions/runs/26588854461

All passing.

schlafly:
I probably need to review better what romancal does, but I'm a little confused.  The old version took read_pattern from the input model and converted it to a list of lists and put it at read_pattern on ramp_data.  The new version no longer looks for read pattern at all and instead looks for read_times, to my eyes basically redefining read_pattern -> read_times for models but not for the ramp data.  Why can't Webb use the old name?  I would have naively expected romancal to break since it uses read_pattern as the name
https://github.com/spacetelescope/romancal/blob/main/romancal/ramp_fitting/ramp_fit_step.py#L259
and haven't yet figured out why it still works with the new name.

melanieclarke:
> I probably need to review better what romancal does, but I'm a little confused. The old version took read_pattern from the input model and converted it to a list of lists and put it at read_pattern on ramp_data. The new version no longer looks for read pattern at all and instead looks for read_times, to my eyes basically redefining read_pattern -> read_times for models but not for the ramp data. Why can't Webb use the old name? I would have naively expected romancal to break since it uses read_pattern as the name https://github.com/spacetelescope/romancal/blob/main/romancal/ramp_fitting/ramp_fit_step.py#L259 and haven't yet figured out why it still works with the new name.

Romancal doesn't use `create_ramp_fit_class`, it creates the ramp fit class directly.  The ramp data class takes a list of lists of read times as the "read_pattern", so even if romancal did use `create_ramp_fit_class`, it wouldn't be correct to directly read `meta.exposure.read_pattern` into the ramp class `read_pattern` attribute.

I'd like for Webb to use `meta.exposure.read_times` instead of `meta.exposure.read_pattern` in the datamodel because we need to directly store a list of lists of read times, not a frame averaging pattern.

At some point, I think it would make sense to move `create_ramp_fit_class` from stcal to jwst to make it clearer that we each need to set up our own ramp fit classes, but I didn't want to do it here.

braingram (approved):
Thanks! Approving for the roman side.

tapastro (approved):

tapastro (approved):

PR #549, opened 2026-06-02
Fix #501: Use absolute URL for STScI logo in README

### LINKED ISSUE #501: pypi shows broken image for rendered readme
https://pypi.org/project/stcal/

Image

My vote is to remove the logo from the readme.

karlhillx:
The relative-path `img` tag (`docs/_static/stsci_pri_combo_mark_white.png`) renders on GitHub but breaks on PyPI. Switch to an absolute `raw.githubusercontent.com` URL so the brandmark displays on both.

Closes #501.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 STScI Calibration algorithms and tools.
 
-<img src="docs/_static/stsci_pri_combo_mark_white.png" width="300"/>
+<img src="https://raw.githubusercontent.com/spacetelescope/stcal/main/docs/_static/stsci_pri_combo_mark_white.png" width="300"/>
 
 > [!IMPORTANT]
 > Linux and MacOS platforms are tested and supported. Windows is not currently supported.


zacharyburnett:
I'm hesitant to say that removing the brandmark entirely is the best solution here

karlhillx:
Fair enough. I can switch the image src to an absolute raw URL so the brandmark renders on PyPI too:
``

Want me to pin to a commit SHA instead of main so old releases don't break retroactively? Either way, I can update the PR.

zacharyburnett:
if we're going to point to a URL anyway then perhaps we should point to a central location

in general PyPI readme rendering is important so I'll bring it up with our upcoming templating work

karlhillx:
Updated to keep the brandmark with an absolute `raw.githubusercontent.com` URL so it renders on PyPI too. If your templating work is going to centralize the logo URL anyway, happy to either land this as the interim fix or close in favor of that — your call.

PR #550, opened 2026-06-04
JP-4318: Updating Ramp Fitting Tests for New STCAL Slope Computation.

kmacdonald-stsci:
Resolves [JP-4318](https://jira.stsci.edu/browse/JP-4318)

This PR addresses using the median rate of a pixel, which is computed across all integrations of a pixel, for the computation of the rate and rateint slopes, using the Poisson variance as part of the weights. This should not happen, since this has the effect of later integrations affecting the slope of earlier integrations. To prevent this the weighting during the computation of rate and rateint slopes use only the read noise variance, rather than a combination of both Poisson and read noise variances.

## Tasks

### DIFF
--- a/docs/stcal/ramp_fitting/description.rst
+++ b/docs/stcal/ramp_fitting/description.rst
@@ -247,7 +247,12 @@ noise is:
 The slope for an integration depends on the slope and the combined variance of each segment's slope:
 
 .. math::
-   slope_{i} = \frac{ \sum_{s}{ \frac{slope_{s}} {var^C_{s}}}} { \sum_{s}{ \frac{1} {var^C_{s}}}}
+   slope_{i} = \frac{ \sum_{s}{ \frac{slope_{s}} {var^R_{s}}}} { \sum_{s}{ \frac{1} {var^R_{s}}}}
+
+This is equivalent to:
+
+.. math::
+   slope_{i} = (\sum_{s}{ \frac{slope_{s}} {var^R_{s}}}) var^R_{i}
 
 Exposure-level computations
 +++++++++++++++++++++++++++
@@ -273,7 +278,13 @@ The overall slope depends on the slope and the combined variance of the slope of
 segments, so is a sum over integration values computed from the segments:
 
 .. math::
-    slope_{o} = \frac{ \sum_{i}{ \frac{slope_{i}} {var^C_{i}}}} { \sum_{i}{ \frac{1} {var^C_{i}}}}
+    slope_{o} = \frac{ \sum_{i}\sum_s{ \frac{slope_{i,s}} {var^R_{i,s}}}} { \sum_{i}{ \frac{1} {var^R_{i}}}}
+
+where the subscript :math:`i, s` means the segment :math:`s` in integration :math:`i`. This is above
+equation is equivalent to:
+
+.. math::
+    slope_{o} = (\sum_{i}\sum_s{ \frac{slope_{i,s}} {var^R_{i,s}}}) var^R_{o}
 
 
 .. _ramp_error_propagation:
--- a/docs/stcal/ramp_fitting/description.rst
+++ b/docs/stcal/ramp_fitting/description.rst
@@ -238,13 +238,14 @@ The variance of the slope for an integration due to Poisson noise is:
 .. math::
    var^P_{i} = \frac{1}{ \sum_{s} \frac{1}{ var^P_{s}}}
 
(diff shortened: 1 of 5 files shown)


kmacdonald-stsci:
Regression tests run here:

https://github.com/spacetelescope/RegressionTests/actions/runs/26763461997

docs/stcal/ramp_fitting/description.rst
```diff
@@ -273,7 +278,13 @@ The overall slope depends on the slope and the combined variance of the slope of
 segments, so is a sum over integration values computed from the segments:
 
 .. math::
-    slope_{o} = \frac{ \sum_{i}{ \frac{slope_{i}} {var^C_{i}}}} { \sum_{i}{ \frac{1} {var^C_{i}}}}
+    slope_{o} = \frac{ \sum_{i}\sum_s{ \frac{slope_{i,s}} {var^R_{i,s}}}} { \sum_{i}{ \frac{1} {var^R_{i}}}}
+
+where the subscript :math:`i, s` means the segment :math:`s` in integration :math:`i`. This is above
```
drlaw1558: 'This is above equation'

drlaw1558:
In order to properly reflect the weights being used in the optional output data products I think we also need to change:

Lines 3191-3192 from:
```
tmp = 1. / seg->var_e;
seg->weight = tmp * tmp;
```
to
```
tmp = 1. / seg->var_r;
seg->weight = tmp
```

Lines 3284-3287 from:
```
tmp = (seg->var_p + seg->var_r);
wt = 1. / tmp;
wt *= wt;
seg->weight = wt;
```
to
```
tmp = seg->var_r;
wt = 1. / tmp;
seg->weight = wt;
```

Lines 3442-3443 from:
```
seg->weight = 1. / seg->var_e;
seg->weight *= seg->weight;
```
to
```
seg->weight = 1. / seg->var_r;
```

docs/stcal/ramp_fitting/description.rst
kmacdonald-stsci: I changed this.

src/stcal/ramp_fitting/src/slope_fitter.c
```diff
@@ -3188,7 +3188,7 @@ ramp_fit_pixel_integration_fit_slope_seg_len1(
     seg->var_e = seg->var_p + seg->var_r;
 
     if (rd->save_opt) {
-        tmp = 1. / seg->var_e;
+        tmp = 1. / seg->var_r;
         seg->weight = tmp * tmp;
```
drlaw1558: One more; I think this line needs to be seg->weight = tmp

drlaw1558 (approved):
Looks good to me in local testing now; thanks @kmacdonald-stsci !

docs/stcal/ramp_fitting/description.rst
```diff
@@ -247,7 +247,12 @@ noise is:
 The slope for an integration depends on the slope and the combined variance of each segment's slope:
```
melanieclarke: This still says combined variance - should it be "read noise variance" instead?

docs/stcal/ramp_fitting/description.rst
melanieclarke: Here too

```suggestion
The overall slope depends on the slope and the read noise variance of the slope of each integration's
```

melanieclarke (commented):
Looks like there are a number of unit tests that need updating still.

docs/stcal/ramp_fitting/description.rst
kmacdonald-stsci: Fixed.

docs/stcal/ramp_fitting/description.rst
kmacdonald-stsci: Fixed.

kmacdonald-stsci:
> Looks like there are a number of unit tests that need updating still.

All the "cases", which tests optional results, were failing due to the change in the weighting computation. They have all been updated and passing now.

tapastro (approved):
I think this PR is good to go, but the paired JWST PR still has some questions. I'll hold off on merging until that one is squared away.

tapastro:
Romancal RTs: https://github.com/spacetelescope/RegressionTests/actions/runs/27356319671

braingram (approved):
Approving for roman.

schlafly (approved):

tapastro:
Running a set of RTs to generate results to okify: https://github.com/spacetelescope/RegressionTests/actions/runs/27568103681

PR #556, opened 2026-06-25
JP-4352: Fix floating point overflow in variance weighting for resampled spectra

melanieclarke:
Resolves [JP-4352](https://jira.stsci.edu/browse/JP-4352)

Fix a floating point overflow problem that typically occurs with NIRSpec point sources and 'ivm' weights.  Typical flux values for point sources in MJy are ~1e-13, error values ~1e-14, and variance values ~1e-28.  The very small variance values can lead to overflow if they are not handled carefully.   This PR fixes one place where variance weights were multiplied together before being divided out, leading to exactly zero error and variance images.  

This regression was introduced in jwst v1.18.0, when the resampling code was refactored and moved from jwst to stcal.  I have added a unit test for data with typical MJy flux scales and verified that it fails without the fix here, so hopefully the regression will not be reintroduced in the future.

## Tasks

### DIFF
--- a/changes/556.resample.rst
+++ b/changes/556.resample.rst
@@ -1,1 +1,1 @@
-Avoid floating point overflow causing exactly zero error and variance images with "ivm" weights for input data in MJy.
+Avoid floating point overflow causing zero-valued error and variance images with "ivm" weights for input data in MJy.
--- a/changes/556.resample.rst
+++ b/changes/556.resample.rst
@@ -1,0 +1,1 @@
+Avoid floating point overflow causing exactly zero error and variance images with "ivm" weights for input data in MJy.
--- a/src/stcal/resample/resample.py
+++ b/src/stcal/resample/resample.py
@@ -1171,6 +1171,8 @@ class Resample:
             wsum = self._variance_info[varname]["wsum"]
             mask = (var >= 0) & np.isfinite(var) & (weight > 0)
             with warnings.catch_warnings():
+                # Multiplying the weights separately should avoid overflow in
+                # most situations. Catch the warning for extreme cases.
                 warnings.filterwarnings("ignore", "overflow encountered", RuntimeWarning)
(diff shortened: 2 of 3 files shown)


melanieclarke:
Regtests for jwst:
https://github.com/spacetelescope/RegressionTests/actions/runs/28186705024

Regtests for romancal:
https://github.com/spacetelescope/RegressionTests/actions/runs/28186761633

All passing.

mcara (approved):
LGTM

melanieclarke:
@emolter - can you please review for jwst?  @schlafly or @braingram - can you review for romancal?

changes/556.resample.rst
```diff
@@ -0,0 +1 @@
+Avoid floating point overflow causing exactly zero error and variance images with "ivm" weights for input data in MJy.
```
emolter:
```suggestion
Avoid floating point overflow causing zero-valued error and variance images with "ivm" weights for input data in MJy.
```
otherwise it sounds like "zero" refers to the number of images instead of their values

emolter (approved):
LGTM just one small comment about the change log

braingram (approved):
LGTM. Is it worth converting some of this to double precision?

melanieclarke:
> LGTM. Is it worth converting some of this to double precision?

We could do the math in this step at double precision, as an alternative to the fix here, but it would require explicit conversion, since the data is stored as float32.   I'm not sure if that's preferable to just being careful not to square variances in the few calculations where it's relevant.

mcara:
> LGTM. Is it worth converting some of this to double precision?

It's a good question. The reason I tried to avoid conversion to double is because of performance penalty and also memory penalty (when we were concerned about how many arrays resample keeps in memory). IMO, proposed adjustment is a good compromise unless we get further evidence that float32 is not enough

PR #562, opened 2026-07-15
JP-3892: Update Jump Documentation

kmacdonald-stsci:
Resolves [JP-3892](https://jira.stsci.edu/browse/JP-3892)

This PR addresses jump step documentation detailing the logical control flow of the step. Also, a small bug was fixed in the computation of `total_sigclip_groups`.

Regression test runs for the bug fix:
jwst:
https://github.com/spacetelescope/RegressionTests/actions/runs/29337654497
romancal:
https://github.com/spacetelescope/RegressionTests/actions/runs/30379195390

## Tasks

### DIFF
--- a/changes/562.jump.rst
+++ b/changes/562.jump.rst
@@ -1,0 +1,1 @@
+Updated the jump step documentation and fixed a small bug in computing ``total_sigclip_groups``.
--- a/docs/stcal/jump/description.rst
+++ b/docs/stcal/jump/description.rst
@@ -21,34 +21,31 @@ in `Anderson & Gordon (2011) <https://ui.adsabs.harvard.edu/abs/2011PASP..123.12
 
 Two-Point Difference Method
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
-The two-point difference method is applied to each integration as follows:
+The two-point difference method is applied as follows:
 
-#. Compute the first differences for each pixel (the difference between
-   adjacent groups)
-#. Compute the clipped median (dropping the largest difference) of the first differences for each pixel.
-   If there are only three first difference values (four groups), no clipping is
-   performed when computing the median.
-#. Use the median to estimate the Poisson noise for each group and combine it
-   with the read noise to arrive at an estimate of the total expected noise for
-   each difference.
-#. Compute the "difference ratio" as the difference between the first differences
-   of each group and the median, divided by the expected noise.
-#. If the largest "difference ratio" is greater than the rejection threshold,
-   flag the group corresponding to that ratio as having a jump.
-#. If a jump is found in a given pixel, iterate the above steps with the
-   jump-impacted group excluded, looking for additional lower-level jumps
-   that still exceed the rejection threshold.
-#. Stop iterating on a given pixel when no new jumps are found or only one
-   difference remains.
-#. If there are only two differences (three groups), the smallest one is compared to the larger
-   one and if the larger one is above a threshold, it is flagged as a jump.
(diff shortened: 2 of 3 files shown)


drlaw1558:
Thanks @kmacdonald-stsci ; I don't think we want to remove the existing information though, just add to it.

In rough words, something like:

"Jumps in the ramps of a given pixel are detected using statistics of the two-point differences between successive groups.  Those statistics either use sigma clipping or the method described by Anderson+Gordon.

If only_use_ints=True and nints > minimum_sigclip_groups: sigma clip across integrations for each group difference in the ramp (e.g., sigma clip groups 3-2 for all integrations, then sigma clip groups 4-3 for all integrations, etc).

If only_use_ints=False and nints*ngroups > minimum_sigclip_groups: sigma clip across all group differences and all integrations simultaneously (e.g., treat all group differences within an integration and in other integrations equally).

Otherwise if there are at least min_diffs_single_pass in each integration do (SOMETHING???) to find jumps.

Otherwise use the 10point iterative approach below."

I'm a little unclear on some things though:

- Even if not sigma clipping there's a split when min_usable_diffs >= twopt_p.min_diffs_single_pass to call look_for_more_than_one_jump, and otherwise to call iterative_jump.  min_diffs_single_pass says that it's a MIRI shower parameters in the docs, but that doesn't seem right- it looks like it affects everything.  What is look_for_more_than_one_jump doing though and how is this different from sigma clipping?  To answer my own question, I think it's the case that the astropy stats sigma clipping is empirically determining sigma from the ensemble of group differences, while the look_for_more_than_one_jump version is theoretically determining sigma from the nominal read+poisson noise.

- I don't think the description in the docs of minimum_groups is accurate.  It says the minimum number to use sigma clipping, but I think it's the minimum number to do ANY type of jump detection?  If so this seems to be labelled incorrectly in many places, from stcal readthedocs to both stcal and jwst inline docs.

kmacdonald-stsci:
@drlaw1558 I pushed an update to the documentation. Please take a look at it and let me know if this is more what you want. I think it also answers your questions.

@aliciacanipe

drlaw1558:
Thanks for clarifying @kmacdonald-stsci ; I had some additional thoughts on ways that we could rearrange some of the text to potentially be clearer.  This was easiest to communicate via a direct PR against your PR.  See https://github.com/kmacdonald-stsci/stcal/pull/2

kmacdonald-stsci:
> Thanks for clarifying @kmacdonald-stsci ; I had some additional thoughts on ways that we could rearrange some of the text to potentially be clearer. This was easiest to communicate via a direct PR against your PR. See [kmacdonald-stsci#2](https://github.com/kmacdonald-stsci/stcal/pull/2)

I integrated the your PR into this PR.

melanieclarke:
LGTM for jwst.  

I'm rerunning regression tests for the bug fix now, and started some romancal ones as well -- link in the top-level description.  @schlafly - can you please review for Roman when they're complete?

drlaw1558 (approved):
Thanks, LG2M

schlafly:
romancal regtest failure is a transient, all good from my perspective.

schlafly (approved):

melanieclarke (approved):

zacharyburnett:
the build failure should be resolved soon, by either merging https://github.com/spacetelescope/stcal/pull/563 or updating the pinned version of the openastronomy build workflow (this also applies to a bunch of other repos, there was an upstream `pipx` change that cascaded down)

PR #563, opened 2026-07-28
[SCSB-274] localize build workflow for trusted publishers

zacharyburnett:
Resolves [SCSB-274](https://jira.stsci.edu/browse/SCSB-274)

- localizes build workflow so we can set up Trusted Publishers (eventually) and not need a PyPI token
- avoids Python 3.12 `pipx` issue with upstream OpenAstronomy build workflow
- applies upcoming package template

**(this will need an update to the branch protection ruleset, because the build workflow is renamed from `build / Build source distribution and wheels` to `build / build`). Until then, a job called `build / Build source distribution and wheels` will be "`Expected`"**

## Tasks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,3 +67,6 @@ jobs:
           subject-path: "dist/*"
       # To upload to PyPI without a token, add this workflow file as a Trusted Publisher in the project settings on the PyPI website
       - uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
+        # TODO remove this when we verify a Trusted Publisher config exists on PyPI
+        with:
+          password: ${{ secrets.PYPI_PASSWORD_STSCI_MAINTAINER }}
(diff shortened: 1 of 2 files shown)


braingram:
Why did the build take 242 minutes? :-O

zacharyburnett:
> Why did the build take 242 minutes? :-O

it looks like running the import tests on the `musllinux_aarch64` wheels took the majority of the build time, I imagine because we don't have upstream wheels for that combo and therefore cibuildhweel had to build all of the dependencies to install them in the virtualenv

braingram (changes requested):
What are the alternatives for either:
- dropping building the aarch64 wheels (has anyone requested these?)
- building them a different way that doesn't take 4 hours

zacharyburnett:
> What are the alternatives for either:
> 
> * dropping building the aarch64 wheels (has anyone requested these?)
> * building them a different way that doesn't take 4 hours

Dropping building the `musllinux_aarch64` wheels worked. The step that made it take 4h was `test-command` so we could theoretically skip those tests, but I don't know if we can configure cibuildwheel to only skip tests on just one specific architecture / platform / compiler combo

PR #570, opened 2026-08-12
change `1.20.0` -> `2.0.0`?

zacharyburnett:
releasing a `1.20.0` would break existing installs of JWST and Romancal if the user updated their local environment


braingram:
I drafted this until we can sort out what's going on.

PR #584, opened 2026-09-12
Likelyfit fewresultants

t-brandt:
Resolves [JP-4461](https://jira.stsci.edu/browse/JP-4461)

Closes #10861

This PR addresses three issues mentioned in JP-4461:
First, a bug:
https://github.com/spacetelescope/stcal/blob/28c22836ea65dd878548fc2a0ebcbaa06ad262f6/src/stcal/ramp_fitting/likely_fit.py#L476
takes two arguments but 
https://github.com/spacetelescope/stcal/blob/28c22836ea65dd878548fc2a0ebcbaa06ad262f6/src/stcal/ramp_fitting/likely_fit.py#L568
supplies three.  

Next, a not-quite-bug:
https://github.com/spacetelescope/stcal/blob/28c22836ea65dd878548fc2a0ebcbaa06ad262f6/src/stcal/ramp_fitting/likely_fit.py#L346
declines to check pixels for jumps if they have insufficiently many clean group differences, but this check only happens after the initial jump flagging by the likelihood algorithm.  It really should happen earlier, and it should ideally respect the difference in the minimum number of differences required with and without group averaging.  The difference will matter if a pixel saturates within the first few reads. 
 
Finally, more of a feature: likelihood fitting can work with as few as two groups (one if we supply a pedestal); it is just jump detection that doesn't.  It would be nice to be able to call likelihood-based ramp fitting with fewer groups and have it skip the jump detection part if there are not enough groups, rather than throwing an error.

Note that this PR changes the behavior of jump detection when there are very few valid resultants, so there may be differences for a small number of pixels in regression tests where a jump was flagged in main but is not flagged in this branch.

## Tasks


schlafly:
Webb regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/34714077345
Roman regression tests: https://github.com/spacetelescope/RegressionTests/actions/runs/34714037988
We expect changes to small numbers of pixels due to turning off jump detection for pixels with too few valid resultants to detect jumps sensibly.

romancal regtests show the expected ~1k pixel differences neatly matching the ~1k changes in the number of detected jumps.

I'm less good at reading the Webb regtest failures; they're similar in that ~1k pixels changed values, consistent with mostly being about changing jump detection.  DQ only changes for tens of pixels, though, which is harder for me to understand.

t-brandt:
Note: I added an optional argument to force skipping of likelihood-based jump detection per a request from the Roman Mission Office.  Default behavior is unchanged.

src/stcal/ramp_fitting/likely_fit.py
```diff
@@ -10,12 +10,14 @@
 
 DELIM = "=" * 80
 SQRT2 = np.sqrt(2)
+# The variable below is used by the JWST pipeline.  In the future, it may
+# be overridden if desired to use the algorithm  with as few as 2 groups.
```
tapastro: Because we outsourced this value to stcal, I'd suggest that we left it to the shared repo to determine the threshold. If the "true threshold" is 2 instead of 4, I'd prefer we change this global and preserve the use of it below.

tapastro:
> I'm less good at reading the Webb regtest failures; they're similar in that ~1k pixels changed values, consistent with mostly being about changing jump detection. DQ only changes for tens of pixels, though, which is harder for me to understand.

As far as I can tell, most of these pixels already had a jump detected from our standalone jump step already run on the data - the small diff shows that of order 10 pixels were not detected in `jump` but were detected by the likelihood method's jump detection. 

I think the changes look good from the JWST side, pending the suggestion to preserve thresholding behavior through the global.

